PR #2, opened 2018-08-10
Add an initial set of resources

svanharmelen:
This initial release offers support for the following Terraform Enterprise resources:

1.  organizations
2.  organization tokens
3.  policies
4.  ssh keys
5.  teams
6.  team access
7.  team membership
8.  team memberships
9.  team tokens
10. variables
11. workspaces

This PR includes acceptance tests and documentation for all resources:

```
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -count=1 -timeout 10m
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestProvider_impl
--- PASS: TestProvider_impl (0.00s)
=== RUN   TestAccTFEOrganization_basic
--- PASS: TestAccTFEOrganization_basic (1.21s)
=== RUN   TestAccTFEOrganization_update
--- PASS: TestAccTFEOrganization_update (3.12s)
=== RUN   TestAccTFEOrganizationToken_basic
--- PASS: TestAccTFEOrganizationToken_basic (1.98s)
=== RUN   TestAccTFESentinelPolicy_basic
--- PASS: TestAccTFESentinelPolicy_basic (2.68s)
=== RUN   TestAccTFESentinelPolicy_update
--- PASS: TestAccTFESentinelPolicy_update (4.86s)
--- PASS: TestAccTFESSHKey_basic (1.94s)
--- PASS: TestAccTFESSHKey_update (3.70s)
=== RUN   TestAccTFETeamAccess_basic
--- PASS: TestAccTFETeamAccess_basic (3.32s)
=== RUN   TestAccTFETeamMember_basic
--- PASS: TestAccTFETeamMember_basic (2.20s)
=== RUN   TestAccTFETeamMembers_basic
--- PASS: TestAccTFETeamMembers_basic (2.28s)
=== RUN   TestAccTFETeamMembers_update
--- PASS: TestAccTFETeamMembers_update (3.74s)
--- PASS: TestAccTFETeam_basic (1.90s)
=== RUN   TestAccTFETeamToken_basic
--- PASS: TestAccTFETeamToken_basic (2.46s)
--- PASS: TestAccTFEVariable_basic (2.73s)
--- PASS: TestAccTFEVariable_update (4.53s)
=== RUN   TestAccTFEWorkspace_basic
--- PASS: TestAccTFEWorkspace_basic (1.87s)
=== RUN   TestAccTFEWorkspace_update
--- PASS: TestAccTFEWorkspace_update (2.94s)
PASS
ok      github.com/terraform-providers/terraform-provider-tfe/tfe       47.553s
```


tfe/provider.go
```diff
@@ -0,0 +1,194 @@
...
+		creds = auth.StaticCredentialsSource(staticTable)
+	}
+
+	return creds
+}
+
+var descriptions = map[string]string{
+	"hostname": "The remote backend hostname to connect to. Default to app.terraform.io.",
+	"token": "The token used to authenticate with the remote backend. If `TFE_TOKEN` is set\n" +
+		"or credentials for the host are configured in the CLI Config File, then those\n" +
```
tombuildsstuff: afaik in-line specified values override environment variables

tfe/resource_tfe_organization.go
```diff
@@ -0,0 +1,156 @@
...
+				ValidateFunc: func(v interface{}, k string) ([]string, []error) {
+					switch v.(string) {
+					case "password", "two_factor_mandatory":
+						// These are the two valid options, so we return without errors.
+						return nil, nil
+					default:
+						return nil, []error{fmt.Errorf(
+							"%q must be either 'password' or 'two_factor_mandatory'", k),
+						}
+					}
```
tombuildsstuff: _minor_ we can replace this with `validation.StringInSlice([]string{ "password", "two_factor_mandatory" }, false)`

tfe/resource_tfe_organization_test.go
```diff
@@ -0,0 +1,207 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckTFEOrganizationDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccTFEOrganization_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckTFEOrganizationExists(
+						"tfe_organization.foobar", org),
+					testAccCheckTFEOrganizationAttributes(org),
```
tombuildsstuff: I'd argue we don't need this since we're validating these below

tfe/resource_tfe_organization_test.go
```diff
@@ -0,0 +1,207 @@
...
+	email = "admin@company.com"
+}`
+
+const testAccTFEOrganization_update = `
+resource "tfe_organization" "foobar" {
+  name = "terraform-updated"
+	email = "admin-updated@company.com"
+	session_timeout = 3600
+	session_remember = 3600
+}`
```
tombuildsstuff: _minor_ can we make the spacing consistent?

tfe/resource_tfe_organization_token.go
```diff
@@ -0,0 +1,84 @@
...
+		Schema: map[string]*schema.Schema{
+			"organization": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"token": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
```
tombuildsstuff: I believe this wants `Sensitive: true`?

tfe/resource_tfe_organization_token.go
```diff
@@ -0,0 +1,84 @@
...
+	organization := d.Get("organization").(string)
+
+	log.Printf("[DEBUG] Create new token for organization: %s", organization)
+	token, err := tfeClient.OrganizationTokens.Generate(ctx, organization)
+	if err != nil {
+		return fmt.Errorf(
+			"Error creating new token for organization %s: %v", organization, err)
+	}
+
+	d.Set("token", token.Token)
```
tombuildsstuff: this should be below `d.SetId` and probably wants a comment explaining why it's here and not in the read (presumably that it's only returned once?)

tfe/resource_tfe_organization_token_test.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+
+	return nil
+}
+
+const testAccTFEOrganizationToken_basic = `
+resource "tfe_organization" "foobar" {
+  name = "terraform-test"
+	email = "admin@company.com"
```
tombuildsstuff: _minor_ it'd be good to make the spacing consistent

tfe/resource_tfe_sentinel_policy.go
```diff
@@ -0,0 +1,167 @@
...
+				Required: true,
+			},
+
+			"enforce_mode": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ValidateFunc: func(v interface{}, k string) ([]string, []error) {
+					switch v.(string) {
+					case "advisory", "hard-mandatory", "soft-mandatory":
```
tombuildsstuff: (as above) we can use `validation.StringInSlice([]string{"foo","bar"}, false)` here

tfe/resource_tfe_sentinel_policy.go
```diff
@@ -0,0 +1,167 @@
...
+	// Get the name and organization.
+	name := d.Get("name").(string)
+	organization := d.Get("organization").(string)
+
+	// Create a new options struct.
+	options := tfe.PolicyCreateOptions{
+		Name: tfe.String(name),
+		Enforce: []*tfe.EnforcementOptions{
+			&tfe.EnforcementOptions{
+				Path: tfe.String(name + ".sentinel"),
```
tombuildsstuff: do we want to add validation to `name` to check that the name doesn't have this suffix?

tfe/resource_tfe_sentinel_policy.go
tombuildsstuff: if there's enum's we can also do this with `string(tfe.EnforcementModeFoo)`

tfe/resource_tfe_sentinel_policy_test.go
```diff
@@ -0,0 +1,183 @@
...
+const testAccTFESentinelPolicy_basic = `
+resource "tfe_organization" "foobar" {
+  name = "terraform-test"
+	email = "admin@company.com"
+}
+
+resource "tfe_sentinel_policy" "foobar" {
+	name = "policy-test"
+	organization = "${tfe_organization.foobar.id}"
+	policy = "main = rule { true }"
```
tombuildsstuff: 🤔 do we want a test for uploading an invalid policy to ensure it errors?

tfe/resource_tfe_team_access.go
```diff
@@ -0,0 +1,124 @@
...
+		Delete: resourceTFETeamAccessDelete,
+
+		Schema: map[string]*schema.Schema{
+			"access": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: func(v interface{}, k string) ([]string, []error) {
+					switch v.(string) {
+					case "admin", "read", "write":
```
tombuildsstuff: (as above)

tfe/resource_tfe_team_access.go
```diff
@@ -0,0 +1,124 @@
...
+			log.Printf("[DEBUG] Team access %s does no longer exist", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading configuration of team access %s: %v", d.Id(), err)
+	}
+
+	// Update config.
+	d.Set("access", string(tmAccess.Access))
+	d.Set("team_id", tmAccess.Team.ID)
```
tombuildsstuff: is `Team` a pointer? if so, what happens if this is nil?

tfe/resource_tfe_team_member.go
```diff
@@ -0,0 +1,117 @@
...
+			log.Printf("[DEBUG] User %q does no longer exist", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading users from team %s: %v", teamID, err)
+	}
+
+	found := false
+	for _, user := range users {
+		if user.Username == username {
```
tombuildsstuff: I'm assuming that TFE is case-sensitive for usernames?

tfe/resource_tfe_team_member.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return nil
+}
+
+func packTeamMemberID(teamID, username string) string {
+	return teamID + "|" + username
+}
+
+func unpackTeamMemberID(id string) (teamID, username string) {
+	s := strings.SplitN(id, "|", 2)
```
tombuildsstuff: should we check if this is 2 segments and error if not? (e.g. when a users importing)

tfe/resource_tfe_team_members.go
```diff
@@ -0,0 +1,161 @@
...
+	// Get the team ID and username..
+	teamID := d.Get("team_id").(string)
+
+	// Create a new options struct.
+	options := tfe.TeamMemberRemoveOptions{}
+
+	// Add all the users that need to be added.
+	for _, username := range d.Get("usernames").(*schema.Set).List() {
+		options.Usernames = append(options.Usernames, username.(string))
+	}
```
tombuildsstuff: this can be removed (also d.Get("usernames") should return nil/an empty set here when the config no longer exists)

tfe/resource_tfe_team_token.go
```diff
@@ -0,0 +1,84 @@
...
+		Schema: map[string]*schema.Schema{
+			"team_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"token": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
```
tombuildsstuff: I believe this is a sensitive token? if so should this have `Sensitive: true`

tfe/resource_tfe_team_token.go
```diff
@@ -0,0 +1,84 @@
...
+	teamID := d.Get("team_id").(string)
+
+	log.Printf("[DEBUG] Create new token for team: %s", teamID)
+	token, err := tfeClient.TeamTokens.Generate(ctx, teamID)
+	if err != nil {
+		return fmt.Errorf(
+			"Error creating new token for team %s: %v", teamID, err)
+	}
+
+	d.Set("token", token.Token)
```
tombuildsstuff: (as above) can we add a comment to explain why this is here and not in the Read

tfe/resource_tfe_team_token.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return nil
+}
+
+func resourceTFETeamTokenDelete(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	// Get the team ID.
+	teamID := d.Get("team_id").(string)
```
tombuildsstuff: we should use `d.Id()` here since this'll be nil otherwise

tfe/resource_tfe_team_members.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return nil
+}
+
+func resourceTFETeamMembersDelete(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	// Get the team ID and username..
+	teamID := d.Get("team_id").(string)
```
tombuildsstuff: we should use the ID here since this'll be nil otherwise

tfe/resource_tfe_variable.go
```diff
@@ -0,0 +1,187 @@
...
+				Required: true,
+			},
+
+			"category": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: func(v interface{}, k string) ([]string, []error) {
+					switch v.(string) {
+					case "env", "terraform":
```
tombuildsstuff: (as above)

tfe/resource_tfe_variable.go
```diff
@@ -0,0 +1,187 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
+			"sensitive": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
```
tombuildsstuff: /technically/ we could use this value in a CustomizeDiff function to set the `value` schema field to a sensitive field too, but that may be a bad idea - WDTY @radeksimko?

website/docs/d/oauth_token.html.markdown
```diff
@@ -0,0 +1,41 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_oauth_token"
+sidebar_current: "docs-tfe-datasource-oauth-token"
+description: |-
+  Use this datasource to get the ID of the OAuth token of an assocaited VCS configuration.
+---
+
+# tfe_oauth_token
+
+Use this datasource to get the ID of the OAuth token of an assocaited VCS
```
tombuildsstuff: `assocaited` -> `associated`

website/docs/d/oauth_token.html.markdown
```diff
@@ -0,0 +1,41 @@
...
+
+```hcl
+data "tfe_oauth_token" "oauth_token" {
+  template_filter = "featured"
+}
+```
+
+### Argument Reference
+
+* `template_filter` - (Required) The template filter. Possible values are `featured`, `self`, `selfexecutable`, `sharedexecutable`, `executable` and `community` (see the Cloudstack API *listTemplate* command documentation).
```
tombuildsstuff: can we link to the Cloudstack API docs?

website/docs/index.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+page_title: "Provider: Terraform Enterprise"
+sidebar_current: "docs-tfe-index"
+description: |-
+  The Terraform Enterprise provider is used to interact with the many resources supported by (Private) Terraform Enterprise. The provider needs to be configured with the proper credentials before it can be used.
+---
+
+# Terraform Enterprise Provider
+
+The Terraform Enterprise provider is used to interact with the many resources
+supported by (Private) [Terraform Enterprise](https://www.hashicorp.com/products/terraform).
```
tombuildsstuff: if this works with SaaS TFE - should we ditch `Private` here and above?

website/docs/index.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+
+We recommend omitting the token which can be provided as an environment
+variable or set as [credentials in the CLI Config File](/docs/commands/cli-config.html#credentials).
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `hostname` - (Optional) The remote backend hostname to connect to. Default
+  to app.terraform.io.
```
tombuildsstuff: can we `quote` app.terraform.io?

website/docs/index.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+* `token` - (Optional) The token used to authenticate with the remote backend.
+  If `TFE_TOKEN` is set or credentials for the host are configured in the CLI
+  Config File, then this will override any saved value for this.
```
tombuildsstuff: (as above) the config overrides the env variable

website/docs/r/organization.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_organization"
+sidebar_current: "docs-tfe-organization"
+description: |-
+  Creates, updates and destroys organizations.
+---
+
+# tfe_organization
+
+Creates, updates and destroys organizations.
```
tombuildsstuff: `Creates, updates and destroys` -> `Manages`

website/docs/d/oauth_token.html.markdown
```diff
@@ -0,0 +1,41 @@
+---
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_oauth_token"
+sidebar_current: "docs-tfe-datasource-oauth-token"
+description: |-
+  Use this datasource to get the ID of the OAuth token of an assocaited VCS configuration.
+---
+
+# tfe_oauth_token
```
tombuildsstuff: I don't see a Data Source for this?

website/docs/r/organization.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+The following arguments are supported:
+
+* `name` - (Required) Name of the organization.
+* `email` - (Required) Admin email address.
+* `session_timeout` - (Optional) Session timeout after inactivity (minutes).
+  Defaults to 20160.
+* `session_remember` - (Optional) Session expiration (minutes). Defaults to
+  20160.
+* `collaborator_auth_policy` - (Optional) Authentication policy (`password`
+  or `two_factor_mandatory`). Defaults to `password`.
```
tombuildsstuff: is this defaulted at the API?

website/docs/r/organization.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Name of the organization.
+* `email` - (Required) Admin email address.
+* `session_timeout` - (Optional) Session timeout after inactivity (minutes).
+  Defaults to 20160.
+* `session_remember` - (Optional) Session expiration (minutes). Defaults to
+  20160.
```
tombuildsstuff: can we quote `20160` here since it's a default value?

website/docs/r/organization_token.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+page_title: "Terraform Enterprise: tfe_organization_token"
+sidebar_current: "docs-tfe-organization-token"
+description: |-
+Generates a new organization token, replacing any existing token.
+---
+
+# tfe_organization_token
+
+Generates a new organization token, replacing any existing token. This token
+can be used to act as the organization service account.
```
tombuildsstuff: should the resource for this error if an existing token exists?

website/docs/r/sentinel_policy.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Name of the policy.
+* `organization` - (Required) Name of the organization.
+* `policy` - (Required) The actual policy itself.
+* `enforce_mode` - (Required) The enforcement level of the policy. Valid
+  values are `advisory`, `hard-mandatory` and `soft-mandatory`. Defaults
+  to `soft-mandatory`.
```
tombuildsstuff: is this defaulted in the API?

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_team"
+sidebar_current: "docs-tfe-team"
+description: |-
+  Creates and destroys teams.
+---
+
+# tfe_team
+
+Creates and destroys teams.
```
tombuildsstuff: Manages teams*

website/docs/r/team_access.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+Basic usage:
+
+```hcl
+resource "tfe_team" "team" {
+  name = "my-team-name"
+  organization = "my-org-name"
+}
+
+resource "tfe_workspoce" "workspace" {
```
tombuildsstuff: `tfe_workspoce` -> `tfe_workspace`

website/docs/r/team_member.html.markdown
```diff
@@ -0,0 +1,41 @@
...
+sidebar_current: "docs-tfe-team-member"
+description: |-
+  Add or remove a user from a team.
+---
+
+# tfe_team_member
+
+Add or remove a user from a team.
+
+~> NOTE on managing team memberships: Terraform currently provides two resources
```
tombuildsstuff: _minor_ I'd recommend updating `NOTE` to be `**NOTE:**` to make it stand out

website/docs/r/team_members.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+sidebar_current: "docs-tfe-team-members"
+description: |-
+  Add or remove a users from a team.
+---
+
+# tfe_team_members
+
+Add or remove a users from a team.
+
+~> NOTE on managing team memberships: Terraform currently provides two resources
```
tombuildsstuff: _minor_ I'd recommend updating `NOTE` to be `**NOTE:**` to make it stand out

website/docs/r/team_members.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_team_members"
+sidebar_current: "docs-tfe-team-members"
+description: |-
+  Add or remove a users from a team.
+---
+
+# tfe_team_members
+
+Add or remove a users from a team.
```
tombuildsstuff: Manages users in a team

website/docs/r/variable.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+resource "tfe_workspoce" "workspace" {
+	name = "my-workspace-name"
+	organization = "${tfe_organization.organization.id}"
+}
+
+resource "tfe_variable" "variable" {
+	key = "my_key_name"
+	value = "my_value_name"
+	category = "terraform"
+  workspace_id = "${tfe_workspoce.workspace.id}"
```
tombuildsstuff: `workspoce` -> `workspace`

website/docs/r/variable.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+Basic usage:
+
+```hcl
+resource "tfe_organization" "organization" {
+  name = "my-org-name"
+	email = "admin@company.com"
+}
+
+resource "tfe_workspoce" "workspace" {
```
tombuildsstuff: `workspoce` -> `workspace`

website/docs/r/workspace.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_workspoce"
+sidebar_current: "docs-tfe-workspace"
+description: |-
+  Workspaces represent running infrastructure managed by Terraform.
+---
+
+# tfe_workspoce
+
+Provides a workspace resource.
```
tombuildsstuff: maybe this'd be better as: `Manages a Workspace in Terraform Enterprise. Workspaces represent running infrastructure managed by Terraform`

website/docs/r/workspace.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+
+## Example Usage
+
+Basic usage:
+
+```hcl
+resource "tfe_workspoce" "my-workspace" {
```
tombuildsstuff: `tfe_workspoce` -> `tfe_workspace`

tfe/config_windows.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"path/filepath"
+	"syscall"
+	"unsafe"
+)
+
+var (
+	shell         = syscall.MustLoadDLL("Shell32.dll")
+	getFolderPath = shell.MustFindProc("SHGetFolderPathW")
+)
```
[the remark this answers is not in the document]
svanharmelen: I don't think so... This file is actually copied from the Terraform repo and if you look closely, it tries to find the users `Application Data` directory and not the `Home` directory.

And since the `Application Data` directory can be redirected to a location outside of the users `Home` directory (which is sometimes done in corporate environments), finding the `Home` directory would not be a guaranteed path to finding the `Application Data` directory.

tfe/provider.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+
+		ConfigureFunc: providerConfigure,
+	}
+}
+
+func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	// Get the hostname and token.
+	hostname := d.Get("hostname").(string)
+	token := d.Get("token").(string)
```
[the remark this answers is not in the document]
svanharmelen: No, not really... The difficulty here is that we want people to be able to use credentials that are configured in the CLI config file. These are loaded in lines L112-L118, so only after this is done, we can determine if we have a value for the token or not.

So we could maybe do a check and return an error below L118. Will add that so users will get an early error when no token was found.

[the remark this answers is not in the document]
tombuildsstuff: sure, below will work - it's mostly so that we return a helpful error message rather than returning a nil-pointer somewhere down the line 👍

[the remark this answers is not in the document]
svanharmelen: It shouldn't return a `nil` pointer as the TFE client would return a `Missing API token` error. But I wouldn't give that error during the provider configuration, so this makes sense 👍

Fixed in the next commit.

tfe/provider.go
```diff
@@ -0,0 +1,194 @@
...
+	return tfe.NewClient(cfg)
+}
+
+// cliConfig tries to find and parse the configuration of the Terraform CLI.
+// This is an optional step, so any errors are ignored.
+func cliConfig() *Config {
+	config := &Config{}
+
+	// Detect the CLI config file path.
+	configFilePath := os.Getenv("TERRAFORM_CONFIG")
```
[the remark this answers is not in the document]
svanharmelen: I don't think we should do that. This is an env var that is currently also used in Terraform itself (part of this logic is copied from the Terraform repo), but it isn't documented there either.

Not sure why it isn't documented, but I only added it here to make sure the behavior is inline with the behavior of Terraform itself. So I don't think I should document or change it here...

tfe/provider.go
svanharmelen: Yeah, I will indeed make sure any value declared in the provider configuration takes precedence over the env variable and the CLI config file. So it will be `inline` > `env var` > `CLI config file`.

Fixed in the next commit.

tfe/resource_tfe_organization_test.go
svanharmelen: I disagree. This test is checking if the values were written correctly to the state file. The other test is checking if the returned object contains the expected values.

tfe/resource_tfe_organization_test.go
svanharmelen: Whoeps! Yes, of course 👍 

Fixed in the next commit.

tfe/resource_tfe_organization_token.go
svanharmelen: Good point!

Fixed in the next commit.

tfe/resource_tfe_organization_token_test.go
svanharmelen: Sure thing... Will fix that in all tests of course 😏

tfe/resource_tfe_organization_token.go
svanharmelen: Why should it be below `d.SetId`?

Moved it and added a comment in the next commit.

tfe/resource_tfe_sentinel_policy.go
svanharmelen: Nice... Fixed in the next commit.

tfe/resource_tfe_organization_test.go
bflad: You can also validate this via `ImportStateVerify` testing, since it seems this resource is easily able to be setup for resource import.

In the schema:

```go
Importer: &schema.ResourceImporter{
  State: schema.ImportStatePassthrough,
},
```

As a new `TestStep`:

```go
{
  ResourceName: "tfe_organization.foobar",
  ImportState: true,
  ImportStateVerify: true,
},
```

Documented at the bottom of the resource documentation:

``````
## Import

`aws_acmpca_certificate_authority` can be imported by using the certificate authority Amazon Resource Name (ARN), e.g.

```
$ terraform import tfe_organization.example OrganizationName
```
``````

File: website/tfe.erb [hunk 45]
```diff
@@ -0,0 +1,66 @@
...
+                </li>
+
+                <li<%= sidebar_current("docs-tfe-index") %>>
+                    <a href="/docs/providers/tfe/index.html">Terraform Enterprise Provider</a>
+                </li>
+
+                <li<%= sidebar_current("docs-tfe-resource") %>>
+                    <a href="#">Resources</a>
+                    <ul class="nav nav-visible">
+                        <li<%= sidebar_current("docs-tfe-organization") %>>
```
bflad: When visiting the `tfe_organization` page it will highlight both `tfe_organization` and `tfe_organization_token` since the `sidebar_current` entry prefixes the other resource.

To workaround this with other providers, we append `-x` to the end of problematic cases like this, e.g. `docs-tfe-organization-x` here in the ERB and in the Markdown topmatter.

As another nitpick: We generally use `docs-resource-` instead of `docs-` in preparation for supporting data sources in the future (without needing to rewrite these later).

website/tfe.erb
```diff
@@ -0,0 +1,66 @@
...
+
+                        <li<%= sidebar_current("docs-tfe-sentinel-policy") %>>
+                            <a href="/docs/providers/tfe/r/sentinel_policy.html">tfe_sentinel_policy</a>
+                        </li>
+
+                        <li<%= sidebar_current("docs-tfe-ssh-key") %>>
+                            <a href="/docs/providers/tfe/r/ssh_key.html">tfe_ssh_key</a>
+                        </li>
+
+                        <li<%= sidebar_current("docs-tfe-team") %>>
```
bflad: Same here -- this `sidebar_current` will highlight all the other team ones unless its updated to not be a prefix of the others.

tfe/resource_tfe_sentinel_policy.go
svanharmelen: That's a god question. According to the API docs the path should always be the `name` of the policy + `.sentinel`. So if the name would be `Sander.sentinel`, the path should be `Sander.sentinel.sentinel`.

But this is a good one to double check to make sure the doc are also covering this corner case. _Checking..._

svanharmelen: Seems to work as advertised. But will double check with the TFE team just to be sure.

tfe/resource_tfe_team_access.go
svanharmelen: Fixed in the next commit.

tfe/resource_tfe_team_access.go
svanharmelen: You know what happens then 😏It would be a serious bug in the TFE API if this value was not returned, but to safeguard us here I added a check 👍 

Fixed in the next commit.

tfe/resource_tfe_team_member.go
svanharmelen: Yes. it saves (and returns) the names in the case that you entered it.

tfe/resource_tfe_team_member.go
svanharmelen: How do you mean when a user is importing? What could/would be the use case were this function is called with an ID that was not set by the `packTeamMemberID` function?

tfe/resource_tfe_team_members.go
svanharmelen: Why can this be removed? Ow because if the config is deleted this is now and empty set... Check so the solution is that I query the API for all the members of the team and then delete those.

Fixed in the next commit.

tfe/resource_tfe_team_token.go
svanharmelen: Fixed in next commit.

tfe/resource_tfe_team_token.go
svanharmelen: Yep... Fixed in next commit.

tfe/resource_tfe_team_token.go
svanharmelen: 👍  Fixed in next commit.

tfe/resource_tfe_team_members.go
svanharmelen: Fixed in the next commit.

tfe/resource_tfe_variable.go
svanharmelen: Fixed in next commit.

tfe/resource_tfe_variable.go
svanharmelen: That would maybe be the cleanest solution... But we can also just set the `value` field to be a always be a sensitive field. But indeed curious what @radeksimko thinks about this one? I'm fine with making a custom diff func if you think that is better...

website/docs/d/oauth_token.html.markdown
svanharmelen: Oeps... This file wasn't suppose to get merged. Will delete it from the commit.

website/docs/d/oauth_token.html.markdown
svanharmelen: As above... This file wasn't suppose to get merged. Will delete it from the commit.

website/docs/index.html.markdown
svanharmelen: Well it supports both SaaS TFE and private TFE installs... So I think it makes sense to write that in the docs.

website/docs/index.html.markdown
svanharmelen: Sure... Fixed in the next commit.

website/docs/index.html.markdown
svanharmelen: Fixed in the next commit.

website/docs/d/oauth_token.html.markdown
svanharmelen: Correct, this should not be in this commit. Fixed in the next commit.

website/docs/r/organization.html.markdown
svanharmelen: Fixed in next commit.

website/docs/r/organization_token.html.markdown
svanharmelen: Good question... Following the API I would say no. But I can see why this could cause unexpected situations. I'll discuss this within the TFE team first.

website/docs/r/sentinel_policy.html.markdown
svanharmelen: Yes... But will update the resource as well... Fixed in next commit.

website/docs/r/team.html.markdown
svanharmelen: Sure :) Fixed in next commit.

website/docs/r/team_access.html.markdown
svanharmelen: Thx! Fixed in next commit.

website/docs/r/team_member.html.markdown
svanharmelen: Fixed in next commit.

website/docs/r/team_members.html.markdown
svanharmelen: Fixed in next commit.

website/tfe.erb
svanharmelen: Good points! Fixed in next commit.

website/tfe.erb
svanharmelen: Fixed as well...

tfe/resource_tfe_organization_test.go
svanharmelen: If you don't mind I would like to postpone adding import (and datasources) the a next PR. This is big enough as it is right now :wink:

But good point and will certainly think about this in the next iteration 👍

website/docs/r/organization.html.markdown
svanharmelen: Fixed in next commit.

tfe/resource_tfe_sentinel_policy.go
svanharmelen: Double checked with the team, and my assumptions are correct. So we don't have to check if the name already end with `.sentinel` here.

tfe/resource_tfe_sentinel_policy_test.go
svanharmelen: Also discussed with the team. Currently there are no checks being done while creating uploading the policy. They are done just before using a policy. So there is no point in trying to upload a bad policy to see if we then get an error back... As that will never happen.

website/docs/r/organization_token.html.markdown
svanharmelen: Discussed this with the team... I will add a `force-recreate` option. If that is set to `true` we just warn an recreate, else we give an error!

Good catch! Fixed in next commit.

svanharmelen:
Closed by mistake, please continue this review in PR #3

PR #12, opened 2018-08-28
website: Explain how to authenticate when managing TFE w/ TFE

nfagerlund:
TFE automatically writes a .terraformrc file with a credentials block when it
does Terraform runs, so theoretically the provider could authenticate
automatically. However, the automatically provided token is a temporary "run"
token with extremely limited privileges, and in practice it cannot manage any of
this provider's resources.

This PR also adjusts some confusing wording that implied the provider might not work with the SaaS.

(Screenshot, bc callout-within-a-list-item markdown is a lil squirrelly so I tested it to be sure:)

screenshot_2018-08-27 provider terraform enterprise - terraform by hashicorp

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -3,17 +3,43 @@ layout: "tfe"
 page_title: "Provider: Terraform Enterprise"
 sidebar_current: "docs-tfe-index"
 description: |-
-  The Terraform Enterprise provider is used to interact with the many resources supported by (Private) Terraform Enterprise. The provider needs to be configured with the proper credentials before it can be used.
+  The Terraform Enterprise provider is used to interact with the many resources supported by Terraform Enterprise. The provider needs to be configured with the proper credentials before it can be used.
 ---
 
 # Terraform Enterprise Provider
 
 The Terraform Enterprise provider is used to interact with the many resources
-supported by (Private) [Terraform Enterprise](https://www.hashicorp.com/products/terraform).
-The provider needs to be configured with the proper credentials before it can
-be used.
+supported by [Terraform Enterprise](https://www.hashicorp.com/products/terraform).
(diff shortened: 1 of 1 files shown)


nfagerlund:
Question for the maintainers! What kind of token should people use with this provider? 

My best guess is "user token belonging to a member of the owners team in each organization you wish to manage?" Org token might be more conceptually appropriate, but 1: can't do cross-org management, and 2: can't manage certain resources within an org (like policies and SSH keys). Owners team service account is promising too, but has the same issue with cross-org access.

I'd like to update the PR with that info before it's merged.

website/docs/index.html.markdown
```diff
@@ -37,5 +39,11 @@ The following arguments are supported:
...
 * `token` - (Optional) The token used to authenticate with Terraform Enterprise.
-	We recommend omitting the token which can be set as `credentials` in the
+  If omitted, the provider will use a token from the `credentials` block in the
   [CLI config file](/docs/commands/cli-config.html#credentials).
+
+    -> **Note:** The CLI config file is only for use with command-line Terraform
+    runs. When using a Terraform Enterprise workspace to manage resources with
+    this provider, you must explicitly set `token` in the provider configuration
+    block. We recommend using an input variable for the token and marking the
+    corresponding variable in the Terraform Enterprise workspace as sensitive.
```
svanharmelen: I would make this a bit more explicit. So when using this in your terminal, we strongly recommend to use a token configured in the CLI config file. And only when this provider is being executed in TFE, we recommend you use an input variable to set the `token` field with a token that has the required rights.

nfagerlund:
Re-worded, and moved some things around. 

I talked to Robert about the token types to confirm that you should use a user token with this provider. Provided extra context in a note so users will know why we're telling them to do that.

svanharmelen:
Thanks @nfagerlund!

PR #16, opened 2018-09-07
Create resource_tfe_organization_vcs.go

djaboxx:
this currently only works for Github, uses updated go-tfe library. A PR has been submitted for this as well.
hashicorp/go-tfe#13


svanharmelen:
Thanks for your efforts @djaboxx! As discussed offline we are going to add support for this in the coming weeks, but in a slightly different way and probably with a different name for the resource.

I'll make sure to keep you in the loop. Again thanks for your efforts and enthusiasm!

PR #20, opened 2018-09-14
Add `terraform import` support for all resources

svanharmelen:
Well, all but one actually. The docs now have a line about the import usage and capabilities for each resource, so it should be clear how to use the import command with the different resources.

This PR superseeds (and includes) the work @thrashr888 started! Thanks for that @thrashr888!


svanharmelen:
I’m not going to merge and release this today (10pm here), so will add those Monday before merging 👍🏻

PR #26, opened 2018-10-25
Clarify ID format for imports

u2mejc:
This error isn't self explanatory. This PR adds clarity for a user to resolve the issue, inline with the error.


svanharmelen:
Thanks @u2mejc! LGTM!

PR #32, opened 2018-11-05
Tidy up docs and error messages for compound import IDs

nfagerlund:
This PR updates the error messages to tell you what they were expecting. (I used the wrong ID format when I was testing Sentinel policies, and ended up going on a minor wild goose chase because I thought I had a different problem.)

Also, there was one inaccurate spot in the docs (`team`).


nfagerlund:
Thanks! 

(NB: I don't have write access to anything in this org, so someone else will need to merge.)

nfagerlund:
@svanharmelen That's all good feedback! I pushed a commit to address it; feel free to squash if you want. Took a different tack for the wording in the docs, but it should still resolve the awkwardness.

svanharmelen:
Thanks @nfagerlund! Looking great now, so I'll go ahead and merge this right in 👍

PR #33, opened 2018-11-06
Add support for Sentinel policy sets

davidcelis:
Now that policy sets have been enabled, we can add support for it to the Terraform provider for TFE. We might want to wait until we've finalized the feature in TFE before this is merged, but we can at least get the feedback loop started!

### DIFF
--- a/tfe/resource_tfe_policy_set.go
+++ b/tfe/resource_tfe_policy_set.go
@@ -1,0 +1,277 @@
+package tfe
+
+import (
+	"fmt"
+	"log"
+
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceTFEPolicySet() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceTFEPolicySetCreate,
+		Read:   resourceTFEPolicySetRead,
+		Update: resourceTFEPolicySetUpdate,
+		Delete: resourceTFEPolicySetDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"global": &schema.Schema{
+				Type:          schema.TypeBool,
+				Optional:      true,
+				Default:       false,
+				ConflictsWith: []string{"workspace_external_ids"},
+			},
+
+			"organization": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"policy_ids": &schema.Schema{
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"workspace_external_ids": &schema.Schema{
+				Type:          schema.TypeSet,
+				Optional:      true,
+				Elem:          &schema.Schema{Type: schema.TypeString},
+				ConflictsWith: []string{"global"},
+			},
+		},
+	}
+}
+
+func resourceTFEPolicySetCreate(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	name := d.Get("name").(string)
+	organization := d.Get("organization").(string)
+
+	// Create a new options struct.
+	options := tfe.PolicySetCreateOptions{
+		Name:   tfe.String(name),
+		Global: tfe.Bool(d.Get("global").(bool)),
+	}
+
+	if desc, ok := d.GetOk("description"); ok {
+		options.Description = tfe.String(desc.(string))
+	}
+
+	// Set up the policies and workspaces.
+	for _, policyID := range d.Get("policy_ids").(*schema.Set).List() {
+		options.Policies = append(options.Policies, &tfe.Policy{ID: policyID.(string)})
+	}
+
+	for _, workspaceID := range d.Get("workspace_external_ids").(*schema.Set).List() {
+		options.Workspaces = append(options.Workspaces, &tfe.Workspace{ID: workspaceID.(string)})
+	}
+
+	log.Printf("[DEBUG] Create policy set %s for organization: %s", name, organization)
+	policySet, err := tfeClient.PolicySets.Create(ctx, organization, options)
+	if err != nil {
+		return fmt.Errorf(
+			"Error creating policy set %s for organization %s: %v", name, organization, err)
+	}
+
+	d.SetId(policySet.ID)
+
+	return resourceTFEPolicySetRead(d, meta)
+}
+
+func resourceTFEPolicySetRead(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	log.Printf("[DEBUG] Read policy set: %s", d.Id())
+	policySet, err := tfeClient.PolicySets.Read(ctx, d.Id())
+	if err != nil {
+		if err == tfe.ErrResourceNotFound {
+			log.Printf("[DEBUG] Policy set %s does no longer exist", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading policy set %s: %v", d.Id(), err)
+	}
+
+	// Update the config.
+	d.Set("name", policySet.Name)
+	d.Set("description", policySet.Description)
+	d.Set("global", policySet.Global)
+
+	if policySet.Organization != nil {
+		d.Set("organization", policySet.Organization.Name)
+	}
+
+	// Update the relationships.
+	var policyIDs []interface{}
+	for _, policy := range policySet.Policies {
+		policyIDs = append(policyIDs, policy.ID)
+	}
+	d.Set("policy_ids", policyIDs)
+
+	if !policySet.Global {
+		var workspaceIDs []interface{}
+		for _, workspace := range policySet.Workspaces {
+			workspaceIDs = append(workspaceIDs, workspace.ID)
+		}
+		d.Set("workspace_external_ids", workspaceIDs)
+	}
+
+	return nil
+}
+
+func resourceTFEPolicySetUpdate(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	name := d.Get("name").(string)
+	global := d.Get("global").(bool)
+
+	// If a user is setting the policy set to "global", make sure the workspaces
+	// that _had_ been set are explicitly removed. This helps keep the policy
+	// set's state in check
+	if d.HasChange("global") && global {
+		// The new set of workspaces will be an empty set, so we don't need it
+		oldSet, _ := d.GetChange("workspace_external_ids")
+		oldWorkspaceIDs := oldSet.(*schema.Set)
+
+		if oldWorkspaceIDs.Len() > 0 {
+			options := tfe.PolicySetDetachFromWorkspacesOptions{}
+
+			for _, workspaceID := range oldWorkspaceIDs.List() {
+				options.Workspaces = append(options.Workspaces, &tfe.Workspace{ID: workspaceID.(string)})
+			}
+
+			log.Printf("[DEBUG] Removing previous workspaces from now-global policy set: %s", d.Id())
+			err := tfeClient.PolicySets.DetachFromWorkspaces(ctx, d.Id(), options)
+			if err != nil {
+				return fmt.Errorf("Error detaching policy set %s from workspaces: %v", d.Id(), err)
+			}
+		}
+	}
+
+	// Don't bother updating the policy set's attributes if they haven't changed
+	if d.HasChange("name") || d.HasChange("description") || d.HasChange("global") {
+		// Create a new options struct.
+		options := tfe.PolicySetUpdateOptions{
+			Name:   tfe.String(name),
+			Global: tfe.Bool(global),
+		}
+
+		if desc, ok := d.GetOk("description"); ok {
+			options.Description = tfe.String(desc.(string))
+		}
+
+		log.Printf("[DEBUG] Update configuration for policy set: %s", d.Id())
+		_, err := tfeClient.PolicySets.Update(ctx, d.Id(), options)
+		if err != nil {
+			return fmt.Errorf(
+				"Error updating configuration for policy set %s: %v", d.Id(), err)
+		}
+	}
+
+	if d.HasChange("policy_ids") {
+		oldSet, newSet := d.GetChange("policy_ids")
+		oldPolicyIDs := oldSet.(*schema.Set).Difference(newSet.(*schema.Set))
+		newPolicyIDs := newSet.(*schema.Set).Difference(oldSet.(*schema.Set))
+
+		// First add the new policies.
+		if newPolicyIDs.Len() > 0 {
+			options := tfe.PolicySetAddPoliciesOptions{}
+
+			for _, policyID := range newPolicyIDs.List() {
+				options.Policies = append(options.Policies, &tfe.Policy{ID: policyID.(string)})
+			}
+
+			log.Printf("[DEBUG] Add policies to policy set: %s", d.Id())
+			err := tfeClient.PolicySets.AddPolicies(ctx, d.Id(), options)
+			if err != nil {
+				return fmt.Errorf("Error adding policies to policy set %s: %v", d.Id(), err)
+			}
+		}
+
+		// Then remove all the old policies.
+		if oldPolicyIDs.Len() > 0 {
+			options := tfe.PolicySetRemovePoliciesOptions{}
+
+			for _, policyID := range oldPolicyIDs.List() {
+				options.Policies = append(options.Policies, &tfe.Policy{ID: policyID.(string)})
+			}
+
+			log.Printf("[DEBUG] Remove policies from policy set: %s", d.Id())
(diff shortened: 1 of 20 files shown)


davidcelis:
NOTE: This depends on go-tfe#33. I made a vendor update for local development, but I'll amend that commit with the correct revisions once that PR is merged.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
svanharmelen: This should also be `Computed: true` as we read back the description in the read function. So if the description is set in the web UI it will be "computed" (aka updated) here as well.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+				Default:  false,
+			},
+
+			"organization": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"policies": &schema.Schema{
```
svanharmelen: This should be `policy_ids` right?

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"workspaces": &schema.Schema{
```
svanharmelen: This should be `workspace_external_ids` right? I added the `external` here because there is some unclarity what workspace ID to use in which cases. As the `workspace_id` in other resources refers to `<ORGANIZATION>/<WORKSPACE>` as all workspace related APIs expect the org and workspace name instead of the external ID.

But in here we _do_ expect the external ID right? So l hope this will make that more clear.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+	d.Set("global", policySet.Global)
+
+	// Update the relationships.
+	var policies []interface{}
+	for _, policy := range policySet.Policies {
+		policies = append(policies, policy.ID)
+	}
+	d.Set("policies", policies)
+
+	if !policySet.Global {
```
svanharmelen: Do we need to check this? I.e. does `policySet.Workspaces` contain any workspaces if its a global policy set? I would expect it to be empty and so the `range` op would be a no-op.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+		Description: tfe.String(description),
+		Global:      tfe.Bool(global),
+	}
+
+	// Set up the policies and workspaces.
+	for _, policyID := range policies.List() {
+		options.Policies = append(options.Policies, &tfe.Policy{ID: policyID.(string)})
+	}
+
+	if !*options.Global {
```
svanharmelen: Since you have `global` defined already, why not reuse that here to make it read just a bit nicer? But to be honest I think we don't need this check at all, as we already checked in the beginning of this function if `global` was `true` and `workspaces` were defined. So the `range` op should be a no-op here.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading policy set %s: %v", d.Id(), err)
+	}
+
+	// Update the config.
+	d.Set("name", policySet.Name)
+	d.Set("description", policySet.Description)
+	d.Set("global", policySet.Global)
```
svanharmelen: We should also update the organization name. It should not be possible to change that without recreating that, but it's still good to update the value to make sure it is still in sync.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+	name := d.Get("name").(string)
+	description := d.Get("description").(string)
+	global := d.Get("global").(bool)
+	organization := d.Get("organization").(string)
+	policies := d.Get("policies").(*schema.Set)
+	workspaces := d.Get("workspaces").(*schema.Set)
+
+	// First off, make sure they aren't trying to supply workspaces for a global set
+	if global && workspaces.Len() > 0 {
+		err := errors.New(`cannot set "workspaces" when "global" is true`)
```
svanharmelen: As you used this exact same error twice, you should probably declare it globally so the same error can be reused in both places.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+			log.Printf("[DEBUG] Remove policies from policy set: %s", d.Id())
+			err := tfeClient.PolicySets.RemovePolicies(ctx, d.Id(), options)
+			if err != nil {
+				return fmt.Errorf("Error removing policies from policy set %s: %v", d.Id(), err)
+			}
+		}
+	}
+
+	if !global && d.HasChange("workspaces") {
+		old, new := d.GetChange("workspaces")
```
svanharmelen: `new` is a keyword in Go which you should not use as a var name.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+		log.Printf("[DEBUG] Update configuration for policy set: %s", d.Id())
+		_, err := tfeClient.PolicySets.Update(ctx, d.Id(), options)
+		if err != nil {
+			return fmt.Errorf(
+				"Error updating configuration for policy set %s: %v", d.Id(), err)
+		}
+	}
+
+	if d.HasChange("policies") {
+		old, new := d.GetChange("policies")
```
svanharmelen: `new` is a keyword in Go which you should not use as a var name.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func resourceTFEPolicySetImporter(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	s := strings.SplitN(d.Id(), "/", 2)
+	if len(s) != 2 {
+		return nil, fmt.Errorf("invalid policy set import format: %s", d.Id())
+	}
+
+	// Set the fields that are part of the import ID.
+	d.Set("organization", s[0])
```
svanharmelen: I don't think this is needed here as a `PolicySet` struct contains a relation to the organization and so it can be retrieved from there. This is inline with the comment I left about reading the org name back in the read function and it would make this import function not needed anymore. Instead we can just use the policy set ID with the `schema.ImportStatePassthrough` importer.

website/docs/r/policy_set.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+* `id` - The ID of the policy set.
+
+## Import
+
+Policy sets can be imported by concatenating the `organization name` and
+`resource id`, e.g.
+
+```shell
+terraform import tfe_policy_set.test my-org-name/polset-wAs3zYmWAhYK7peR
+```
```
svanharmelen: There is a PR (#32) to update the import documentation just a tiny bit, so good to update this so that its inline with the upcoming changes.

website/docs/r/policy_set.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+
+Basic usage:
+
+```hcl
+resource "tfe_policy_set" "test" {
+  name = "my-policy-set"
+  description = "A brand new policy set"
+  global = false
+  organization = "my-org-name"
+
```
svanharmelen: Is this blank line on purpose? Usually we only add blank lines when we have a configuration block, so it shouldn't be needed here.

tfe/resource_tfe_sentinel_policy_test.go
```diff
@@ -197,6 +204,7 @@ resource "tfe_organization" "foobar" {
 
 resource "tfe_sentinel_policy" "foobar" {
   name = "policy-test"
+	description = "An updated test policy"
```
svanharmelen: There seems to be some tabs vs spaces misalignment here?

tfe/resource_tfe_sentinel_policy_test.go
```diff
@@ -184,6 +190,7 @@ resource "tfe_organization" "foobar" {
 
 resource "tfe_sentinel_policy" "foobar" {
   name = "policy-test"
+	description = "A test policy"
```
svanharmelen: There seems to be some tabs vs spaces misalignment here?

tfe/resource_tfe_sentinel_policy.go
```diff
@@ -27,6 +27,11 @@ func resourceTFESentinelPolicy() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
svanharmelen: This one should be `Computed: true` as well.

tfe/resource_tfe_sentinel_policy.go
```diff
@@ -125,23 +133,22 @@ func resourceTFESentinelPolicyRead(d *schema.ResourceData, meta interface{}) err
...
-		// Create a new options struct.
-		options := tfe.PolicyUpdateOptions{
-			Enforce: []*tfe.EnforcementOptions{
-				&tfe.EnforcementOptions{
-					Path: tfe.String(d.Get("name").(string) + ".sentinel"),
-					Mode: tfe.EnforcementMode(tfe.EnforcementLevel(d.Get("enforce_mode").(string))),
-				},
+	// Create a new options struct.
+	options := tfe.PolicyUpdateOptions{
+		Description: tfe.String(d.Get("description").(string)),
```
svanharmelen: Since `description` is an optional field we need to check if we got a value and so if we should set this. If not we would override any manual set descriptions here.

tfe/resource_tfe_sentinel_policy.go
```diff
@@ -125,23 +133,22 @@ func resourceTFESentinelPolicyRead(d *schema.ResourceData, meta interface{}) err
 func resourceTFESentinelPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
 	tfeClient := meta.(*tfe.Client)
 
-	if d.HasChange("enforce_mode") {
```
svanharmelen: Why did you remove this check?

tfe/resource_tfe_sentinel_policy.go
```diff
@@ -58,13 +63,15 @@ func resourceTFESentinelPolicy() *schema.Resource {
...
+	// Get the name, description and organization.
 	name := d.Get("name").(string)
+	description := d.Get("description").(string)
 	organization := d.Get("organization").(string)
 
 	// Create a new options struct.
 	options := tfe.PolicyCreateOptions{
-		Name: tfe.String(name),
+		Name:        tfe.String(name),
+		Description: tfe.String(description),
```
svanharmelen: Since this is an optional field, we need to check if we want to set this.

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+	if global && workspaces.Len() > 0 {
+		err := errors.New(`cannot set "workspaces" when "global" is true`)
+		return fmt.Errorf(
+			"Error creating policy set %s for organization %s: %v", name, organization, err)
+	}
+
+	// Create a new options struct.
+	options := tfe.PolicySetCreateOptions{
+		Name:        tfe.String(name),
+		Description: tfe.String(description),
```
svanharmelen: Since description is optional, we should check if we should set if (if its provided by the user).

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,300 @@
...
+			}
+		}
+	}
+
+	// Don't bother updating the policy set's attributes if they haven't changed
+	if d.HasChange("name") || d.HasChange("description") || d.HasChange("global") {
+		// Create a new options struct.
+		options := tfe.PolicySetUpdateOptions{
+			Name:        tfe.String(name),
+			Description: tfe.String(description),
```
svanharmelen: We should only set description if its provided.

tfe/resource_tfe_policy_set.go
davidcelis: The API will actually always return workspaces for a policy set. This is because global policy sets are still technically related to workspaces (in this case, all of the organization's workspaces) even if not having been _directly_ associated to workspaces by the user.

So this is still necessary, as we don't want to directly set this attribute in the provider for global policy sets

tfe/resource_tfe_policy_set.go
davidcelis: I could do this, but… I don't see anywhere else in the existing workspaces where we explicitly update the organization name; I figured it'd be consistent to also not do it here, but would you still like me to add it?

tfe/resource_tfe_sentinel_policy.go
davidcelis: hmm I'm not 100% sure what you mean by this… Do you mean that, if someone has set a description for a policy in the TFE UI but does not have a description set in their provider config, the provider config should leave it alone and not replace it with `nil` or an empty string?

tfe/resource_tfe_sentinel_policy.go
davidcelis: I'll add it back with the change for the comment above

davidcelis:
@svanharmelen I pushed up fixes for your feedback, so take another look when you get the chance 😄

tfe/resource_tfe_policy_set_test.go
```diff
@@ -0,0 +1,617 @@
...
+
+const testAccTFEPolicySet_basic = `
+resource "tfe_organization" "foobar" {
+  name = "terraform-test"
+  email = "admin@company.com"
+}
+
+resource "tfe_policy_set" "my_set" {
+  name = "my-set"
+	description = "A set for some of my Sentinel policies"
```
svanharmelen: Spaces/tabs? Or justy too many of either one of them :wink: Seems the same for all other `description` fields as well.

website/docs/r/policy_set.html.markdown
svanharmelen: PR #32 is merged, so please use the same format here (it was changed again just before it god merged).

tfe/resource_tfe_sentinel_policy.go
```diff
@@ -125,22 +136,29 @@ func resourceTFESentinelPolicyRead(d *schema.ResourceData, meta interface{}) err
...
 
-	if d.HasChange("enforce_mode") {
+	changeEnforce := d.HasChange("enforce_mode")
+	if d.HasChange("description") || changeEnforce {
 		// Create a new options struct.
-		options := tfe.PolicyUpdateOptions{
-			Enforce: []*tfe.EnforcementOptions{
+		options := tfe.PolicyUpdateOptions{}
+
+		if changeEnforce {
```
davidcelis: I'm checking this again inside of the existing conditional because TFE will always create a new policy config when it receives an `enforce` param. So although we want to submit this API request if either the description or enforce changed, I don't want to set `enforce` unless it has changed, because I don't want TFE to be needlessly creating additional policy configurations

svanharmelen: Yeah I understand 👍

tfe/resource_tfe_policy_set.go
```diff
@@ -0,0 +1,277 @@
...
+
+			log.Printf("[DEBUG] Remove policies from policy set: %s", d.Id())
+			err := tfeClient.PolicySets.RemovePolicies(ctx, d.Id(), options)
+			if err != nil {
+				return fmt.Errorf("Error removing policies from policy set %s: %v", d.Id(), err)
+			}
+		}
+	}
+
+	if !global && d.HasChange("workspace_external_ids") {
```
davidcelis: I know you had a question about this block and the above one, specifically with regards to why both `if` check blocks are needed… I'll try to walk you through why I've separated these two pieces out like this, and why the first check is also the first thing we do in this function. It largely has to do with the way the API works, as well as trying to avoid a specific scenario. Consider the following steps:

1. I have an organization that has five workspaces. `ws-123`, `ws-456`, `ws-789`, `ws-abc` and `ws-def`
2. I create a policy set with the default `global = false` and specify a couple of these workspaces as set members (`ws-123` and `ws-789`).
3. Later, I revisit this and decide to make the policy set global. I set `global = true` and remove the `workspace_external_ids` attribute entirely.

Without the top statement, we would:

1. Update the policy set itself (name/description/global flag)
2. Update the list of policies
3. And then we'd get to this block I've highlighted here and attempt to "update" the workspaces (in this case, that's just changed to an empty set so we would try to submit an API request that removes `ws-123` and `ws-789` from the set)

However, once a set has been flagged as global, the TFE API does not allow directly modifying the set's workspaces relationship; this API request would error out. So we have to handle the removal before we submit the request that actually flags the set as global.

I also can't move this block to the top in favor of the one already there, because it messes with what is essentially the reverse of the scenario I outlined at the start of this comment (i.e. changing from being global). For example: I remove the `global = true` statement from my policy set and then specify a few workspaces in the `workspace_external_ids` property. If we tried to just handle this in one combined check at the top, we would:

1. Detect the change to `workspace_external_ids` and attempt to submit an API request that adds the workspace
2. Get an error back from the API because we have not yet submitted the API request that unflags the set as global.

So we have to use a different order of operations depending on whether a set is changing global from `false => true` or from `true => false`.

I hope this long-winded comment made sense and adequately explained why everything is ordered this way 😅

svanharmelen: Thanks for the long and detailed explanation! I does raise another question, but I'll first merge this in and then have a look at the other thought this triggers.

Thanks for your work and patience!!

PR #42, opened 2019-01-03
Add support for managing OAuth Clients

### LINKED ISSUE #15: need resource to create VCS connection for organization.
It would be very helpful if the process of connecting a Version Control System to a TFE Organization was made simpler and automatic through the usage of a new Terraform Resource.

svanharmelen:
Closes #15

### DIFF
--- a/tfe/provider.go
+++ b/tfe/provider.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/url"
 	"os"
 
 	tfe "github.com/hashicorp/go-tfe"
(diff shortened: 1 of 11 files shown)


tfe/resource_tfe_oauth_client.go
```diff
@@ -0,0 +1,141 @@
...
+	d.Set("service_provider", string(oc.ServiceProvider))
+
+	switch len(oc.OAuthTokens) {
+	case 0:
+		d.Set("oauth_token_id", "")
+	case 1:
+		d.Set("oauth_token_id", oc.OAuthTokens[0].ID)
+	default:
+		return fmt.Errorf("Unexpected number of OAuth tokens: %d", len(oc.OAuthTokens))
+	}
```
tombuildsstuff: I'd argue we could just pick the first one if it exists here, rather than failing?

tfe/provider.go
```diff
@@ -16,10 +17,9 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 )
 
-const (
-	defaultHostname = "app.terraform.io"
-	serviceID       = "tfe.v2"
-)
+const defaultHostname = "app.terraform.io"
+
+var serviceIDs = []string{"tfe.v2", "tfe.v2.1"}
```
tombuildsstuff: given the function below pulls out the first value found, do we not want to start with the newest supported and work back?

Issue #19, opened 2018-09-14
Read-only workspace ID from the API add to Workspace resource


petems:
Right now, the ID given for a workspace from Terraform is `org-name|workspace-name`

However, if you want to use the API to do something, like trigger a run or such, you have to use the workspace ID as defined by the TFE API, which is an alphanumeric like `ws-oMQsEub2cEKmPBAs`:

```
{
  "data": {
    "id": "ws-mD5bmJ8ry3uTzuHi", #this bit
    "type": "workspaces",
    "attributes": {
      "name": "workspace-1",
      "environment": "default",
      "auto-apply": false,
      "locked": false,
      "created-at": "2018-03-08T22:30:00.404Z",
      "working-directory": null,
      "terraform-version": "0.11.3",
      "permissions": {
        "can-update": true,
        "can-destroy": true,
        "can-queue-destroy": true,
        "can-queue-run": true,
        "can-update-variable": true,
        "can-lock": true,
        "can-read-settings": true
      },
      "actions": {
        "is-destroyable": true
      }
    },
```

It would be good to have this as a read-only varible for resources, so it can be output on creation so it can be used outside of Terraform

svanharmelen:
The TFE API is a bit nasty with regards to the workspace API calls. If you want to retrieve or update the workspace, you need the organization and workspace name. But indeed some other API's that reference a workspace, do this by the external workspace ID.

Can you share the use case for this (why having this in the TF state is needed)? Of course we can add the external ID, but I'm a bit worried that people are going to be even more confused when they see that ID next to the value that is called `id` in the TF state 😏

petems:
Yeah `id` would be confusing... we can call it something else if needed. Like `workspace_id` or `external_id` or something 😅 

My usecase is the ability to trigger runs in TFE. I'm creating 500 workspaces with TFE, and want to trigger a plan in all of them. the only way to trigger plans is using the `ws` id. 

```
output "ids_of_all_workspaces" {
  value = ["${tfe_workspace.5mb_workspace.*.id}"]
}
```

but got the other ID... 😞 

There's no easy way of outputting all the ID's from Terraform for consumption with another tool (like Postman/Curl/Ruby script)

One solution could be to have a run/plan/trigger resource in the TFE provider, but I feel like run/plan triggering falls outside of the scope of the provider, since it's not a "State". It might even be a usecase for a provisioner almost?

svanharmelen:
Fixed by PR #21

PR #43, opened 2019-01-03
Add a couple of data sources

svanharmelen:
- tfe_ssh_key
- tfe_team
- tfe_team_access
- tfe_workspace
- tfe_workspace_ids

### DIFF
--- a/tfe/data_source_ssh_key.go
+++ b/tfe/data_source_ssh_key.go
@@ -1,0 +1,63 @@
+package tfe
+
+import (
+	"fmt"
+
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceTFESSHKey() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTFESSHKeyRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"organization": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func dataSourceTFESSHKeyRead(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	// Get the name and organization.
+	name := d.Get("name").(string)
+	organization := d.Get("organization").(string)
+
+	// Create an options struct.
+	options := tfe.SSHKeyListOptions{}
+
+	for {
+		l, err := tfeClient.SSHKeys.List(ctx, organization, options)
+		if err != nil {
+			return fmt.Errorf("Error retrieving SSH keys: %v", err)
+		}
+
+		for _, k := range l.Items {
+			if k.Name == name {
+				d.SetId(k.ID)
+				return nil
+			}
+		}
+
+		// Exit the loop when we've seen all pages.
+		if l.CurrentPage >= l.TotalPages {
+			break
+		}
+
+		// Update the page number to get the next page.
+		options.PageNumber = l.NextPage
+	}
+
+	return fmt.Errorf("Could not find SSH key %s/%s", organization, name)
+}
--- a/tfe/data_source_ssh_key.go
+++ b/tfe/data_source_ssh_key.go
@@ -15,13 +15,11 @@ func dataSourceTFESSHKey() *schema.Resource {
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
 			},
 
 			"organization": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
 			},
 		},
 	}
--- a/tfe/data_source_ssh_key_test.go
+++ b/tfe/data_source_ssh_key_test.go
@@ -1,0 +1,51 @@
+package tfe
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccTFESSHKeyDataSource_basic(t *testing.T) {
+	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckTFESSHKeyDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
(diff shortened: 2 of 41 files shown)


website/docs/d/workspace.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+# Data Source: tfe_workspace
+
+Use this data source to get information about a workspace.
+
+## Example Usage
+
+```hcl
+data "tfe_workspace" "test" {
+  name = "my-workspace-name"
+  organization = "my-org-name"
```
nfagerlund: Hmm, I'm a little concerned about this interface being unwieldy, but I'm not sure whether the thing I want is possible or not...

My use case for this is shown in [this example policies config](https://github.com/hashicorp/tfe-policies-example/blob/master/main.tf). Basically I have a bunch of policy set resources, and each one needs to provide a bunch of workspace IDs, which I'd rather access by name. But with per-workspace data sources, I'd need to write four extra lines of configuration for every name I want to use, and I'd still have the problem of double-bookkeeping, where I have to keep my set of `tfe_workspace` data sources up to date. That's not too onerous in that toy example, but in a real-life situation with 50+ workspaces instead of just 5, it's kind of brutal. 

Would it be possible to also make a `tfe_workspaces` or `tfe_workspace_ids` data source, which would have an `ids` attribute with a _map_ value? Or is that not feasible because of something about the way providers work?

svanharmelen: I think a data source can have an `ids` attribute which contains a `map` of workspace names and their external IDs. But in that case I think the question is what kind of filter you want to use to select which IDs should be added to the map?

If that filter requires you to specify all names of the workspaces you want to select, you end up with having to declare the the same list twice. Once in the `tfe_workspace_ids` data source and once in the `tfe_policy_set` resource. Yet I agree that this would still be better then the current solution for your use case.

But after looking at your example, I think there is a much better way to make this resource user friendly. I don't think we should mix and match workspace IDs in TFE resources. So instead of using the external IDs, we should probably just use the workspace names here.

Exporting the external ID of a workspace is only done for specific use cases (see #19 for more details), so that ID should itself not be used in other resources. I guess this one slipped through when reviewing the PR.

So I'll work on a PR to update the `tfe_policy_set` resource today so you can have a look at it when you get online. I think that would be a much better solution that using a data source with a map.

svanharmelen: Please checkout PR #44 and then mainly the updated docs as those will show the UX changes when using the resource: https://github.com/terraform-providers/terraform-provider-tfe/pull/44/files#diff-237abdf1e9ce57c21537937ad3e09eea

website/docs/d/workspace.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+* `organization` - (Required) Name of the organization.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The ID of the workspace within Terraform. This is a custom ID that is
+  needed because the Terraform Enterprise workspace related API calls require
+  the organization and workspace name instead of the actual workspace ID.
+* `auto_apply` - Indicated whether to automatically apply changes when a
```
bill-rich: Indicated -> Indicates

website/docs/d/workspace.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+  Terraform plan is successful.
+* `ssh_key_id` - The ID of an SSH key assigned to the workspace.
+* `queue_all_runs` - Indicated whether all runs should be queued.
```
bill-rich: Indicated -> Indicates

tfe/data_source_ssh_key.go
```diff
@@ -0,0 +1,63 @@
...
+
+func dataSourceTFESSHKey() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTFESSHKeyRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
bill-rich: No need for `ForceNew` on data sources.

tfe/data_source_ssh_key_test.go
```diff
@@ -0,0 +1,51 @@
...
+					resource.TestCheckResourceAttrSet("data.tfe_ssh_key.foobar", "id"),
+				),
+			},
+		},
+	})
+}
+
+func testAccTFESSHKeyDataSourceConfig(rInt int) string {
+	return fmt.Sprintf(`
+resource "tfe_organization" "foobar" {
```
bill-rich: The HCL in the tests and docs can be run through `terraform format` to make sure it is all consistent.

svanharmelen: Yeah I know, but I'm trying to avoid that as it means copy/pasting the content every time you work on a test or the docs. Not really a nice workflow IMO, but I will update them nevertheless.

tfe/data_source_ssh_key.go
svanharmelen: Removed `ForceNew` from all data sources.

website/docs/d/workspace.html.markdown
svanharmelen: 👍

website/docs/d/workspace.html.markdown
svanharmelen: 👍

tfe/data_source_ssh_key_test.go
svanharmelen: Yeah I know, she pointed me to that as well 👍

tfe/data_source_workspace.go
```diff
@@ -0,0 +1,103 @@
...
+				Computed: true,
+			},
+
+			"working_directory": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"vcs_repo": &schema.Schema{
+				Type:     schema.TypeSet,
```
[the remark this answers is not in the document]
svanharmelen: Will update it to a list 👍

Not sure if people need it, but since the info is available I think we should just return it. Pretty sure that if I remove it that within a few days or weeks someone comes to ask for it 😉

tfe/data_source_workspace_ids.go
```diff
@@ -0,0 +1,85 @@
...
+
+	options := tfe.WorkspaceListOptions{}
+	for {
+		wl, err := tfeClient.Workspaces.List(ctx, organization, options)
+		if err != nil {
+			return fmt.Errorf("Error retrieving workspaces: %v", err)
+		}
+
+		for _, w := range wl.Items {
+			if names["*"] || names[w.Name] {
```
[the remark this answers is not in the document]
svanharmelen: I would expect that a `tfe_organization` data source would provide details about the organization itself. Not about all resources (e.g. just about all other resources of this provider) that can be part of an organization.

Going down that road would probably open the doors to making `tfe_organziation` a one-size-fits-all data source to get all your teams, users, workspaces, variables enz. So I prefer to keep away from that and use this solution instead.

PR #48, opened 2019-01-10
Make it clear people should not use an org token

svanharmelen:


nfagerlund:
We talked about this in Slack, so to record some of that here:

- Based on our current stated rationale for why the org token exists, I don't think this restriction makes any sense. 
- We can resolve that by fixing the bug (which is apparently a non-starter for major customer reasons, alas), or by changing the stated purpose of the org token.
    - To do that, we need more clarity about what the org token is for. As best as I can tell, with the current restrictions it's only useful for **one-time setup operations at the beginning of a Terraform Enterprise organization's lifecycle,** and should not be used for any further management of TFE resources? Is that actually what we want to say about it?
    - Once we know that, we also need to change all the rest of the messaging about the token. Mostly the page about service accounts, the new guide to managing policy sets (which currently recommends using an org token, I think?), and I think the index page of the API docs.

Anyway, I would rather not merge this without a coherent story for why the org token exists and what exactly I need to do in order to not shoot myself in the foot with it. But if that turns out to take way too long, then I guess we can merge this first.

svanharmelen:
I'm looking at better solutions and gone down the rabbit hole to find out the exact customer restrictions in order to see/discuss if we can make a few tiny changes that would properly solve this.

So to be continued...

svanharmelen:
I'm going to close this as it looks like we can solve this issue on the API side after all 😃

PR #49, opened 2019-01-14
add 'plan' access level to tfe_team_access resource

hagurganus:
This enables the 'plan' access level on the tfe_team_access resource, which is mentioned as an option in the [TFE permissions documentation](https://www.terraform.io/docs/enterprise/users-teams-organizations/permissions.html).

I've confirmed that this access level is supported by the pTFE API (v201812-1) by setting some of the workspace access for teams to `plan` by making [the relevant API call](https://www.terraform.io/docs/enterprise/api/team-access.html#add-team-access-to-a-workspace).


hagurganus:
Hmmm... I'm not sure why that test would have failed, but I'll look into this.

svanharmelen:
@hagurganus the problem is that you updated a vendored dependency directly in the vendor directory, instead of first updating the dependency itself and then update the vendored version in this repo.

I just did those things for you, so if you rebase this PR on `master` I think you should be good and we can merge this PR 👍

svanharmelen:
LGTM! Thanks @hagurganus!

hagurganus:
Thanks for the feedback, @svanharmelen! I'm just getting spun up with contributing to TF providers, so your comments were immensely helpful. I look forward to contributing again in the future.

PR #51, opened 2019-01-16
Try to find a workspace by external ID before removing it

svanharmelen:
When a workspace name is changed out side of TF, we should still be able
to find it using it's external ID. So we first try that before we delete
the workspace from the state.

### DIFF
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -176,13 +176,48 @@ func resourceTFEWorkspaceRead(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Read configuration of workspace: %s", name)
 	workspace, err := tfeClient.Workspaces.Read(ctx, organization, name)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


svanharmelen:
@paddycarver I fully agree, but this one is kind of hard to test. The use case for this is when someone changes the name of the workspace outside of the use of TF. So given our current test harnas I'm not sure how I can/should test that?

The only thing I can come up with is a custom function that uses direct API calls to change the name as a test step. But that feels ... nasty. What do you think?

paddycarver:
> The only thing I can come up with is a custom function that uses direct API calls to change the name as a test step. But that feels ... nasty. What do you think?

That's generally how we do it in the Google provide, writing a custom resource.TestStep or TestCheck that manipulates the resource using direct API calls. YMMV.

svanharmelen:
Well in that case, I'll have look at it 👍 I do miss the test here myself as well, so if we already use that pattern elsewhere I'm fine with using it here as well...

svanharmelen:
@paddycarver just added a test, would be nice if you can double check it before I merge the PR.

PR #52, opened 2019-01-17
Add organization_access to Team resource

chrisarcand:
Also adds updating to teams, now that we have more than just a name that can be directly updated instead of recreating the whole team.

### DIFF
--- a/tfe/resource_tfe_team.go
+++ b/tfe/resource_tfe_team.go
@@ -13,6 +13,7 @@ func resourceTFETeam() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceTFETeamCreate,
 		Read:   resourceTFETeamRead,
+		Update: resourceTFETeamUpdate,
 		Delete: resourceTFETeamDelete,
 		Importer: &schema.ResourceImporter{
 			State: resourceTFETeamImporter,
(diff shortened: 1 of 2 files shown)


tfe/resource_tfe_team.go
```diff
@@ -22,14 +23,38 @@ func resourceTFETeam() *schema.Resource {
...
+						},
+
+						"manage_workspaces": &schema.Schema{
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+
+						"manage_vcs_settings": &schema.Schema{
+							Type:     schema.TypeBool,
+							Required: true,
```
chrisarcand: ℹ️ I think it makes sense to allow `organization_access` to be completely optional (all these would default to `false` on a team creation, then) but IF you specify it you must specify all three settings; I think it's weird to not be explicit as you're _partially_ depending on the server's default. Feels like you should go all-in or not. Pointing this out though, to ask if this is reasonable or weird given usual conventions.

svanharmelen: I don't this these should be required, right? They should be optional and I would give them a default that is inline with the defaults using in the UI when creating a new group (so I think the default should be `false`).

chrisarcand: See my comment above. Still think so? I can change if you like, but that was my thinking.

chrisarcand: I guess I don't care either way, defaulting is fine; I'll change it.

svanharmelen: Ah, we cross posted :wink: I don't see why that would be required? If I only want to give `manage_workspaces` access, then why should I need to fill in the other values as well?

To be clear, you do not depend on the servers default as the schema can define a default that would also be in the docs.

chrisarcand: > To be clear, you do not depend on the servers default as the schema can define a default that would also be in the docs.

Yeah that's what made me realize I really shouldn't care 😄Changed!

svanharmelen:
As discussed, we will not merge this PR right now due to expected authorization changes. So I'll close this PR for now and we can pick this up later this quarter when those changes are merged. Thx!

PR #53, opened 2019-01-17
User a list instead of a set for a workspace vcs_repo

svanharmelen:


chrisarcand:
✅  LGTM! (I do not have write access to this org/project)

PR #58, opened 2019-02-06
Update description to not be on a bullet point

codingdiaz:


svanharmelen:
Nice catch @codingdiaz! Thanks!

PR #65, opened 2019-02-20
Allow setting hostname via TFE_HOSTNAME ENV variable

radeksimko:
This addresses one part of https://github.com/terraform-providers/terraform-provider-tfe/issues/31 (just the hostname part, not credentials).

There doesn't seem to be _any_ way to set the hostname dynamically now, as opposed to credentials.

### DIFF
--- a/tfe/provider.go
+++ b/tfe/provider.go
@@ -50,7 +50,7 @@ func Provider() terraform.ResourceProvider {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Description: descriptions["hostname"],
-				Default:     defaultHostname,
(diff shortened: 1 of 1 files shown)


phinze:
Hi @svanharmelen! Just ran into this while trying to get the acceptance tests passing to test #76. Can you share a bit more about what changes need to be made beyond @radeksimko's tweak so I can shepherd this in?

svanharmelen:
As discussed offline, I think we can merge this now. But probably only after adding a `TFE_TOKEN` env var and updating the docs.

svanharmelen:
This is implemented and merged in PR #78

PR #66, opened 2019-02-20
test: Expose TFE_MEMBER_USERNAME & TFE_MEMBER_SECOND_USERNAME

radeksimko:


tfe/resource_tfe_team_members_test.go
```diff
@@ -200,10 +226,12 @@ resource "tfe_team" "foobar" {
...
 resource "tfe_team_members" "foobar" {
   team_id   = "${tfe_team.foobar.id}"
-  usernames = ["admin", "sander"]
-}`
+  usernames = ["admin", "%s"]
+}`, username)
+}
 
-const testAccTFETeamMembers_update = `
+func testAccTFETeamMembers_update(username string) string {
```
svanharmelen: I think making these functions which both take a `username` makes it a bit less clear what is being tested. Any reason why you prefer this format over following the same approach I used here: https://github.com/terraform-providers/terraform-provider-tfe/blob/master/tfe/resource_tfe_oauth_client_test.go#L106-L118

So use global variables (e.g. `TFE_USER1`, `TFE_USER2`) defined once, versus redefining them in every test that needs it? As I could see a use for these same vars in these tests as well: https://github.com/terraform-providers/terraform-provider-tfe/blob/master/tfe/resource_tfe_team_member_test.go#L12-L77

tfe/resource_tfe_team_members_test.go
```diff
@@ -2,23 +2,30 @@ package tfe
...
 
 	tfe "github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 )
 
 func TestAccTFETeamMembers_basic(t *testing.T) {
 	users := []*tfe.User{}
+	username := os.Getenv("TFE_MEMBER_USERNAME")
```
svanharmelen: I think I prefer something like `TFE_USER1` and `TFE_USER2` to keep it a bit shorter and to allow them to be used in other places as well.

tfe/resource_tfe_team_members_test.go
```diff
@@ -95,6 +114,12 @@ func TestAccTFETeamMembers_import(t *testing.T) {
 	})
 }
 
+func testCheckResourceAttrWithIndexesAddr(name, format string, idx *int, value string) resource.TestCheckFunc {
```
svanharmelen: As you defined this as a generic helper function, I doubt if it should be declared in this file? It will now be hidden away here, invisible to other people writing tests. So maybe we should place this in `provider_test.go` instead?

And I wondered why you want a `*int` here? As you don't test for it being `nil`, why not just use and `int`?

tfe/resource_tfe_team_members_test.go
radeksimko: Sure, not a problem at all - I will rename these variables. 👍

tfe/resource_tfe_team_members_test.go
radeksimko: I actually found a better/cleaner way to do this. The `*int` is just a leftover of my poor copy-pasta skill. I will remove this function altogether.

tfe/resource_tfe_team_members_test.go
radeksimko: > I think making these functions which both take a `username` makes it a bit less clear what is being tested.

You're right. I'm thinking about removing `testAccTFETeamMembers_update` completely and changing the interface of the first function to this

```go
func testAccTFETeamMembers_basic(usernames []string) string
```

that way it's more obvious from the body of the test that we're changing one username as the subject of test (list of usernames) is more explicit and visible. Also we can reduce LOC that way as the same function can be used for both stages of the test (creation & update).

> Any reason why you prefer this format over following the same approach I used here

So the "function with variables"-approach is used throughout the AWS provider and many other major providers. The main reason being that it can scale (the same config can be reused anywhere) and provides compile-time validation (as opposed to plain `fmt.Sprintf` which would just render `%(missing variable)` missing variables or `%(incorrect type)`for wrong order/type).

Lastly passing a variable to function call makes it more obvious what's being tested in the place where it's being tested (test body). i.e. I don't need to look at the whole template to understand that a username is (a) even configurable and (b) where does it come from.

That said I'd be open to adding a helper function to the SDK (at some point) so we can source such variables without having to import `os`, but I'm not sure if there's much benefit, other than swapping one import for another. We'd probably need to figure out how to couple it more tightly with `testAccCheckEnvVariables` (which should also be native part of the SDK).

> As I could see a use for these same vars in these tests as well

That's fine because these tests don't seem to be calling an API, so it's safe to keep your username hardcoded there as we're not constrained by its real existence in TFE IMO?

svanharmelen: > You're right. I'm thinking about removing `testAccTFETeamMembers_update` completely and changing the interface of the first function to this

I think that would make it even less clear to be honest 😂 And I think you wouldn't need a slice of usernames in there right? The body only uses a single username variable.

As for the rest of this discussion, I'm fine with changing it to a function. I'm just (overly) sensitive to inconsistencies. So currently this provider only uses `consts` or `vars` that contain the configs. Introducing this new approach versus reusing the existing approach makes me shiffer a bit and makes me want to change all the tests inputs to functions so its consistent again 😉

radeksimko: > And I think you wouldn't need a slice of usernames in there right? The body only uses a single username variable.

Well I'd make all usernames configurable, i.e. I'd pull out `admin` too.

> I'm fine with changing it to a function. I'm just (overly) sensitive to inconsistencies.

So I'm happy to retain consistency within the TFE provider and change other `const`s to `func`s, but I just learned to live in a chaotic inconsistent world where things are moving, mainly because maintaining consistency effectively is hard in all the big providers. 😂

svanharmelen: > but I just learned to live in a chaotic inconsistent world

Guess I'm not quite there yet.. Still trying to hold on 😂

radeksimko:
@svanharmelen PTAL. I made all the changes.
I can send a separate PR changing all the other `const`s to `func`s once you approve this one.

svanharmelen:
Never mind making a PR for updating all the tests! I'll be working on this anyway, so if it keeps itching I'll update then in a later PR.

PR #72, opened 2019-03-22
Bump go-tfe to 0.3.13

justincampbell:
Fixes https://github.com/terraform-providers/terraform-provider-tfe/issues/71


sudomateo:
Do we need to change the `value` attribute from required to optional?
https://github.com/terraform-providers/terraform-provider-tfe/blob/master/tfe/resource_tfe_variable.go#L29-L33

justincampbell:
@sudomateo I think if it's `Required` then an empty string `""` would still be allowed. Is there a reason to allow it to be omitted other than for API consistency?

sudomateo:
@justincampbell That makes sense to me. Leaving it required would mean the value would either be `""` or some non-empty string. I don't have any other reason besides API consistency.

svanharmelen:
> I think if it's Required then an empty string "" would still be allowed. Is there a reason to allow it to be omitted other than for API consistency?

I prefer API consistency, so please make the `Value` field optional and default to `""` to express it has the exact same behavior as the API (and the docs).

justincampbell:
@svanharmelen Updated to 0.3.13 and marked value as optional

PR #75, opened 2019-04-10
Make ssh_key_id in workspace not computed

mikenomitch:
- Removes `Computed: true` from ssh_key_id definition in schema
- Adds `Default: ""` to ssh_key_id definition in schema
- Adds test
- Fixes link in documentation


mikenomitch:
I'm not super confident that this is what we want just yet, so let's sync up before merging this!

mikenomitch:
@svanharmelen Anything else you think this needs? Lgtm 👍

PR #76, opened 2019-05-22
Update to Terraform v0.12 SDK

phinze:
This should add v0.12 support for this provider - if the acceptance tests pass!


raizyr:
:shipit:

phinze:
Local acceptance tests are passing! Will ask for one more quick review and then this is good to go.

PR #78, opened 2019-06-04
Add support for TFE_HOSTNAME and TFE_TOKEN environment variables

alisdair:
Follow-up to @radeksimko's #65 as a separate PR from my fork because I don't have push access.

@svanharmelen, is this what you intended?

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,5 +69,7 @@ $ make testacc
 
 Testing
 -------
-A hostname and token must be provided in order to run the acceptance tests. We recomment configuring
+A hostname and token must be provided in order to run the acceptance tests. We recommend configuring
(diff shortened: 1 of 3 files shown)


alisdair:
I got the acceptance tests to pass locally on this branch and updated the README with some extended documentation on how to run them.

svanharmelen:
Looks good, but maybe we should update this as well: https://github.com/terraform-providers/terraform-provider-tfe/blob/master/website/docs/index.html.markdown

alisdair:
> Looks good, but maybe we should update this as well: https://github.com/terraform-providers/terraform-provider-tfe/blob/master/website/docs/index.html.markdown

Agreed, thanks for the pointer. Done in b84a576.

PR #80, opened 2019-06-14
[WIP] r/policy_set: add support for VCS policy sets

vancluever:
This adds the necessary changes to allow for VCS policy sets.

VCS-backed policy sets are defined by supplying the vcs_repo - and
optionally, policies_path, arguments to the tfe_policy_set resource.
This is in lieu of policy_ids.

Changing either vcs_repo or policies_path will result in a new
resource being forced.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/terraform-providers/terraform-provider-tfe
 
 require (
 	github.com/hashicorp/go-hclog v0.0.0-20190109152822-4783caec6f2e // indirect
-	github.com/hashicorp/go-tfe v0.3.17
+	github.com/hashicorp/go-tfe v0.3.18
 	github.com/hashicorp/go-version v1.1.0
 	github.com/hashicorp/hcl v0.0.0-20180404174102-ef8a98b0bbce
 	github.com/hashicorp/terraform v0.12.0
--- a/go.sum
+++ b/go.sum
@@ -170,8 +170,8 @@ github.com/hashicorp/go-slug v0.3.0/go.mod h1:I5tq5Lv0E2xcNXNkmx7BSfzi1PsJ2cNjs3
 github.com/hashicorp/go-sockaddr v0.0.0-20180320115054-6d291a969b86/go.mod h1:7Xibr9yA9JjQq1JpNB2Vw7kxv8xerXegt+ozgdvDeDU=
 github.com/hashicorp/go-tfe v0.3.16 h1:GS2yv580p0co4j3FBVaC6Zahd9mxdCGehhJ0qqzFMH0=
 github.com/hashicorp/go-tfe v0.3.16/go.mod h1:SuPHR+OcxvzBZNye7nGPfwZTEyd3rWPfLVbCgyZPezM=
-github.com/hashicorp/go-tfe v0.3.17 h1:4n4Z1Kr5hXKI2ntJLyso9yYFHjSBVYRGgK8BG48H6OQ=
-github.com/hashicorp/go-tfe v0.3.17/go.mod h1:SuPHR+OcxvzBZNye7nGPfwZTEyd3rWPfLVbCgyZPezM=
+github.com/hashicorp/go-tfe v0.3.18 h1:Rhiq89MbHE1J5/fRnJBjBhh0cXykE0/hPvmpY3ra1Wk=
+github.com/hashicorp/go-tfe v0.3.18/go.mod h1:SuPHR+OcxvzBZNye7nGPfwZTEyd3rWPfLVbCgyZPezM=
 github.com/hashicorp/go-uuid v1.0.0 h1:RS8zrF7PhGwyNPOtxSClXXj9HA8feRnJzgnI1RJCSnM=
 github.com/hashicorp/go-uuid v1.0.0/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/bN7x4byOro=
 github.com/hashicorp/go-uuid v1.0.1 h1:fv1ep09latC32wFoVwnqcnKJGnMSdBanPczbHAYm1BE=
--- a/tfe/resource_tfe_policy_set.go
+++ b/tfe/resource_tfe_policy_set.go
@@ -43,6 +43,20 @@ func resourceTFEPolicySet() *schema.Resource {
 				ConflictsWith: []string{"workspace_external_ids"},
 			},
 
+			"policies_path": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"policy_ids"},
+			},
+
+			"policy_ids": {
+				Type:          schema.TypeSet,
+				Optional:      true,
+				Elem:          &schema.Schema{Type: schema.TypeString},
+				ConflictsWith: []string{"vcs_repo", "policies_path"},
+			},
+
 			"vcs_repo": {
 				Type:          schema.TypeList,
 				Optional:      true,
@@ -76,20 +90,6 @@ func resourceTFEPolicySet() *schema.Resource {
 				},
 			},
 
-			"policies_path": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				ForceNew:      true,
-				ConflictsWith: []string{"policy_ids"},
-			},
-
-			"policy_ids": {
-				Type:          schema.TypeSet,
-				Optional:      true,
-				Elem:          &schema.Schema{Type: schema.TypeString},
-				ConflictsWith: []string{"vcs_repo", "policies_path"},
-			},
-
 			"workspace_external_ids": {
 				Type:          schema.TypeSet,
 				Optional:      true,
@@ -117,6 +117,14 @@ func resourceTFEPolicySetCreate(d *schema.ResourceData, meta interface{}) error 
 		options.Description = tfe.String(desc.(string))
 	}
 
+	if policiesPath, ok := d.GetOk("policies_path"); ok {
+		options.PoliciesPath = tfe.String(policiesPath.(string))
+	}
+
+	for _, policyID := range d.Get("policy_ids").(*schema.Set).List() {
+		options.Policies = append(options.Policies, &tfe.Policy{ID: policyID.(string)})
+	}
+
 	// Get and assert the VCS repo configuration block.
 	if v, ok := d.GetOk("vcs_repo"); ok {
 		vcsRepo := v.([]interface{})[0].(map[string]interface{})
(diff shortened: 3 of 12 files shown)


vancluever:
Depends on hashicorp/go-tfe#77.

tfe/resource_tfe_policy_set.go
```diff
@@ -43,10 +45,50 @@ func resourceTFEPolicySet() *schema.Resource {
 				ConflictsWith: []string{"workspace_external_ids"},
 			},
 
+			"vcs_repo": {
+				Type:          schema.TypeList,
```
davidcelis: `TypeList` tripped me up but judging from the other resources, that's right!

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/terraform-providers/terraform-provider-tfe
 
 require (
 	github.com/hashicorp/go-hclog v0.0.0-20190109152822-4783caec6f2e // indirect
-	github.com/hashicorp/go-tfe v0.3.17
+	github.com/hashicorp/go-tfe v0.3.18-0.20190613152246-ce09432e1250
```
svanharmelen: We really should make sure we only use released versions in here.

go.sum
```diff
@@ -172,6 +172,8 @@ github.com/hashicorp/go-tfe v0.3.16 h1:GS2yv580p0co4j3FBVaC6Zahd9mxdCGehhJ0qqzFM
 github.com/hashicorp/go-tfe v0.3.16/go.mod h1:SuPHR+OcxvzBZNye7nGPfwZTEyd3rWPfLVbCgyZPezM=
 github.com/hashicorp/go-tfe v0.3.17 h1:4n4Z1Kr5hXKI2ntJLyso9yYFHjSBVYRGgK8BG48H6OQ=
 github.com/hashicorp/go-tfe v0.3.17/go.mod h1:SuPHR+OcxvzBZNye7nGPfwZTEyd3rWPfLVbCgyZPezM=
+github.com/hashicorp/go-tfe v0.3.18-0.20190613152246-ce09432e1250 h1:3tzwJ9TtTVHbBLnYZj+QJQlTkx7hic4QlmO4j4ofWCs=
+github.com/hashicorp/go-tfe v0.3.18-0.20190613152246-ce09432e1250/go.mod h1:SuPHR+OcxvzBZNye7nGPfwZTEyd3rWPfLVbCgyZPezM=
```
svanharmelen: Make sure you execute `go mod tidy` as well to clean up older versions.

tfe/resource_tfe_policy_set.go
svanharmelen: I switched it from `TypeSet` to `TypeList` a few months ago. I would need to search Asana (which I can't anymore :wink:) to recall why we needed to make that change, but it do remember was in order to fix or improve something.

tfe/resource_tfe_policy_set.go
```diff
@@ -43,10 +45,50 @@ func resourceTFEPolicySet() *schema.Resource {
...
+							Default:  false,
+						},
+
+						"oauth_token_id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+			},
```
svanharmelen: Something about the ordering 🤐

tfe/resource_tfe_policy_set.go
```diff
@@ -76,6 +118,19 @@ func resourceTFEPolicySetCreate(d *schema.ResourceData, meta interface{}) error
...
 	}
 
+	if vcsOpts := d.Get("vcs_repo").([]interface{}); len(vcsOpts) > 0 {
+		options.VCSRepo = &tfe.VCSRepoOptions{
+			Branch:            tfe.String(vcsOpts[0].(map[string]interface{})["branch"].(string)),
+			Identifier:        tfe.String(vcsOpts[0].(map[string]interface{})["identifier"].(string)),
+			IngressSubmodules: tfe.Bool(vcsOpts[0].(map[string]interface{})["ingress_submodules"].(bool)),
+			OAuthTokenID:      tfe.String(vcsOpts[0].(map[string]interface{})["oauth_token_id"].(string)),
+		}
+	}
```
svanharmelen: I think it would be nice to reuse the same logic used [here](https://github.com/terraform-providers/terraform-provider-tfe/blob/master/tfe/resource_tfe_workspace.go#L145-L159). Partly because it does an extra check for the `branch` and also because (IMHO) it looks a little cleaner.

tfe/resource_tfe_policy_set.go
```diff
@@ -119,6 +174,31 @@ func resourceTFEPolicySetRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("organization", policySet.Organization.Name)
 	}
 
+	// Set VCS policy set options.
+	d.Set("policies_path", policySet.PoliciesPath)
```
svanharmelen: We should probably move this one to line 172 as it's just a simple assignment like the others listed there.

tfe/resource_tfe_policy_set.go
```diff
@@ -275,3 +355,36 @@ func resourceTFEPolicySetDelete(d *schema.ResourceData, meta interface{}) error
...
+	}
+
+	// Ensure that policies_path is *not* supplied when vcs_repo is
+	// undefined.
+	if !vcsOk && pathOk {
+		return errors.New("policies_path must not be supplied when vcs_repo is empty")
+	}
+
+	return nil
+}
```
svanharmelen: Can you explain why we need this custom logic here? Is there no way we can achieve the same using by configuring the schema correctly?

tfe/resource_tfe_policy_set_test.go
```diff
@@ -465,3 +541,88 @@ resource "tfe_policy_set" "foobar" {
...
+  default = "%s"
+}
+
+variable "vcs_branch" {
+  default = "%s"
+}
+
+variable "vcs_policies_path" {
+  default = "%s"
+}
```
svanharmelen: Let's not make the tests more complicated as they need to and just interpolate the values where we need them instead of adding variables. In these tests this doesn't add value other that making the test code longer.

tfe/resource_tfe_policy_set_test.go
```diff
@@ -222,6 +224,80 @@ func TestAccTFEPolicySetImport(t *testing.T) {
...
+				t.Skip("please set TFE_VCS_IDENTIFIER to run this test")
+			}
+
+			if v := os.Getenv("TFE_POLICY_SET_VCS_BRANCH"); v == "" {
+				t.Skip("please set TFE_POLICY_SET_VCS_BRANCH to run this test")
+			}
+
+			if v := os.Getenv("TFE_POLICY_SET_VCS_PATH"); v == "" {
+				t.Skip("please set TFE_POLICY_SET_VCS_PATH to run this test")
+			}
```
svanharmelen: I think it would be cleaner to use the same approach we used for the `GITHUB_TOKEN` env var. We define it in `provider_test.go` so that all these env vars are described at one location. And then the tests here can be updated to test against the Go vars.

website/docs/r/policy_set.html.markdown
```diff
@@ -40,10 +60,33 @@ The following arguments are supported:
...
-* `policy_ids` - (Required) A list of Sentinel policy IDs.
 * `workspace_external_ids` - (Optional) A list of workspace external IDs. If
   the policy set is `global`, this value _must not_ be provided.
+* `vcs_repo` - The [VCS repository settings](#vcs-repository-settings) for this
+  policy set. Forces a new resource if changed.
+* `policies_path` - (Optional) The sub-path within the attached VCS repository
+  to ingress when using `vcs_repo`. All files and directories outside of this
+  sub-path will be ignored. This option can only be supplied when `vcs_repo` is
+  present. Forces a new resource if changed.
+* `policy_ids` - A list of Sentinel policy IDs.
```
svanharmelen: Something about ordering 😶

go.mod
vancluever: This is a WIP (hence the heading), so this will be all cleaned up after hashicorp/go-tfe#77 is merged.

go.sum
vancluever: Thanks, this is a WIP right now as the SDK is not fully merged, but I will remember to do that afterwards. I usually do, but forgot this time.

tfe/resource_tfe_policy_set.go
vancluever: That's fine. It should be `TypeList` anyway unless we need to avoid duplicates. I have been avoiding sets in that case as well too to avoid the side-effects of hashing and the kind of un-intuitive nature of the duplicate squishing with sets (in addition to how fragile it is really, given that the hashing implies an exact match).

There is a new block type coming to TF 0.12 that will allow you to specify blocks by key, which is what I will probably be favoring for cases where duplicates need to be avoided.

tfe/resource_tfe_policy_set.go
vancluever: No, we don't allow mutual inclusivity in the SDK right now individual validations only operate at the individual attribute level, without the ability to access other fields.

tfe/resource_tfe_policy_set_test.go
vancluever: Not 100% true:

* At a general level, adding verbs as variables allows the verbs to be passed through `terraform fmt -` without risk of failures. This was more a TF 0.11 and earlier thing, but it's worth noting.
* If referenced more than once, it reduces stutter in the Sprintf generation.

I'll change this though as this is not necessarily the case here.

tfe/resource_tfe_policy_set_test.go
vancluever: I would rather that this test be skipped on `make testacc` versus failed which is what the general pre-check function does. While (I'm assuming) GITHUB_TOKEN is generally needed to run all of the tests, someone may want to run the test suite without a VCS repository set up.

Note that this is also the first test that requires VCS access - we don't seem to be doing any workspace VCS tests. So adding these checks to fail would break any existing tests until the appropriate (live) repository fixture was added.

website/docs/r/policy_set.html.markdown
vancluever: Ordering? Do you mean alphabetical ordering? Not too sure what you're getting at here.

tfe/resource_tfe_policy_set.go
vancluever: I'm kind of wondering if, based on the outcome of https://github.com/hashicorp/go-tfe/pull/77#issuecomment-502185845, if we even need any of these checks. We could always just have the resource default to creating an empty set without VCS if none of the fields are specified. I can also test to see what happens if you do specify a `policies_path` field without VCS repo options - the docs imply that it fails, but if it ignores it, we can just do that too. We could even just omit it from the API request for legacy sets on our side anyway and just remove this code, and indicate in the docs that it's ignored.

vancluever:
I think the changes here should comprise most of the final changes necessary to put this through. 

@svanharmelen I'd still like to hear your thoughts/response of the ordering of the options. I'll defer to your judgement here.

tfe/resource_tfe_policy_set.go
vancluever: I've gone and removed the default. This gets rid of the need for the validation during diff customization.

I've re-purposed one of the tests to make sure that it does the expected default behavior, in addition to another one that validates that going from a populated set to an empty set via default behavior is possible.

svanharmelen:
@vancluever I made a few small changes to make sure all logic is consistent (mainly the tests). It would be great if you could review and run the tests one more time to make sure they all pass. If they do, this should be good to be merged and released 👍

vancluever:
Looks good, this all passes. I will be merging shortly!

PR #81, opened 2019-06-26
Ensure that changes are applied for monorepos

alisdair:
The monorepo fields were not being checked for changes, so if you ran an apply where only file triggers or trigger prefixes changed, it would not persist your data.

Includes a test which only updates "file triggers enabled"; this test fails without the code change.

### DIFF
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -303,7 +303,8 @@ func resourceTFEWorkspaceUpdate(d *schema.ResourceData, meta interface{}) error 
 	}
 
 	if d.HasChange("name") || d.HasChange("auto_apply") || d.HasChange("queue_all_runs") ||
-		d.HasChange("terraform_version") || d.HasChange("working_directory") || d.HasChange("vcs_repo") {
+		d.HasChange("terraform_version") || d.HasChange("working_directory") || d.HasChange("vcs_repo") ||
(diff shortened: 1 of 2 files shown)


tfe/resource_tfe_workspace.go
```diff
@@ -303,7 +303,8 @@ func resourceTFEWorkspaceUpdate(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if d.HasChange("name") || d.HasChange("auto_apply") || d.HasChange("queue_all_runs") ||
-		d.HasChange("terraform_version") || d.HasChange("working_directory") || d.HasChange("vcs_repo") {
+		d.HasChange("terraform_version") || d.HasChange("working_directory") || d.HasChange("vcs_repo") ||
+		d.HasChange("file_triggers_enabled") || d.HasChange("trigger_prefixes") {
```
svanharmelen: I wonder if it makes more sense to write this check as `if !d.HasChange(“ssh_key_id”)`?

If this method gets called it means something changed, and the only special case seems to be the SSH key stuff.

alisdair: I think that would prevent updates to any other fields at the same time as SSH key changes, which seems incorrect.

svanharmelen: Ah yes... You are correct on that one 👍🏻I'll silence myself 😉

PR #86, opened 2019-08-08
Add support for notification configuration resource

### LINKED ISSUE #68: Please add ‘notification_configuration’(?) resource
With the release of the notifications feature for TFE on 3/1, and the corresponding [Notification Configurations](https://www.terraform.io/docs/enterprise/api/notification-configurations.htm) API, it’d be nice to see that added as a resource in the TFE provider.

I may endeavor to do this myself, but I thought I’d add a ‘feature request’ here in the Issues tracker in case anyone else is considering doing the same work.

lafentres:
## Description

This adds support and documentation for creating, updating, destroying, and importing notification configurations. It resolves #68.

#### Notification configuration resources require:
1. name
1. destination_type - one of slack or generic
1. url - url the notifications will go to
1. workspace_external_id - external id of the workspace this notification configuration belongs to

#### Optionally, you can provide:
1. enabled - true or false, defaults to false
1. token - key used for providing an HMAC signature on notification requests, only allowed with destination_type of generic
1. triggers - set of triggers, defaults to empty set (no notification trigger on any run events)

Updating destination_type or workspace_external_id will force the creation of a new notification configuration resource. 

## Testing
The examples from the documentation I added or from the test cases work and you can use them to generate notification configurations if you pull this down and use it locally. 

## Documentation
* [Notification configuration API docs](https://www.terraform.io/docs/cloud/api/notification-configurations.html)
* [go-tfe notification configuration docs](https://godoc.org/github.com/hashicorp/go-tfe#NotificationConfiguration)


lafentres:
I might have been a little overzealous with the tests, so please let me know if there are tests that are unnecessary. Or if there are more tests that need to be added :)

tfe/resource_tfe_notification_configuration.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		return fmt.Errorf("Error reading notification configuration %s: %v", d.Id(), err)
+	}
+
+	// Update config
+	d.Set("destination_type", notificationConfiguration.DestinationType)
+	d.Set("enabled", notificationConfiguration.Enabled)
+	d.Set("name", notificationConfiguration.Name)
+	// Don't set token here, as it is write only
+	// and setting it here would make it blank
```
davidcelis: Thank you for adding this comment! I noticed that resources like `OrganizationToken` and `TeamToken` also don't set the actual token value, but with no comment as to why.

tfe/resource_tfe_notification_configuration_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+func TestAccTFENotificationConfiguration_wrongDestinationType(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckTFENotificationConfigurationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccTFENotificationConfiguration_wrongDestinationType,
+				ExpectError: regexp.MustCompile(``),
```
davidcelis: Should there be an expected error message here instead of an empty string?

tfe/resource_tfe_notification_configuration_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+func TestAccTFENotificationConfiguration_wrongTriggers(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckTFENotificationConfigurationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccTFENotificationConfiguration_wrongTriggers,
+				ExpectError: regexp.MustCompile(``),
```
davidcelis: Same question here ☝️

tfe/resource_tfe_notification_configuration.go
lafentres: I pieced this together after a lot of trial and error. It would be nice to get confirmation from someone who has done a lot of work with providers though.

tfe/resource_tfe_notification_configuration_test.go
lafentres: This one and the other test you commented on are potentially test cases I don't need to cover. I'm not sure about them. 

Basically I wanted to test that the validate functions I set on in the schema actually work. But since they are built-in terraform helpers, they have their own error messages that I don't control. I thought it might be bad if I wrote tests that broke every time those error messages changed. But maybe this is a sign I don't need to test these?

davidcelis: @lafentres Great point. Looking in the other test files, I don't really see anybody writing tests around these sorts of built-in validations. I can see the argument for doing so, just to verify that the validators are set up in the first place, but without that convention already in place I do think you could remove these two tests!

tfe/resource_tfe_notification_configuration_test.go
```diff
@@ -0,0 +1,436 @@
...
+					// Values in triggers attribute are tested by testCheckTFENotificationConfigurationAttributes
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "triggers.#", "0"),
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "url", "http://example.com"),
+				),
+			},
+			{
+				ResourceName:            "tfe_notification_configuration.foobar",
+				ImportState:             true,
```
vancluever: The import test here should be spun off into another test, unless it needs to be in basic for some reason?

tfe/resource_tfe_notification_configuration_test.go
```diff
@@ -0,0 +1,436 @@
...
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "triggers.#", "0"),
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "url", "http://example.com"),
+				),
+			},
+
+			{
+				ResourceName:            "tfe_notification_configuration.foobar",
+				ImportState:             true,
```
vancluever: As with basic - I'd remove the import test unless it's needed for something. Generally it's not needed for update tests because you're just testing that the resource can be imported from scratch.

tfe/resource_tfe_notification_configuration_test.go
vancluever: I have a counterpoint! 😛 

I think that testing on error messages _is_ valuable, as it tests to make sure that you are getting the correct error. Validation errors are great to test on too if you know for sure you're expecting a certain validation to fail. I'd say add them, along with the error message.

`helper/validation` is pretty stable and you can be sure that the error messages won't change often - if at all. If tests break because of a change, you will know for sure and can update the expected errors accordingly.

tfe/resource_tfe_notification_configuration_test.go
```diff
@@ -0,0 +1,436 @@
...
+					// Values in triggers attribute are tested by testCheckTFENotificationConfigurationAttributes
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "triggers.#", "1"),
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "url", "http://example.com"),
+				),
+			},
+			{
+				ResourceName:            "tfe_notification_configuration.foobar",
+				ImportState:             true,
```
vancluever: Same comment re: import tests.

tfe/resource_tfe_notification_configuration_test.go
```diff
@@ -0,0 +1,436 @@
...
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No instance ID is set")
+		}
+
+		nc, err := tfeClient.NotificationConfigurations.Read(ctx, rs.Primary.ID)
+		if err != nil {
+			return err
+		}
+
+		*notificationConfiguration = *nc
```
vancluever: Let's try and avoid assignment of values in check functions like this that are designed to do something else - it kind of hides the intent.

Usually, if I need something like this, I add another function to do it, or I just write an inline anonymous function.

Looking at how it's used though, you could probably just move all of the logic that reads in `notificationConfiguration` into `testAccCheckTFENotificationConfigurationAttributes`, et al, as that check function will also get the state and you can fetch the ID to pass into `tfeClient.NotificationConfigurations.Read` from there. You could also write a helper to get the data if you are worried about duplication.

tfe/resource_tfe_notification_configuration_test.go
```diff
@@ -0,0 +1,436 @@
...
+resource "tfe_workspace" "foobar" {
+  name         = "workspace-test"
+  organization = "${tfe_organization.foobar.id}"
+}
+
+resource "tfe_notification_configuration" "foobar" {
+  name             			= "notification_basic"
+  destination_type 			= "generic"
+  url              			= "http://example.com"
+  workspace_external_id     = "${tfe_workspace.foobar.external_id}"
```
vancluever: Might want to filter your test fixtures through `terraform fmt`.

website/docs/r/notification_configuration.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_notification_configuration"
+sidebar_current: "docs-resource-tfe-notification-configuration"
+description: |-
+  Manages notifications configurations.
+---
+
+# tfe_notification_configuration
+
+Terraform Cloud can be configured to send notifications for run state transitions.
```
vancluever: _Very minor, (almost) ignore_ 

Most Terraform resource/data source pages start with something along the lines of:

> The `tfe_notification_configuration` resource can be used to manage [notification configurations](https://www.terraform.io/docs/cloud/api/notification-configurations.html) within Terraform Cloud.

That doesn't look like that's something that's happening in this provider though, so I wouldn't worry about it for now. 😛

website/docs/r/notification_configuration.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+* `name` - (Required) Name of the notification configuration.
+* `destination_type` - (Required) The type of notification configuration payload to send. 
+  Valid values are `generic` or `slack`.
+* `enabled` - (Optional) Whether the notification configuration should be enabled or not.
+  Disabled configurations will not send any notifications. Defaults to `false`.
+* `token` - (Optional) A write-only secure token for the notification configuration, which can
+  be used by the receiving server to verify request authenticity when configured for notification
+  configurations with a destination type of `generic`. A token set for notification configurations
+  with a destination type of `slack` is not allowed and will result in an error. Defaults to `null`.
+* `triggers` - (Optional) The array of triggers for which this notification configuration will
```
vancluever: `[]` as a default here might not be 100% descriptive. Would a better way to describe the default be `No triggers are configured by default`?

tfe/resource_tfe_notification_configuration_test.go
svanharmelen: Agree. It should probably have just one dedicated import test.

tfe/resource_tfe_notification_configuration_test.go
svanharmelen: @vancluever this is actually a super common use pattern within providers and is also used everywhere in this provider, so I strongly advice to keep it like this.

website/docs/r/notification_configuration.html.markdown
svanharmelen: In this provider those snipped are usually copied from the API doc's as they should already effectively describe what this is for.

website/docs/r/notification_configuration.html.markdown
svanharmelen: I would just remove the `Defaults to '[]'` part as it's not really a default. It's an optional field which doesn't get assigned any value by default.

tfe/resource_tfe_notification_configuration_test.go
svanharmelen: I too think we should exclude these tests from the provider. Next to it being very uncommon to test API responses like this is the provider, it also seems (to me at least) that this is not really testing the provider, but instead testing the API SDK.

But that one (the [go-te](https://github.com/hashicorp/go-tfe/) package) actually does test all this quite extensively already and if you see places where it doesn't, it probably adds more value to update the related tests in there?

tfe/resource_tfe_notification_configuration.go
```diff
@@ -0,0 +1,205 @@
...
+				Optional:  true,
+				Sensitive: true,
+			},
+
+			"triggers": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice(
```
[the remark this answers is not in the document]
lafentres: @vancluever, yep, this worked :) In addition to passing the tests I wrote, it also stopped me from putting in various invalid strings when I was creating notifications with Terraform locally.

tfe/resource_tfe_notification_configuration_test.go
lafentres: I was going to make this a dedicated import test but when reading through [the docs on imports](https://www.terraform.io/docs/extend/resources/import.html), it recommended against that. Did I misunderstand that or are the docs out of date?

website/docs/r/notification_configuration.html.markdown
lafentres: I removed the `Defaults to '[]'` line and added a note about no triggers being configured if you omit this attribute.

tfe/resource_tfe_notification_configuration_test.go
lafentres: I went ahead and removed these tests, since it doesn't really match up with what is being done in other places in the provider.

tfe/resource_tfe_notification_configuration_test.go
lafentres: Removed, I think your point about other tests not including this is a good one.

tfe/resource_tfe_notification_configuration_test.go
lafentres: Done! :)

tfe/resource_tfe_notification_configuration_test.go
vancluever: @svanharmelen this does look like it's the case, sounds good.

website/docs/r/notification_configuration.html.markdown
vancluever: Yep, that's why I advised to keep.

@lafentres feel free to resolve this at your leisure.

tfe/resource_tfe_notification_configuration.go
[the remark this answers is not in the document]
vancluever: Weird. I'm asking @radeksimko to weigh in here as I'm not too sure how this is working given the restrictions in place. Maybe the acceptance testing suite is not using `InternalValidate` after all.

tfe/resource_tfe_notification_configuration.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// Add triggers set to the options struct
+	for _, trigger := range d.Get("triggers").(*schema.Set).List() {
+		options.Triggers = append(options.Triggers, trigger.(string))
+	}
+
+	log.Printf("[DEBUG] Update notification configuration: %s", d.Id())
+	_, err := tfeClient.NotificationConfigurations.Update(ctx, d.Id(), options)
+	if err != nil {
+		return fmt.Errorf("Error updating notification configuration %s: %v", d.Id(), err)
```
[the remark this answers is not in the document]
vancluever: I'd argue that deviating from the error format casing could create problems in the future should we ever decide to change how errors are formatted or prefix them with something else. Additionally, I think this is applied inconsistently throughout the codebase of other providers and is probably more a matter of the contributor's personal preference and whether or not they are using golint.

Regardless, this is why I noted this as being very minor, I would not expect to see it changed here as it's what is consistent with the rest of the providers. @lafentres feel free to resolve at your leisure as well.

tfe/resource_tfe_notification_configuration_test.go
vancluever: I kind of wonder if maybe the example needs some clarification? I've never seen the import test combined like that. The `existing TestStep ` part may also be pertaining to what you would need to do to set up an import case.

lafentres: I went ahead and pulled it out into its own test. That's consistent with the rest of the provider. 

I'm making a note to investigate this more later though, because under [the "Additional Information" section in the docs](https://www.terraform.io/docs/extend/resources/import.html#additional-information), it seems to explicitly says you shouldn't make a separate import test.

vancluever: @lafentres ah, yeah that lays it out much more clearly. This could have changed since I was writing providers on a regular basis, so I would say definitely go by the document in the future! The enablement team would probably be the best people to talk here about this, and is the team that probably put this in.

tfe/resource_tfe_notification_configuration.go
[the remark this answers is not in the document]
radeksimko: This works just fine because you are validating each element, not the whole set. In other words the `ValidateFunc` is associated with `TypeString`, not `TypeSet`.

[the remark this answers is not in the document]
vancluever: @radeksimko what I'm wondering about is how it's passing `InternalValidate` with the current code in place: https://github.com/hashicorp/terraform/blob/v0.12.6/helper/schema/schema.go#L831-L836

Issue #82, opened 2019-07-02
Support insecure_skip_tls_verify option


dan-wittenberg:
Other providers support this option, i.e.
https://github.com/terraform-providers/terraform-provider-triton#setup

This helps during initial automated setups of pTFE where you might not yet have a real signed cert to use.

chrisarcand:
This was added long ago, so closing; thanks!

PR #95, opened 2019-09-26
Add an option to skip certificate verifications

svanharmelen:
This should solve #82, but as I don't have a pTFE install to test against it would be great if this can be tested by someone from support before merging this one.


etopeter:
Can we get PR reviewed?

petems:
@beekus Could we get some 👀 on this one?

koikonom:
Since this was a pretty old PR I rebased it on top of master which is why I had to force push.

lafentres (approved):

PR #97, opened 2019-10-09
Fix terraform_version description.

rrijkse:
The `trigger_prefixes` line was embedded into the `terraform_version` line. Just cleaning that up.


svanharmelen:
Thanks @rrijkse! LGTM!

PR #100, opened 2019-11-01
Improve test documentation and rename policy set environment variables

lafentres:
I frequently struggle to get these tests running so I thought it was time to document the process. I also changed `TFE_VCS_IDENTIFIER`, `TFE_POLICY_SET_VCS_BRANCH`, and `TFE_POLICY_SET_VCS_PATH` to `GITHUB_POLICY_SET_IDENTIFIER`, `GITHUB_POLICY_SET_BRANCH`, and `GITHUB_POLICY_SET_PATH` to make it a little more obvious what they are used for.


svanharmelen:
@lafentres I personally think these the env vars should start with `TFE` as they are used to configure VCS agnostic fields in the TFE provider and are not well known/used Github env vars.

lafentres:
> @lafentres I personally think these the env vars should start with `TFE` as they are used to configure VCS agnostic fields in the TFE provider and are not well known/used Github env vars.

@svanharmelen ah, I hadn't thought of it that way. I was thinking it might be easier to understand if the prefix indicated where the information in the variable is coming from (GitHub, TFE/TFC, etc). I was also matching some updates I recently did to go-tfe docs. But if that doesn't make sense and/or that's not how these are generally used, I can revisit this.

thrashr888:
I think I generally agree that we should prefer our own prefix namespacing. `GITHUB_*` seems like it should be something that comes from GitHub. For example, `GITHUB_TOKEN` comes from GitHub and applies to auth against its API. `TFE_*` was specific for our app and applied to our app's settings. Also, with the old names you could theoretically change the VCS provider without changing the `TFE_*` values.

That said, within this context it doesn't seem to matter much and it's just stylistic, so I'm fine to go either way.

lafentres:
I'm going to leave these changes as I have them for now to keep things consistent with go-tfe and because at the moment we are only testing GitHub. 

I definitely understand what you are saying about the TFE prefix and the policy set fields. This all makes me feel like these values sit in kind of a weird place in that, for the tests anyway, the value doesn't exist in TFC/TFE yet. So it is info that kind of comes from GitHub but that is specific to TFC/TFE. I think when/if we update the tests to work with other VCS providers, this is something we should address. I'm going to keep track of this in the board I'm putting together to keep track of work for this repo.

PR #114, opened 2020-01-13
Add organization membership resource

beekus:
## TODO:
- [x] Resource documentation
- [x] Merge https://github.com/hashicorp/go-tfe/pull/98 and update `go.mod`

## Description
This PR adds the `organization-membership` resource. This resource will always destroy/create on updating since a user's email address can only be changed by that user alone. This creates some interesting scenarios when updating the email address in the configuration:

* **The email address is only changed in the TF configuration.** This will cause the invitation to be rescinded, or the user to be kicked out of the org and invited again with the new email address if the user has already accepted the invitation. It would be up to the person applying the configuration change to catch this and fix it.

* **The email address is changed by the user in the UI, but not changed in the TF configuration.** The next time the configuration is applied, the user will be kicked out of the organization and a new invitation will be created for the original email address. Similarly, it is up to the person applying the configuration change to catch and fix it.

* **The email address is changed in the configuration and the UI.** Terraform recognizes this scenario and realizes that the plan has no changes to apply.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/terraform-providers/terraform-provider-tfe
 
 require (
 	github.com/hashicorp/go-hclog v0.0.0-20190109152822-4783caec6f2e // indirect
-	github.com/hashicorp/go-tfe v0.3.30
+	github.com/hashicorp/go-tfe v0.3.31
 	github.com/hashicorp/go-version v1.2.0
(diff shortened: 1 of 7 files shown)


beekus:
I don't think we should merge this as I think it makes the experience actually worse and more confusing for users of the provider.

PR #138, opened 2020-02-21
Reset trigger prefixes for workspace resources when removed from the config

nicsnet:
## Description

Fixes a bug for `tfe_workspace` resource updates. Previously when creating a `tfe_workspace` resource with a `trigger_prefixes` configuration, this option remained even when the `trigger_prefixes` config option was removed during subsequent updates to this resource. `terraform plan` would continue outputting this change, but it would not take effect.

## Testing plan

``` 
 # tfe_workspace.pleng_staging_us_west_2 will be updated in-place
  ~ resource "tfe_workspace" "pleng_staging_us_west_2" {
        auto_apply            = false
        external_id           = "ws-6YmtKxQffbqJpVE6"
        file_triggers_enabled = true
        id                    = "chavo4/pleng-staging-us-west-2"
        name                  = "pleng-staging-us-west-2"
        operations            = true
        organization          = "chavo4"
        queue_all_runs        = true
        terraform_version     = "0.12.12"
      ~ trigger_prefixes      = [
          - "567694625192/us-west-2",
        ]
        working_directory     = "567694625192/us-west-3"

        vcs_repo {
            identifier         = "nicsnet/terraform-provider-nics"
            ingress_submodules = false
            oauth_token_id     = "#######"
        }
    }
````

```
$ make testacc

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 0.15.0 (Unreleased)
+## 0.14.1 (Unreleased)
+BUG FIXES:
+
+* t/tfe_workspace: Issues with updating `working_directory` and `trigger_prefixes` when removed from the configuration. Special note: if you have workspaces which are configured through the TFE provider, but have set the working directory or trigger prefixes manually, through the UI, you'll need to update your configuration.
+
 ## 0.14.0 (February 20, 2020)
 
 FEATURES:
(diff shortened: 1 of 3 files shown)


acespacecase:
Everything looks good on my end, I just have one question. Is it within scope to allow `working_directory` to be reset? It can't be cleared once it's been set via the provider, similar to the `trigger_prefixes`.

nicsnet:
The `working_directory` fix is part of this https://github.com/terraform-providers/terraform-provider-tfe/pull/137

PR #144, opened 2020-03-10
Fix/var error

scottwinkler:
## Description

The Read() operation on `tfe_variable` currently returns a hard error if the workspace that the variable is part of does not exist. This is unnecessary and inconvenient should a workspace be deleted before resources managing the variables are deleted. It is reasonable to assume that if a workspace no longer exists, then neither do any  of the variables that are part of it.

## Testing plan

1.  Create a workspace. Deploy a `tfe_variable` resource to that workspace. Delete the workspace then run a `terraform plan`. The plan will fail because the read on the workspace fails, even though it shouldn't


etopeter:
@scottwinkler Can you please resolve conflicts? This is a must have fix someone that wants to manage variables outside of workspace. Having manually delete state resources every time external workspace gets deleted is super annoying. I hope this goes out next release.

scottwinkler:
rebased to master

chrisarcand:
Resolved in https://github.com/hashicorp/terraform-provider-tfe/commit/e9b7a0d639209d022b814ba70e2f096835f6da63

PR #155, opened 2020-04-03
team: add organization_access

bendrucker:
## Description

Adds support for specifying `organization_access` and `visibility` to `tfe_team`:

https://www.terraform.io/docs/cloud/api/teams.html#request-body

## Testing plan

1.  Add an `organization_block` or `visibility` to a `tfe_team`
1.  Build the provider and apply

```
TESTARGS="-run ^TestAccTFETeam_" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run ^TestAccTFETeam_ -timeout 15m
?   	github.com/terraform-providers/terraform-provider-tfe	[no test files]
--- PASS: TestAccTFETeam_basic (4.39s)
=== RUN   TestAccTFETeam_full
--- PASS: TestAccTFETeam_full (3.69s)
=== RUN   TestAccTFETeam_import
--- PASS: TestAccTFETeam_import (3.79s)
PASS
ok  	github.com/terraform-providers/terraform-provider-tfe/tfe	12.189s
?   	github.com/terraform-providers/terraform-provider-tfe/version	[no test files]
```

## Module Updates

This depends on https://github.com/hashicorp/go-tfe/pull/113 (v0.7). I updated the version in `go.mod`, ran `go mod tidy`, and ran `go mod vendor` to update the vendor directory.


bendrucker:
Any chance I can get a review for this? Thanks!

bendrucker:
I went ahead and added `visibility` here as well, since otherwise I would have run into conflicts. Tests are passing, docs are updated, let me know if there's anything else I can do here!

bendrucker:
@chrisarcand Anything I can do to help get this released? It's not urgent, just hoping to close out it. Happy to split up the go-tfe upgrade and the actual changes required here if that's easier.

tfe/resource_tfe_team.go
```diff
@@ -24,26 +27,60 @@ func resourceTFETeam() *schema.Resource {
...
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+					},
+				},
+			},
+			"visibility": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
```
chrisarcand, line 59: Hmm I'm not sure we want this to be marked as a computed value. `Computed` is used to represent values that are not user configurable or aren't known at time of an `apply` or `plan` operation, which this shouldn't be.

tfe/resource_tfe_team.go
```diff
@@ -24,26 +27,60 @@ func resourceTFETeam() *schema.Resource {
...
-	// Get the name and organization.
+	// Get team attributes.
 	name := d.Get("name").(string)
 	organization := d.Get("organization").(string)
 
 	// Create a new options struct.
 	options := tfe.TeamCreateOptions{
-		Name: tfe.String(name),
+		Name:               tfe.String(name),
+		OrganizationAccess: getTeamOrganizationAccess(d),
```
chrisarcand, line 79: Given that this is also an optional block, I think this might be clearer to follow the same `if v, ok := d.GetOk("organization_access"); ok {` pattern; that's what I did in [my version of this](https://github.com/terraform-providers/terraform-provider-tfe/pull/52/files#diff-d0f8a2703c6d17783be2da858fb5047dR77-R86), too.

tfe/resource_tfe_team.go
bendrucker: I set this based on the following comment from the plugin SDK:

```
// If Computed is true, then the result of this value is computed
// (unless specified by config) on creation.
```

I take that to mean that any with a server set default could be "computed," though this probably moot when it's an optional field where the zero value of the type is equal to the server value. If the server were capable of defaulting to `true`, I think this might need to be set though? I can remove it here.

bendrucker:
Thanks! Glad my timing was good. 😉 

The caveats makes sense, appreciate the explanation. I agree that differing authorization will inevitably cause problems and that it's really a general limitation.

I'll get an update test added/passing and then address the other comments. Should be later this week!

tfe/resource_tfe_team.go
bendrucker: Hmm, was mis-remembering what values `visibility` could have. I think this needs to be computed since the API will respond with `visibility = "secret"` regardless of whether the user has set a value and we'll persist that value to state. I would expect Terraform to present a diff from `"secret" -> null` on subsequent runs if we don't account for this. I think computed takes care of this, but correct me if I'm wrong.

bendrucker:
Back a bit sooner than expected 🙂! Resolved the docs/style suggestions, thanks for those. Remaining are:

1. Whether `visibility` needs to be computed to prevent its server-assigned default from being raised as a diff (https://github.com/terraform-providers/terraform-provider-tfe/pull/155#discussion_r423897156)
2. I added a test that updates both visibility and organization access, but was unable to reproduce the issue you described where updates aren't propagated. Since `testAccCheckTFETeamExists` reads directly from API and then we're making assertions on the response, I feel pretty confident the update is happening. But if you tell me what you were seeing I'm happy to look into it.

bendrucker:
And for good measure, all team acceptance tests are still passing:

```
$ TESTARGS="-run TestAccTFETeam_" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFETeam_ -timeout 15m
?   	github.com/terraform-providers/terraform-provider-tfe	[no test files]
--- PASS: TestAccTFETeam_basic (3.94s)
=== RUN   TestAccTFETeam_full
--- PASS: TestAccTFETeam_full (3.60s)
=== RUN   TestAccTFETeam_full_update
--- PASS: TestAccTFETeam_full_update (5.90s)
=== RUN   TestAccTFETeam_import
--- PASS: TestAccTFETeam_import (3.59s)
PASS
ok  	github.com/terraform-providers/terraform-provider-tfe/tfe	17.288s
?   	github.com/terraform-providers/terraform-provider-tfe/version	[no test files]
```

tfe/resource_tfe_team.go
```diff
@@ -24,20 +27,49 @@ func resourceTFETeam() *schema.Resource {
...
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+						"manage_workspaces": {
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+						"manage_vcs_settings": {
+							Type:     schema.TypeBool,
+							Optional: true,
```
chrisarcand, line 51: Each of these should be `Default: false`

tfe/resource_tfe_team.go
chrisarcand: > since the API will respond with visibility = "secret" regardless of whether the user has set a value 

Nope, the API should respond with whatever value the user set - and valid options are `"secret"` or `"organization"`. The default is `"secret"`.

* `Computed` is very much for unknown values at plan-time that are _not_ user-configurable - things like dates and UUIDs.
* `Default` is for known values to be used when something isn't explicitly set within the configuration.

For our case here, these are all values that have known defaults that we can provide at plan-time; so they can all be defaulted appropriately. (So this visibility section should not be computed, and should have a default of `"secret"`).

When that's all finished, given a configuration like this:

```hcl
resource "tfe_team" "test-1" {
  name         = "my-team-name"
  organization = var.organization
  organization_access {
    manage_policies = true
    manage_workspaces = false
  }
}
```

We should see a plan output like this:

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # tfe_team.test-1 will be created
  + resource "tfe_team" "test-1" {
      + id           = (known after apply)
      + name         = "my-team-name"
      + organization = "my-cool-organization"
      + visibility   = "secret"

      + organization_access {
          + manage_policies     = true
          + manage_vcs_settings = false
          + manage_workspaces   = false
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

See the [Schema Behaviors documentation](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html) for more info on both of these behaviors.

bendrucker:
Thank you Chris! Added those defaults and removed `Computed`.

PR #157, opened 2020-04-06
[docs] update tfe_workspace import docs to use workspace id

ryanking:
## Description

It seems that workspace ID, rather than name, is now required for an import.

I was following the directions to do an import and discovered that using the workspace name does not work, but the workspace ID does.

## Testing plan

This is just a documentation change, so no testing required?


lafentres:
Good catch and thanks for taking care of this! 
It looks like you and the author of #156 ran into both ran into this and submitted PRs. I'm going to go ahead and close this one, as I've already merged #156.

PR #184, opened 2020-06-12
tfe_team_access: Add Custom workspace access support

chrisarcand:
## Description

- [x] **Requires a go-tfe release with https://github.com/hashicorp/go-tfe/pull/123**

These changes contain:

* Added support for Custom workspace access in the [Team Access API](https://www.terraform.io/docs/cloud/api/team-access.html).
* In-place updating of Team Accesses rather than always ForceNew behavior

The new API behavior is tough to model because of a number of limitations within the provider SDK, such as:

* Schema validations are limited to the single attribute they are defined on; you cannot validate something with the additional context of another attribute's value in the resource.
* The SDK cannot discern between something defined _only_ in state, or _only_ in configuration. Some assumptions can be made (and are made in these changes) via GetChange(), but it's hacky at best.

After a few different iterations attempting to conditionally allow a user to edit custom permissions only when access is set to custom, the core of this design is to place custom access within a new `permissions` block which is validated at the schema level to conflict with `access`; `custom` access itself and the read-only aspect of them when `access` is not custom is then handled by plan-time deductions with `CustomizeDiff`.

A huge thank you to Martin Atkins (@apparentlymart) for all the time spent assisting me on getting this right ❤️ 

## Testing plan

* Use `tfe_team_access` resource to manage [Team Access](https://www.terraform.io/docs/cloud/api/team-access.html). Honestly, just try and break it. Maybe output the data source to make sure that works, as well.

```
» TESTARGS="-run TestAccTFETeamAccess" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFETeamAccess -timeout 15m
=== RUN   TestAccTFETeamAccessDataSource_basic
--- PASS: TestAccTFETeamAccessDataSource_basic (10.83s)
=== RUN   TestAccTFETeamAccess_write
--- PASS: TestAccTFETeamAccess_write (8.13s)
=== RUN   TestAccTFETeamAccess_custom
--- PASS: TestAccTFETeamAccess_custom (9.81s)
=== RUN   TestAccTFETeamAccess_updateToCustom
--- PASS: TestAccTFETeamAccess_updateToCustom (12.87s)
=== RUN   TestAccTFETeamAccess_updateFromCustom
--- PASS: TestAccTFETeamAccess_updateFromCustom (13.10s)
=== RUN   TestAccTFETeamAccess_import
--- PASS: TestAccTFETeamAccess_import (9.03s)
PASS
ok      github.com/terraform-providers/terraform-provider-tfe/tfe       64.374s
```
## Behavior walkthrough

Creating a new team access with standard Write access:

image

Creating a new team access with standard Write access (not available at plan time):

image

Creating a new team access with Custom access:

image

Creating a new team access with Custom access (not available at plan time):

image

Updating access from a standard level (write) to custom access:

image

Updating access from custom to standard access:

image

Updating access from custom to standard access (not available at plan time):

image

### DIFF
--- a/tfe/resource_tfe_team_access.go
+++ b/tfe/resource_tfe_team_access.go
@@ -60,6 +60,13 @@ func resourceTFETeamAccess() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							Computed: true,
+							AtLeastOneOf: []string{
+								"permissions.0.runs",
+								"permissions.0.variables",
+								"permissions.0.state_versions",
+								"permissions.0.sentinel_mocks",
+								"permissions.0.workspace_locking",
+							},
 							ValidateFunc: validation.StringInSlice(
 								[]string{
 									string(tfe.RunsPermissionRead),
--- a/tfe/resource_tfe_team_access.go
+++ b/tfe/resource_tfe_team_access.go
@@ -58,15 +58,7 @@ func resourceTFETeamAccess() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"runs": {
 							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
-							AtLeastOneOf: []string{
-								"permissions.0.runs",
-								"permissions.0.variables",
-								"permissions.0.state_versions",
-								"permissions.0.sentinel_mocks",
-								"permissions.0.workspace_locking",
-							},
+							Required: true,
 							ValidateFunc: validation.StringInSlice(
 								[]string{
 									string(tfe.RunsPermissionRead),
@@ -79,8 +71,7 @@ func resourceTFETeamAccess() *schema.Resource {
 
 						"variables": {
 							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
+							Required: true,
 							ValidateFunc: validation.StringInSlice(
 								[]string{
 									string(tfe.VariablesPermissionNone),
@@ -93,8 +84,7 @@ func resourceTFETeamAccess() *schema.Resource {
 
 						"state_versions": {
 							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
+							Required: true,
 							ValidateFunc: validation.StringInSlice(
 								[]string{
 									string(tfe.StateVersionsPermissionNone),
@@ -108,8 +98,7 @@ func resourceTFETeamAccess() *schema.Resource {
 
 						"sentinel_mocks": {
 							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
+							Required: true,
 							ValidateFunc: validation.StringInSlice(
 								[]string{
 									string(tfe.SentinelMocksPermissionNone),
@@ -121,8 +110,7 @@ func resourceTFETeamAccess() *schema.Resource {
 
 						"workspace_locking": {
 							Type:     schema.TypeBool,
-							Optional: true,
-							Computed: true,
+							Required: true,
 						},
 					},
 				},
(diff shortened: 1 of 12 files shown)


tfe/resource_tfe_team_access.go
```diff
@@ -166,3 +345,33 @@ func resourceTFETeamAccessImporter(d *schema.ResourceData, meta interface{}) ([]
...
 	return []*schema.ResourceData{d}, nil
 }
+
+func setCustomOrComputedPermissions(d *schema.ResourceDiff, meta interface{}) error {
+	if _, ok := d.GetOk("access"); ok {
+		if d.HasChange("access") {
+			_, n := d.GetChange("access")
+			new := n.(string)
+
+			if new != "custom" {
```
chrisarcand, line 355: ⚠️ Martin brought up that this might be problematic for when some sort of interpolation is done that cannot be known at plan-time. I don't really think there's any way around this limitation, and at this point think it's reasonable to expect that changes to resources that affect authorization in the application probably should _not_ be done only at apply-time anyway ;)

lafentres:
you might already know this but this PR will need to include updates to the docs for the team access resource and data source (in the website folder)

lafentres:
possibly a bug:
when i update my config from `access = "plan"` to `permissions {}` nothing happens. i get an empty/no changes plan. this seems like something that should throw an error. could you maybe add a `MinItems: 1` to the schema for the permissions attribute?

```
resource "tfe_team_access" "test" {
  access         = "plan"
  team_id        = tfe_team.team-1.id
  workspace_id   = tfe_workspace.workspace-1.id
}
```

```
resource "tfe_team_access" "test" {
  permissions {}
  team_id        = tfe_team.team-1.id
  workspace_id   = tfe_workspace.workspace-1.id
}
```

tfe/resource_tfe_team_access.go
apparentlymart: It is in theory possible to address it by calling [`d.NewValueKnown("custom")`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceDiff.NewValueKnown) and doing something special if it returns `false`, but I'm not really sure _what_ you'd do if it were false... that would indicate that at apply time the value could be either `custom` or not `custom`, which would mean that `permissions` could either be "computed" (unknown) or not.

One way we could avoid potential into-corner-painting :paintbrush: here would be to make an explicit block of this argument being unknown, so users who do this can see explicitly that it isn't allowed. That is, to _begin_ this function with an early exit for that case:

```go
if !d.NewValueKnown("custom") {
    return fmt.Errorf("'access' cannot be derived from a value that is unknown during planning")
}
```

This error message is not particularly actionable, but I think that can be justified by the fact that this situation ought to be very rare and so this test is there only as a safety to prevent something odd/unclear happening in that unlikely event, and prevent that odd behavior from potentially becoming a compatibility constraint.

tfe/resource_tfe_team_access.go
```diff
@@ -44,6 +49,78 @@ func resourceTFETeamAccess() *schema.Resource {
 				),
 			},
 
+			"permissions": {
```
lafentres, line 52: this is actually kinda how i wish the api was, with the individual permissions nested under a permissions block 😄

tfe/resource_tfe_team_access.go
lafentres: can you explain what this means a little more? do you have an example of interpolation that can't be known at plan time? i'm sorry, i'm just having a hard time wrapping my head around this and what sort of things we should be looking out for.

tfe/resource_tfe_team_access.go
chrisarcand: A discussion was had about that, actually; it was decided in the end that it's too confusing with several API responses already containing `permissions` hashes that relate to what _you are able to do with the resource_, and not 'what you are granting through this resource as here.

tfe/resource_tfe_team_access.go
chrisarcand: @lafentres With data resources that have computed values, the read is deferred until the apply phase - so if you were to do an interpolation involving that to determine the access level here, that would be an example of something where the logic here (which needs to be `"custom"` at plan time) would break.

I like Martin's suggestion of avoiding it altogether by restricting that behavior with an error, good idea!

chrisarcand:
> you might already know this but this PR will need to include updates to the docs for the team access resource and data source (in the website folder)

Whoooops 😄  Definitely.

> could you maybe add a MinItems: 1 to the schema for the permissions attribute?

Sadly that wouldn't do anything; `MinItems` refers to the number of blocks (because each block is really stored in the same list), not the arguments within the block.

I tried using the shiny new `AtLeastOneOf` validation from the SDK for this reason but couldn't get it to work; gives an odd error. I'll toy with it a bit more but did know this and gave up, not caring. `permissions {}` looks odd but harmless, as it just means each value is computed with custom access (so each value from the API is the default, their lowest permission level, respectively.

chrisarcand:
@lafentres @apparentlymart  This is finally, mercifully, done. Ready for review!

Since the first review:

* Any value being unknown at plan time is now accounted for. You _may_ interpolate `access`, but trying to interpolate a custom `permission` is forbidden.
* All `permission` arguments are required if the block is specified.
* Documentation ✅
* Released go-tfe and bumped!

chrisarcand:
(Rebased master)

website/docs/d/team_access.html.markdown
```diff
@@ -31,4 +31,9 @@ The following arguments are supported:
 In addition to all arguments above, the following attributes are exported:
 
 * `id` The team access ID.
-* `access` - The type of access granted.
+* `access` - The type of access granted to the team on the workspace.
+* `permissions.0.runs` - The permission granted to runs. Valid values are `read`, `plan`, or `apply`
+* `permissions.0.variables` - The permissions granted to variables. Valid values are `none`, `read`, or `write`
+* `permissions.0.state_versions` - The permissions granted to state versions. Valid values are `none`, `read-outputs`, `read`, or `write`
+* `permissions.0.sentinel_mocks` - The permissions granted to Sentinel mocks. Valid values are `none` or `read`
+* `permissions.0.workspace_locking` - Whether permission is granted to manually lock the workspace or not.
```
lafentres, line 39: this should bring these docs closer to the other docs we have that use similar blocks.
```suggestion
* `access` - The type of access granted to the team on the workspace.
* `permissions` - The permissions granted to the team on the workspaces for each whatever.

The `permissions` block contains:

* `runs` - The permission granted to runs. Valid values are `read`, `plan`, or `apply`
* `variables` - The permissions granted to variables. Valid values are `none`, `read`, or `write`
* `state_versions` - The permissions granted to state versions. Valid values are `none`, `read-outputs`, `read`, or `write`
* `sentinel_mocks` - The permissions granted to Sentinel mocks. Valid values are `none` or `read`
* `workspace_locking` - Whether permission is granted to manually lock the workspace or not.
```

website/docs/r/team_access.html.markdown
```diff
@@ -36,10 +36,20 @@ resource "tfe_team_access" "test" {
...
+
+  The arguments for this block are:
+
+  - `runs` - (Required) The permission to grant the team on the workspace's runs. Valid values are `read`, `plan`, or `apply`.
+  - `variables` - (Required) The permission to grant the team on the workspace's variables. Valid values are `none`, `read`, or `write`.
+  - `state_versions` - (Required) The permission to grant the team on the workspace's state versions. Valid values are `none`, `read`, `read-outputs`, or `write`.
+  - `sentinel_mocks` - (Required) The permission to grant the team on the workspace's generated Sentinel mocks, Valid values are `none` or `read`.
+  - `workspace_locking` - (Required) Boolean determining whether or not to grant the team permission to manually lock/unlock the workspace.
+
+At least one of `access` or `permissions` must be specified, but not both. Whichever is omitted will automatically reflect the state of the other.
```
lafentres, line 52: same thing as above, make a little more consistent with other docs:
```suggestion
* `access` - (Optional) Type of fixed access to grant. Valid values are `admin`, `read`, `plan`, or `write`. To use `custom` permissions, use a `permissions` block instead. This value _must not_ be provided if `permissions` is provided.
* `permissions` - (Optional) Permissions to grant using [custom workspace permissions](https://www.terraform.io/docs/cloud/users-teams-organizations/permissions.html#custom-workspace-permissions). This value _must not_ be provided if `access` is provided.

The `permissions` block supports:

* `runs` - (Required) The permission to grant the team on the workspace's runs. Valid values are `read`, `plan`, or `apply`.
* `variables` - (Required) The permission to grant the team on the workspace's variables. Valid values are `none`, `read`, or `write`.
* `state_versions` - (Required) The permission to grant the team on the workspace's state versions. Valid values are `none`, `read`, `read-outputs`, or `write`.
* `sentinel_mocks` - (Required) The permission to grant the team on the workspace's generated Sentinel mocks, Valid values are `none` or `read`.
* `workspace_locking` - (Required) Boolean determining whether or not to grant the team permission to manually lock/unlock the workspace.

-> **Note:** At least one of `access` or `permissions` _must_ be provided, but not both. Whichever is omitted will automatically reflect the state of the other.
```

lafentres: could also take the optional bit off that i added here and leave it neither require or optional since it is special 🤷 
or leave the conflicts with thing in there.

chrisarcand: 😬 I actually used a nested list here already: https://www.terraform.io/docs/providers/tfe/r/team.html#argument-reference

But looking at the site, I guess the markdown renderer we use doesn't actually respect nested lists, so it looks wrong and I should change that too 🤦🏻‍♂️

image

PR #190, opened 2020-07-13
support organisation member import

kevholditch:
## Description

adding support for import for tfe_organization_membership.  If you have
TFE running and have existing teams then you need to be able to import
those memberships so you can use TF to manage them.

## Testing plan

1.  You cannot currently import organisation membership as it is not supported on the resource

```
=== RUN   TestAccTFEOrganizationMembership_import
--- PASS: TestAccTFEOrganizationMembership_import (6.98s)
```

```
$ make testacc

...
```


preston-hf:
That's all there was to it? Very nice, I need this :)

lafentres:
Hi there! It looks like we were both working on adding import at the same time. I had added it to support the [work I was doing on notification configurations](https://github.com/terraform-providers/terraform-provider-tfe/pull/191). I just merged and released all of that this morning so support for organization membership import is officially available as of v0.20.0.

preston-hf:
Awesome thanks!

Issue #200, opened 2020-08-05
Error: Provider produced inconsistent result after apply - Registry Module


scottwinkler:
I am using the`tfe_registry_module` to create modules in each organization. Here is my code:

```
resource "tfe_oauth_client" "oauth-client" {
  organization     = var.organization
  api_url          = "${var.github.address}/api/v3"
  http_url         = "${var.github.address}/"
  oauth_token      = var.github.token
  service_provider = "github_enterprise"
}

resource "tfe_registry_module" "registry-module" {
 for_each = toset(var.vcs_repo_identifiers)
  vcs_repo {
    display_identifier = each.key
    identifier         = each.key
    oauth_token_id     = tfe_oauth_client.oauth-client.oauth_token_id
  }
}
```

var.vcs_repo_identifiers is a list of modules to publish, currently it is lengt two.

When I apply this the plan succeeds and actually the modules are created except i get a runtime error:

```
This is a bug in the provider, which should be reported in the provider's own issue tracker.
2020-08-04T18:24:10.361-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: 2020/08/04 18:24:10 [DEBUG] TFE API Response Details:
2020-08-04T18:24:10.361-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: ---[ RESPONSE ]--------------------------------------
2020-08-04T18:24:10.361-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: HTTP/2.0 404 Not Found
2020-08-04T18:24:10.361-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: Connection: close
2020-08-04T18:24:10.361-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: Cache-Control: no-cache
2020-08-04T18:24:10.362-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: Content-Type: application/vnd.api+json; charset=utf-8
2020-08-04T18:24:10.362-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: Date: Wed, 05 Aug 2020 01:24:10 GMT
2020-08-04T18:24:10.362-0700 [DEBUG] plugin.terraform-provider-tfe_v0.20.0_x4: Server: nginx/1.17.10
[...]

PR #215, opened 2020-09-10
Check if module has been imported

koikonom:
## Description

When creating a new module in the registry, we need to check if the
registry has finished importing it before returning. If we don't and the
Read() function is called before the process is over then we end up with
an invalid terraform state.

```
❯ TESTARGS='-run TestAccTFERegistryModule_vcs' make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFERegistryModule_vcs -timeout 15m
=== RUN   TestAccTFERegistryModule_vcs
--- PASS: TestAccTFERegistryModule_vcs (34.29s)
PASS
ok      github.com/terraform-providers/terraform-provider-tfe/tfe       35.253s
?       github.com/terraform-providers/terraform-provider-tfe/version   [no test files]


...
```

### DIFF
--- a/tfe/resource_tfe_registry_module.go
+++ b/tfe/resource_tfe_registry_module.go
@@ -1,9 +1,11 @@
 package tfe
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	tfe "github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -83,6 +85,18 @@ func resourceTFERegistryModuleCreate(d *schema.ResourceData, meta interface{}) e
 	if err != nil {
 		return fmt.Errorf(
 			"Error creating registry module from repository %s: %v", *options.VCSRepo.Identifier, err)
+	}
+
+	doneCh := make(chan bool)
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(5)*time.Minute)
+	defer cancel()
+
+	go checkModuleExists(ctx, tfeClient, registryModule.Organization.Name, registryModule.Name, registryModule.Provider, doneCh)
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("timeout while waiting for module to finish importing in registry.")
+	case <-doneCh:
+		log.Printf("[DEBUG] Finished waiting for module to finish importing in registry.")
 	}
 
 	d.SetId(registryModule.ID)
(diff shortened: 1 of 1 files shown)


koikonom:
Closes #200.

tfe/resource_tfe_registry_module.go
```diff
@@ -85,6 +87,18 @@ func resourceTFERegistryModuleCreate(d *schema.ResourceData, meta interface{}) e
...
 	}
 
+	doneCh := make(chan bool)
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(5)*time.Minute)
+	defer cancel()
+
+	go checkModuleExists(ctx, tfeClient, registryModule.Organization.Name, registryModule.Name, registryModule.Provider, doneCh)
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("timeout while waiting for module to finish importing in registry.")
```
lafentres, line 97: nit: almost all of the errors are capitalized currently so we should capitalize this one for consistency
```suggestion
		return fmt.Errorf("Timeout while waiting for module to finish importing in registry")
```

tfe/resource_tfe_registry_module.go
```diff
@@ -171,3 +185,15 @@ func resourceTFERegistryModuleImporter(d *schema.ResourceData, meta interface{})
...
+	for {
+		_, err := tfeClient.RegistryModules.Read(ctx, organization, name, module_provider)
+		if err != nil {
+			time.Sleep(time.Duration(100) * time.Millisecond)
+			continue
+		}
+		doneCh <- true
+		break
+	}
+}
```
lafentres, line 199: since we're waiting for up to 5 minutes for the module to finish importing, it might be a good to make the sleep time a little longer here so we aren't sending so many requests to TFC/TFE. maybe something like 5-10 seconds? or possibly a very simple exponential backoff of 1 second, 2, 4, etc. what do you think?

lafentres (commented):
this looks great! tested it out locally and things work perfectly. i just have one question/concern about the duration we use for waiting before retrying.

tfe/resource_tfe_registry_module.go
koikonom: Good call! As a matter of fact I found out that the plugin SDK provides this functionality on its own (https://godoc.org/github.com/hashicorp/terraform-plugin-sdk/helper/resource#Retry). 

I changed the code to use that instead.

lafentres (approved):
looks good! tested it locally again and things look great 🎉

PR #220, opened 2020-09-21
Import branch

koikonom:
## Description

Importing a workspace with a non-default branch fails to import the branch.

```
PASS tfe.TestAccTFEWorkspace_importVCSBranch (4.10s)
...
```

### DIFF
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -163,7 +163,6 @@ func resourceTFEWorkspaceCreate(d *schema.ResourceData, meta interface{}) error 
 
 		options.VCSRepo = &tfe.VCSRepoOptions{
 			Identifier:        tfe.String(vcsRepo["identifier"].(string)),
-			Branch:            tfe.String(vcsRepo["branch"].(string)),
 			IngressSubmodules: tfe.Bool(vcsRepo["ingress_submodules"].(bool)),
 			OAuthTokenID:      tfe.String(vcsRepo["oauth_token_id"].(string)),
 		}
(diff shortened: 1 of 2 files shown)


tfe/resource_tfe_workspace.go
```diff
@@ -163,6 +163,7 @@ func resourceTFEWorkspaceCreate(d *schema.ResourceData, meta interface{}) error
 
 		options.VCSRepo = &tfe.VCSRepoOptions{
 			Identifier:        tfe.String(vcsRepo["identifier"].(string)),
+			Branch:            tfe.String(vcsRepo["branch"].(string)),
```
lafentres, line 166: does this override the conditional setting of branch on [lines 171-175](https://github.com/terraform-providers/terraform-provider-tfe/pull/220/files#diff-cc7250d176415243a532f4d62525f045R171-R175)? 
it seemed like previously, this was attempting to only set the branch if it was actually set in the config. 
i'm pretty sure it will work either way but i think we might want to pick one way or the other and remove the redundant code.

tfe/resource_tfe_workspace.go
```diff
@@ -231,6 +232,7 @@ func resourceTFEWorkspaceRead(d *schema.ResourceData, meta interface{}) error {
 	if workspace.VCSRepo != nil {
 		vcsConfig := map[string]interface{}{
 			"identifier":         workspace.VCSRepo.Identifier,
+			"branch":             workspace.VCSRepo.Branch,
```
lafentres, line 235: sort of the same question here as the one above. i think the goal is to only set the branch if we actually get one back  from the API request and it isn't `""`. there's clearly something wrong with the way we were attempting to do this before but i'm not sure what

tfe/resource_tfe_workspace.go
koikonom: You are right on this one, I missed it on the initial review because the diff was too short :) Good catch!

tfe/resource_tfe_workspace.go
koikonom: In this case we should remove the if block below. Read()'s result should not be affected by the HCL config at all, instead it should be reporting what the resource actually looks like. 

I'll update this PR shortly!

lafentres (approved):
this looks great to me! imported and created some workspaces locally and everything worked as expected 👍🏼

Issue #183, opened 2020-06-08
TF crash when running plan against TFE


mgarfias-avaya:
TF crashes on me (inside TF Cloud) when running the attached plan.  Also have attached the complete TFC log file.

[tfe-admin.zip](https://github.com/terraform-providers/terraform-provider-tfe/files/4748440/tfe-admin.zip)

[run-6H9GCXrcGjThZFua-apply.log](https://github.com/terraform-providers/terraform-provider-tfe/files/4748448/run-6H9GCXrcGjThZFua-apply.log)

I'm not sure if this is a provider issue or TFC.  Please advise.

```2020/06/08 18:39:43 [TRACE] module.workspace_ixcc-pips: eval: *terraform.EvalWriteState
2020/06/08 18:39:43 [TRACE] EvalWriteState: removing state object for module.workspace_ixcc-pips.tfe_workspace.this
2020/06/08 18:39:43 [TRACE] module.workspace_ixcc-pips: eval: *terraform.EvalIf
2020/06/08 18:39:43 [TRACE] module.workspace_ixcc-pips: eval: *terraform.EvalIf
2020/06/08 18:39:43 [TRACE] module.workspace_ixcc-pips: eval: *terraform.EvalWriteDiff
2020/06/08 18:39:43 [TRACE] module.workspace_ixcc-pips: eval: *terraform.EvalApplyPost
2020/06/08 18:39:43 [ERROR] module.workspace_ixcc-pips: eval: *terraform.EvalApplyPost, err: rpc error: code = Canceled desc = context canceled
2020/06/08 18:39:43 [ERROR] module.workspace_ixcc-pips: eval: *terraform.EvalSequence, err: rpc error: code = Canceled desc = context canceled
2020/06/08 18:39:43 [TRACE] [walkApply] Exiting eval tree: module.workspace_ixcc-pips.tfe_workspace.this
2020/06/08 18:39:43 [TRACE] vertex "module.workspace_ixcc-pips.tfe_workspace.this": visit complete
2020/06/08 18:39:43 [ERROR] module.workspace_ixcc-aks-dev2: eval: *terraform.EvalDiff, err: rpc error: code = Canceled desc = context canceled
2020/06/08 18:39:43 [ERROR] module.workspace_ixcc-aks-dev2: eval: *terraform.EvalSequence, err: rpc error: code = Canceled desc = context canceled
[...]

PR #223, opened 2020-09-23
Fix type conversion bug

koikonom:
## Description

While iterating over a trigger prefix list sometimes items can be nil.
We are now checking if conversion is possible instead of "blindly"
converting them and potentially causing a crash.


koikonom:
Closes #183 .

lafentres (approved):
this looks good to me! tested it locally with this config and compared/after:
```
resource "tfe_workspace" "test-workspace" {
  name = "test-workspace"
  organization = tfe_organization.test-org.name
  trigger_prefixes = ["", "test"]
}
```

it seems like the previous code was assuming that the elements of `trigger_prefixes` were never just empty strings. it might make sense to add some sort of validation for each element eventually, that way people get a warning that they've provided an improper trigger prefix.

PR #226, opened 2020-09-30
Clarify credential sources

koikonom:
This PR adds a few more ways on how to pass credentials to the provider and points out a potential gotcha when using the provider within TF[CE].


website/docs/index.html.markdown
```diff
@@ -37,11 +37,20 @@ organizations, and organization tokens cannot manage certain resource types
...
 
-There are three ways to provide the required token:
-
-- On the CLI, omit the `token` argument and set a `credentials` block in your
+There are a few ways to provide the required token:
+
+- The provider can read the TFE_TOKEN environment variable and the token
+stored there to authenticate. Please note that if you are using TFC or TFE
+then you need to set the environment in your workspace, even if you're using the
+`remote` backend.
```
lafentres, line 45:
```suggestion
- Setting the `TFE_TOKEN` environment variable: The provider can read the 
`TFE_TOKEN` environment variable and the token stored there to authenticate. 
Please note that if you are using Terraform Cloud or Terraform Enterprise, you 
will need to set the `TFE_TOKEN` environment variable in your workspace, even 
if you're using the `remote` backend.
```

website/docs/index.html.markdown
```diff
@@ -37,11 +37,20 @@ organizations, and organization tokens cannot manage certain resource types
...
+
+- On the CLI you can use the credentials generated by [`terraform login`](https://www.terraform.io/docs/commands/login.html)
+by setting the value of the environment variable `TERRAFORM_CONFIG` to the location
+of the file created (for example `~/.terraform.d/credentials.tfrc.json` on Linux systems).
```
lafentres, line 49:
```suggestion
- Using `terraform login` to generate credentials: On the CLI you can use the credentials 
generated by [`terraform login`](https://www.terraform.io/docs/commands/login.html)
by setting the value of the environment variable `TERRAFORM_CONFIG` to the location
of the file created (for example `~/.terraform.d/credentials.tfrc.json` on Linux systems).
```

website/docs/index.html.markdown
```diff
@@ -37,11 +37,20 @@ organizations, and organization tokens cannot manage certain resource types
...
+ 
+- Also on the CLI, omit the `token` argument and set a `credentials` block in your
   [CLI config file](/docs/commands/cli-config.html#credentials).
```
lafentres, line 52:
```suggestion
- Setting a `credentials` block in your CLI config file: On the CLI you can omit the 
`token` argument from your Terraform config and set a `credentials` block in your 
[CLI config file](/docs/commands/cli-config.html#credentials).
```

website/docs/index.html.markdown
```diff
@@ -37,11 +37,20 @@ organizations, and organization tokens cannot manage certain resource types
...
-- Set the `TFE_TOKEN` environment variable.
+
 - In a Terraform Enterprise workspace, set `token` in the provider
   configuration. Use an input variable for the token and mark the corresponding
   variable in the workspace as sensitive.
```
lafentres, line 56:
```suggestion
- Setting the `token` argument in the provider configuration: In a Terraform Cloud or 
Terraform Enterprise workspace you can set `token` in the provider configuration. 
Use an input variable for the token and mark the corresponding variable in the 
workspace as sensitive.
```

lafentres (approved):
👏🏼 🎉 
this looks great. i had some suggestions to try to make it clear that each thing was a separate option and not necessarily things you should try all together. let me know if you think it makes sense.

PR #230, opened 2020-10-20
Randomize org name

koikonom:
We are adding a suffix to the org name for each test. That way we can
have multiple people running the acceptance tests on the same instance.

Also a potential panic mid-test does not require an immediate cleanup
since the next run will create a different org.


lafentres (approved):
chef

PR #233, opened 2020-10-24
Invert org and workspace names

jeffcasavant:
## Description

I noticed this on a failed import:

```
Error: error retrieving workspace <org name> from organization <workspace name>: Error reading configuration of workspace <org name>/<workspace name>: resource not found
```

This log message is in the opposite order of the resource ID.  This PR inverts the refs in the log message to fix it.

I'm having a bit of trouble running the acceptance tests; not sure what's wrong, but I figured that this is small enough it might not matter.


chrisarcand (approved):
Thanks for submitting this!

PR #239, opened 2020-11-06
Add `cost_estimation_enabled` to `tfe_organization` resource

cooltoast:
## Description

The `tfe_organization` resource is missing an option to set cost estimation settings for its workspaces. 

The TFE API already supports setting this attribute: https://www.terraform.io/docs/cloud/api/organizations.html#request-body

## Testing plan

1. Create a TFE organization without setting `cost_estimation_enabled`.
  a. From my observations, TFE sets the default value to `true` if Cost Estimation is globally enabled through [TFE Admin Settings](https://www.terraform.io/docs/enterprise/admin/integration.html#cost-estimation-integration), and sets default to `false` if Cost Estimation is not globally enabled. This complicates checking the `cost_estimation_enabled` default value, so I did not include that in the tests.
2. Explicitly set `cost_estimation_enabled` to `true`, verify the change is made.
3. Explicitly set `cost_estimation_enabled` to `false`, verify the change is made.

```
$ TESTARGS="-run TestAccTFEOrganization_\(update\|basic\)" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEOrganization_\(update\|basic\) -timeout 15m
=== RUN   TestAccTFEOrganization_basic
--- PASS: TestAccTFEOrganization_basic (1.79s)
=== RUN   TestAccTFEOrganization_update
--- PASS: TestAccTFEOrganization_update (3.97s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	6.011s
```

### DIFF
--- a/tfe/resource_tfe_organization.go
+++ b/tfe/resource_tfe_organization.go
@@ -61,7 +61,7 @@ func resourceTFEOrganization() *schema.Resource {
 			"cost_estimation_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				Default:  true,
 			},
 		},
 	}
--- a/tfe/resource_tfe_organization.go
+++ b/tfe/resource_tfe_organization.go
@@ -75,13 +75,8 @@ func resourceTFEOrganizationCreate(d *schema.ResourceData, meta interface{}) err
 
 	// Create a new options struct.
 	options := tfe.OrganizationCreateOptions{
-		Name:                   tfe.String(name),
-		Email:                  tfe.String(d.Get("email").(string)),
-		SessionTimeout:         tfe.Int(d.Get("session_timeout_minutes").(int)),
-		SessionRemember:        tfe.Int(d.Get("session_remember_minutes").(int)),
-		CollaboratorAuthPolicy: tfe.AuthPolicy(tfe.AuthPolicyType(d.Get("collaborator_auth_policy").(string))),
-		OwnersTeamSAMLRoleID:   tfe.String(d.Get("owners_team_saml_role_id").(string)),
-		CostEstimationEnabled:  tfe.Bool(d.Get("cost_estimation_enabled").(bool)),
+		Name:  tfe.String(name),
+		Email: tfe.String(d.Get("email").(string)),
 	}
 
 	log.Printf("[DEBUG] Create new organization: %s", name)
--- a/tfe/resource_tfe_organization.go
+++ b/tfe/resource_tfe_organization.go
@@ -57,6 +57,12 @@ func resourceTFEOrganization() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+
+			"cost_estimation_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
@@ -105,6 +111,7 @@ func resourceTFEOrganizationRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("session_remember_minutes", org.SessionRemember)
 	d.Set("collaborator_auth_policy", org.CollaboratorAuthPolicy)
 	d.Set("owners_team_saml_role_id", org.OwnersTeamSAMLRoleID)
+	d.Set("cost_estimation_enabled", org.CostEstimationEnabled)
 
 	return nil
 }
@@ -136,6 +143,11 @@ func resourceTFEOrganizationUpdate(d *schema.ResourceData, meta interface{}) err
 	// If owners_team_saml_role_id is supplied, set it using the options struct.
 	if ownersTeamSAMLRoleID, ok := d.GetOk("owners_team_saml_role_id"); ok {
 		options.OwnersTeamSAMLRoleID = tfe.String(ownersTeamSAMLRoleID.(string))
+	}
+
+	// If cost_estimation_enabled is supplied, set it using the options struct.
+	if costEstimationEnabled, ok := d.GetOkExists("cost_estimation_enabled"); ok {
+		options.CostEstimationEnabled = tfe.Bool(costEstimationEnabled.(bool))
 	}
 
 	log.Printf("[DEBUG] Update configuration of organization: %s", d.Id())
--- a/tfe/resource_tfe_organization.go
+++ b/tfe/resource_tfe_organization.go
@@ -75,8 +75,13 @@ func resourceTFEOrganizationCreate(d *schema.ResourceData, meta interface{}) err
 
 	// Create a new options struct.
 	options := tfe.OrganizationCreateOptions{
-		Name:  tfe.String(name),
-		Email: tfe.String(d.Get("email").(string)),
+		Name:                   tfe.String(name),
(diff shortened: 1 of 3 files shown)


cooltoast:
@chrisarcand @koikonom does this PR look ok?

cooltoast:
@chrisarcand how about now?? 😄 

I'd love to use this new variable for some TFE management I'm doing

website/docs/r/organization.html.markdown
```diff
@@ -34,6 +34,7 @@ The following arguments are supported:
 * `collaborator_auth_policy` - (Optional) Authentication policy (`password`
   or `two_factor_mandatory`). Defaults to `password`.
 * `owners_team_saml_role_id` - (Optional) The name of the "owners" team.
+* `cost_estimation_enabled` - (Optional) Whether or not the cost estimation feature is enabled for all workspaces in the organization. Defaults to `false`.
```
chrisarcand, line 37: This was actually recently changed to `true`, and the documentation is out of date 😖  I've [opened a PR to get that changed](https://github.com/hashicorp/terraform-website/pull/1497), and provided that the documentation there doesn't need tweaking, we can put that here as well:

```suggestion
* `cost_estimation_enabled` - (Optional) Whether or not the cost estimation feature is enabled for all workspaces in the organization. Defaults to true on Terraform Cloud when the organization is on the Teams & Governance tier (else, the value is false and cannot be changed) and always true on Terraform Enterprise.
```

chrisarcand (changes requested):
Looking good, thank you! Just a couple tweaks to make:

website/docs/r/organization.html.markdown
cooltoast: > always true on Terraform Enterprise.

Ah I see, thanks for clarifying that point, since I was trying to observe the behavior (`1a` under the "Testing Plan").

tfe/resource_tfe_organization.go
```diff
@@ -138,6 +145,11 @@ func resourceTFEOrganizationUpdate(d *schema.ResourceData, meta interface{}) err
 		options.OwnersTeamSAMLRoleID = tfe.String(ownersTeamSAMLRoleID.(string))
 	}
 
+	// If cost_estimation_enabled is supplied, set it using the options struct.
+	if costEstimationEnabled, ok := d.GetOkExists("cost_estimation_enabled"); ok {
+		options.CostEstimationEnabled = tfe.Bool(costEstimationEnabled.(bool))
+	}
```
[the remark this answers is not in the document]
cooltoast: Will do!

website/docs/r/organization.html.markdown
chrisarcand: Gah, to be honest I had forgotten about the Terraform Enterprise Admin option you mention 😅 (I work primarily on Terraform Cloud).

There's yet one more caveat in that this value should behave as I mention above, _but_ in TFE even if it's `true`, the Admin setting you mention must also be enabled for runs to execute a cost estimate. I'm not sure if that detail is important here or not, but we could add it in:

```
In Terraform Cloud, defaults to true when the organization is on the Teams & Governance tier (else, the value is false and cannot be changed). Always defaults to true in Terraform Enterprise; Cost Estimation must also be enabled by Site Administrators.
```

chrisarcand: After a bit of worthsmithing, I think this is the clearest:

`Whether or not the cost estimation feature is enabled for all workspaces in the organization. Defaults to true. In Terraform Cloud, this value is always false and ignored if the organization does not have Teams & Governance features. In Terraform Enterprise, cost estimation must also be enabled in Site Administration.`

cooltoast: Ok sure, but I'll defer changing this until you and @nfagerlund decide on the final copy?

Looks like he added a [new suggestion](https://github.com/hashicorp/terraform-website/pull/1497#discussion_r523266373) recently.

tfe/resource_tfe_organization.go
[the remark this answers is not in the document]
cooltoast: ^ Added this to the `Create`, as well as the other missing attributes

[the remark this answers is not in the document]
cooltoast: Actually this is at the bottom of the `Create`, and already has the logic for those optional params, so we don't need to set it on `Create` right?
```
return resourceTFEOrganizationUpdate(d, meta)
```

I added a unit test (`TestAccTFEOrganization_full`) to cover this case, where we want to create an Organization with the optional params set.

cooltoast:
Tests passing
```
TESTARGS="-run TestAccTFEOrganization_.* -v" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEOrganization_.* -v -timeout 15m
=== RUN   TestAccTFEOrganization_basic
--- PASS: TestAccTFEOrganization_basic (1.93s)
=== RUN   TestAccTFEOrganization_full
--- PASS: TestAccTFEOrganization_full (1.81s)
=== RUN   TestAccTFEOrganization_update
--- PASS: TestAccTFEOrganization_update (4.27s)
=== RUN   TestAccTFEOrganization_import
--- PASS: TestAccTFEOrganization_import (1.84s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	10.134s
```

tfe/resource_tfe_organization.go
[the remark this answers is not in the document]
chrisarcand: Ahhh you're right, I missed that, interesting. Good catch!

I do really dislike that, as it's a needless extra request in an operation that very much supports those options on the create request itself. Still, that's a thing beyond the scope of your changes here and I think you've made the right changes to what's already there 👍🏻

chrisarcand (changes requested):
Everything looks great! Here's the final draft of the documentation for this option, so once this is in, this will be ready to go 🚀 

> Whether or not the cost estimation feature is enabled for all workspaces in the organization. Defaults to true. In a Terraform Cloud organization which does not have Teams & Governance features, this value is always false and cannot be changed. In Terraform Enterprise, Cost Estimation must also be enabled in Site Administration.

cooltoast:
@chrisarcand thanks for the review, I updated the copy, I think we're all set here?

chrisarcand:
Yup, I believe so! I did realize one thing we might want to change though: Because of the odd nature of defaults by platform, it might make sense to let the value be `Computed`, letting the platform determine the default if an argument was not given. 

For the purposes of avoiding another messaging cycle, I'm just going to merge it as-is and make this change myself, if appropriate. 

Thanks so much for your contribution! I'll get this in the next release. 🙇🏻‍♂️

chrisarcand (approved):

cooltoast:
I could have changed it to `Computed` if you wanted me to 😄 

Anyways thanks for reviewing and merging! I'll start using this variable now 🚀

chrisarcand:
> I could have changed it to Computed if you wanted me to 😄

Ah for sure! 💞  I was just avoiding another round of review for something so simple, to get this out ASAP! 👍🏻 Thanks again for the contribution!

PR #241, opened 2020-11-13
Enable setting SSH private key on all available OAuth clients for cloning Git submodules

cooltoast:
## Description

I have a VCS Provider (Gitlab EE) that I need to clone submodules for, and according to the [docs](https://www.terraform.io/docs/cloud/vcs/index.html#ssh-keys), I need to specify the private key for my OAuth client.

> However, if the organization repositories include Git submodules that can only be accessed via SSH, an SSH key can be added along with the OAuth credentials.

The problem is that the `tfe_oauth_client` resource is blocking me from doing so, since I do not use Azure DevOps Server.

I've manually set the private key in TFE to verify this can work, but I'd like this resource to let me do so from Terraform.

I am also happy to update the [API docs](https://github.com/hashicorp/terraform-website/blob/master/content/source/docs/cloud/api/oauth-clients.html.md) as well, to clarify that the private key is required for ADO Server and Git submodules.

## Testing plan
Verified I can set private key on OAuth client manually (as described in docs above), and TFE is able to clone submodules successfully.

### DIFF
--- a/vendor/github.com/hashicorp/go-tfe/oauth_client.go
+++ b/vendor/github.com/hashicorp/go-tfe/oauth_client.go
@@ -140,9 +140,6 @@ func (o OAuthClientCreateOptions) valid() error {
 	if o.ServiceProvider == nil {
 		return errors.New("service provider is required")
 	}
-	if validString(o.PrivateKey) && *o.ServiceProvider != *ServiceProvider(ServiceProviderAzureDevOpsServer) {
-		return errors.New("Private Key can only be present with Azure DevOps Server service provider")
-	}
 	return nil
 }
 
--- a/website/docs/r/oauth_client.html.markdown
+++ b/website/docs/r/oauth_client.html.markdown
@@ -52,7 +52,7 @@ The following arguments are supported:
 * `http_url` - (Required) The homepage of your VCS provider (e.g.
   `https://github.com` or `https://ghe.example.com`).
 * `oauth_token` - (Required) The token string you were given by your VCS provider.
-* `private_key` - (Required for `ado_server`) The text of the private key associated with your Azure DevOps Server account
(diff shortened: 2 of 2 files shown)


cooltoast:
## Update

I'm entirely unable to add a private key programmatically, as this seems to be a problem with the [TFE API](https://www.terraform.io/docs/cloud/api/oauth-clients.html#create-an-oauth-client) as well.

My only way to enable submodule clones for Gitlab EE is manually through the OAuth (VCS Provider) settings in the TFE UI.

What is our best course of action here to enable this programmatically (via API/Terraform)?

___

(Unsuccessful request)
```
$ cat payload.json
{
  "data": {
    "type": "oauth-clients",
    "attributes": {
      "service-provider": "gitlab_enterprise_edition",
      "http-url": "...",
      "api-url": "...",
      "oauth-token-string": "...",
      "private-key": "-----BEGIN RSA PRIVATE KEY----    ...    -----END RSA PRIVATE KEY-----"
    }
  }
}

$ curl \
  --header "Authorization: Bearer $TOKEN" \
  --header "Content-Type: application/vnd.api+json" \
  --request POST \
  --data @payload.json \
  https://<TFE_HOSTNAME>/api/v2/organizations/test/oauth-clients

{"errors":[{"status":"422","title":"invalid attribute","detail":"Private key cannot be used with given service provider gitlab_enterprise_edition","source":{"pointer":"/data/attributes/private-key"}}]}%
```

cooltoast:
I'm gonna close this as it's 1 year old now 😅

kreynoldsf5:
@cooltoast, 

I'm having the same issue on a "github.com" VCS provider. Since the PR went nowhere, did you find a way to do this in the provider (and not just in the GUI)?

PR #242, opened 2020-11-17
Add support for Terraform Cloud Agents

### LINKED ISSUE #204: [Feature Request] Updated Workspace to allow for "agent" Operation type.
With the release of TFC for Business, workspaces within TFC can now be configured to run on private agents.

Currently the `tfe_workspace` resource has an argument `operations` that is a boolean.

The request is that this argument is deprecated and a new `execution-mode` argument added, which would be a string that must be one of the following: ["remote", "local", "agent"].

This will allow newer TFC4B users to properly configure a workspace and leverage the new agent feature.

You can manually set this setting with the following payload to `https://app.terraform.io/api/v2/organizations/${tfe_org_name}/workspaces/${WS_ID}`:

```
{
  "data": {
    "id": "${WS_ID}",
    "type": "workspaces",
    "attributes": {
      "environment": "Default",
      "execution-mode": "agent",
      "auto-apply": true,
      "queue_all_runs": true
    },
    "relationships": {
      "agent-pool": {
        "data": {
          "type": "agent-pools",
          "id": "${AGENT_POOL_ID}"
        }
      }
    }
  }
}
```

chrisarcand:
## Description

This adds support for [Terraform Cloud Agents](https://www.terraform.io/docs/cloud/agents/index.html), now that [agent pools have a full resource lifecycle](https://www.terraform.io/docs/cloud/agents/index.html#managing-agent-pools) and can be managed via Terraform. Specifically this adds a `tfe_agent_pool` resource and datasource (https://www.terraform.io/docs/cloud/api/agents.html) and allows configuring them for use in a Terraform Cloud workspace with the new `execution-mode` attribute.

Closes #204 

## Testing plan

* Agent pool management should behave as you expect (it's a simple resource)
* Workspace configuration with `execution_mode` and `agent_pool_id` should behave smoothly, allowing for continued use of the deprecated `operations` argument or a seamless switch to `execution_mode`. Their use should be mutually exclusive.  The only known rough edge I have found is moving from `execution_mode = "agent"` back to `operations = true/false` fails schema validations, because of the limitations of validating existing state. This is acceptable though, as if you're already using execution_mode there is _no_ reason to revert back to operations.
* Agent pool imports should work: simply import by ID, you don't even need the org name: `terraform import tfe_agent_pool.name apool-xxxx`

## External links

* Requires https://github.com/hashicorp/go-tfe/pull/153

## TODO

- [x] Provider documentation
- [x] ~Deprecate `operations`? Might save that for a separate change.~
- [x] Add Terraform Enterprise disclaimers to documentation
- [x] Bump go-tfe, after verifying CI passes with a ref.

### DIFF
--- a/tfe/resource_tfe_agent_pool.go
+++ b/tfe/resource_tfe_agent_pool.go
@@ -51,7 +51,7 @@ func resourceTFEAgentPoolCreate(d *schema.ResourceData, meta interface{}) error 
 
 	d.SetId(agentPool.ID)
 
-	return resourceTFEAgentPoolUpdate(d, meta)
+	return resourceTFEAgentPoolRead(d, meta)
 }
 
 func resourceTFEAgentPoolRead(d *schema.ResourceData, meta interface{}) error {
--- a/tfe/resource_tfe_agent_pool.go
+++ b/tfe/resource_tfe_agent_pool.go
@@ -1,0 +1,107 @@
+package tfe
+
+import (
+	"fmt"
+	"log"
+
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceTFEAgentPool() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceTFEAgentPoolCreate,
+		Read:   resourceTFEAgentPoolRead,
+		Update: resourceTFEAgentPoolUpdate,
+		Delete: resourceTFEAgentPoolDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"organization": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceTFEAgentPoolCreate(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	// Get the name and organization.
+	name := d.Get("name").(string)
+	organization := d.Get("organization").(string)
+
+	// Create a new options struct.
+	options := tfe.AgentPoolCreateOptions{
+		Name: tfe.String(name),
+	}
+
+	log.Printf("[DEBUG] Create new agent pool for organization: %s", organization)
+	agentPool, err := tfeClient.AgentPools.Create(ctx, organization, options)
+	if err != nil {
+		return fmt.Errorf(
+			"Error creating agent pool %s for organization %s: %v", name, organization, err)
+	}
+
+	d.SetId(agentPool.ID)
+
+	return resourceTFEAgentPoolUpdate(d, meta)
+}
+
+func resourceTFEAgentPoolRead(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	log.Printf("[DEBUG] Read configuration of agent pool: %s", d.Id())
+	agentPool, err := tfeClient.AgentPools.Read(ctx, d.Id())
+	if err != nil {
+		if err == tfe.ErrResourceNotFound {
+			log.Printf("[DEBUG] agent pool %s does no longer exist", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading configuration of agent pool %s: %v", d.Id(), err)
+	}
+
+	// Update the config.
+	d.Set("name", agentPool.Name)
+
+	return nil
+}
+
+func resourceTFEAgentPoolUpdate(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	// Create a new options struct.
+	options := tfe.AgentPoolUpdateOptions{
+		Name: tfe.String(d.Get("name").(string)),
+	}
+
+	log.Printf("[DEBUG] Update agent pool: %s", d.Id())
+	_, err := tfeClient.AgentPools.Update(ctx, d.Id(), options)
+	if err != nil {
+		return fmt.Errorf("Error updating agent pool %s: %v", d.Id(), err)
+	}
+
+	return resourceTFEAgentPoolRead(d, meta)
+}
+
+func resourceTFEAgentPoolDelete(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	log.Printf("[DEBUG] Delete agent pool: %s", d.Id())
+	err := tfeClient.AgentPools.Delete(ctx, d.Id())
+	if err != nil {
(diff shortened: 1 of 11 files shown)


tfe/resource_tfe_agent_pool.go
```diff
@@ -0,0 +1,107 @@
...
+	log.Printf("[DEBUG] Create new agent pool for organization: %s", organization)
+	agentPool, err := tfeClient.AgentPools.Create(ctx, organization, options)
+	if err != nil {
+		return fmt.Errorf(
+			"Error creating agent pool %s for organization %s: %v", name, organization, err)
+	}
+
+	d.SetId(agentPool.ID)
+
+	return resourceTFEAgentPoolUpdate(d, meta)
```
lafentres, line 54: i think this should be a call to `resourceTFEAgentPoolRead`
```suggestion
	return resourceTFEAgentPoolRead(d, meta)
```

tfe/resource_tfe_agent_pool.go
```diff
@@ -0,0 +1,107 @@
...
+			log.Printf("[DEBUG] agent pool %s does no longer exist", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading configuration of agent pool %s: %v", d.Id(), err)
+	}
+
+	// Update the config.
+	d.Set("name", agentPool.Name)
+
```
lafentres, line 73: not sure if it is absolutely necessary here but i usually set all the attributes on read (i think this is most important when you have an importer though)
```suggestion
	// Update the config.
	d.Set("name", agentPool.Name)
	d.Set("organization", agentPool.Organization)

```

tfe/resource_tfe_agent_pool.go
```diff
@@ -0,0 +1,107 @@
...
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceTFEAgentPool() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceTFEAgentPoolCreate,
+		Read:   resourceTFEAgentPoolRead,
+		Update: resourceTFEAgentPoolUpdate,
+		Delete: resourceTFEAgentPoolDelete,
```
lafentres, line 16: what do you think about adding the ability to import agent pools? since we didn't have provider support the first time around?

tfe/resource_tfe_agent_pool_test.go
```diff
@@ -0,0 +1,157 @@
...
+const testAccTFEAgentPool_update = `
+resource "tfe_organization" "foobar" {
+  name  = "tst-terraform"
+  email = "admin@company.com"
+}
+
+resource "tfe_agent_pool" "foobar" {
+  name         = "agent-pool-updated"
+  organization = "${tfe_organization.foobar.id}"
+}`
```
lafentres, line 157: it would be good for all of these org names to have a random int on the end like the other resources to avoid tests breaking whenever some of the tests fail

tfe/resource_tfe_workspace_test.go
```diff
@@ -903,6 +989,82 @@ resource "tfe_workspace" "foobar" {
...
+}
+
+resource "tfe_workspace" "foobar" {
+  name         = "workspace-test"
+  organization = "${tfe_organization.foobar.id}"
+  execution_mode = "agent"
+  agent_pool_id = "${tfe_agent_pool.foobar.id}"
+}`, rInt)
+}
+
```
lafentres, line 1067: same thing here, i think the org names need the random int added onto the end.

tfe/resource_tfe_agent_pool.go
chrisarcand: It "should be" update, based on the [template I originally borrowed from](https://github.com/hashicorp/terraform-provider-tfe/blob/0e8c31650fd1ef963226b96e6b2c03d0e5f4098d/tfe/resource_tfe_ssh_key.go#L61)

I'm guessing this is another case of intended before I really don't like, [which I discovered in another PR recently](https://github.com/hashicorp/terraform-provider-tfe/pull/239#discussion_r525373874). 😕

This is all to say I agree, and think we should fix those other places when time permits 😑  Good catch.

tfe/resource_tfe_agent_pool.go
chrisarcand: Yes, absolutely. Wasn't a concern months ago when I wrote this so I totally forgot, let's do it!

tfe/resource_tfe_agent_pool_test.go
chrisarcand: Doh this was just forgotten - note (during a rebase) I already converted all the configurations to use functions and not consts, I just forgot to actually add the interpolated value in the string 🤦  Good catch.

tfe/resource_tfe_workspace.go
```diff
@@ -55,9 +57,23 @@ func resourceTFEWorkspace() *schema.Resource {
...
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"operations"},
+			},
+
+			"agent_pool_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"operations"},
```
lafentres, line 76: nit: we generally try to keep these in alphabetical order after the most important identifying attributes (like name and organization). so agent_pool_id should be up above auto_apply and execution_mode should be above file_triggers_enabled

chrisarcand: Huh. I kept these close to `operations` intentionally but that sounds good to me 👍🏻

tfe/resource_tfe_workspace.go
```diff
@@ -54,16 +62,24 @@ func resourceTFEWorkspace() *schema.Resource {
 				Default:  false,
 			},
 
+			"execution_mode": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"operations"},
+			},
```
lafentres, line 70: missed this before but we should probably validate the execution mode values. it'd be better if we'd stored these as constants in go-tfe but we can always fix that later
something like this should work:
```suggestion
			"execution_mode": {
				Type:          schema.TypeString,
				Optional:      true,
				Computed:      true,
				ConflictsWith: []string{"operations"},
				ValidateFunc: validation.StringInSlice(
					[]string{
						"agent",
						"local",
						"remote",
					},
					false,
				),
			},
```

lafentres (approved):
i ran this through several different test scenarios and it all looks good to me (aside from that small note about validating execution mode).

i did get some warnings locally about needing to run go mod vendor when building this. it might just be me so this might be something you can ignore. 
i'm not sure if i'm just on a different go version or if some of the vendor directories didn't get committed with the last PR but it might be worth running go mod vendor again and making sure all the directories are added/tracked (i've had this get messed up on me before).

PR #246, opened 2020-11-23
Upgrade SDK to v2.x & Drop Terraform 0.11 support

### LINKED ISSUE #108: Update TFE provider to use terraform v0.12.x syntax
Since we are getting deprecation warnings for v0.11.x, we should already have TFE also using the newer 0.12 syntax.

chrisarcand:
## Description

This PR upgrades the provider to use the v2.x of the [Terraform SDK](https://github.com/hashicorp/terraform-plugin-sdk). This allows for a number of benefits for provider development; the [V2 upgrade guide](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html) is a great overview of some of the initial changes.

### ⚠️ Breaking change ⚠️

As a result of this major SDK upgrade, only Terraform CLI 0.12 and later can be supported past the current release (0.23.0). Users who wish to continue using Terraform 0.11 may continue to use version 0.23.x or earlier. This move will also consolidate all example code to use Terraform 0.12 syntax. Examples using the older syntax style can still be viewed by browsing older versions of the provider in the [Terraform Registry](https://registry.terraform.io/providers/hashicorp/tfe/0.23.0).

[The announcement that official HashiCorp-maintained providers would be dropping support for Terraform 0.11](https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers) was made in November 2019, over a year ago.

### In Detail

* Upgrades the Terraform SDK to version 2.x; a major version bump; makes all necessary changes to support this transition. Any APIs that are merely deprecated but not removed are _not_ changed. For example, only the required [helper/schema context-aware functions](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#more-support-for-context-context) changes were made (all resources still use any deprecated equivalents, `CreateFunc`, etc).
* Drops support for Terraform < 0.12; updates documentation and deprecated interpolation syntax.
* Due to the SDK upgrade, it was discovered that some of the test configurations aren't actually compatible with Terraform 0.13 (which is installed automatically by the SDK as latest), causing CI to fail. Instead of making even more changes, I have instead pinned acceptance tests to use Terraform 0.12 (as it was before, or maybe 0.11? I don't know, with the legacy SDK) which still pass. Fixing these configs would be good to do later.

## Upgrading

Upgrade details will be found in the release notes, but in case you've found your way here, instead:

No configuration changes should be necessary _in using this provider_ (changes might be necessary in upgrading to Terraform 0.12+, though). As mentioned in this provider's documentation, it is considered best practice to pin a provider version to avoid bringing in potentially breaking changes automatically. To remain on previous versions of the provider until such time that you are able to accommodate upgrading to Terraform 0.12+, pin the installed version. Remember that as this provider is pre-1.0, 'minor' changes may signal breaking changes. In this example, we pin our configuration to only use patch versions of 0.23.x:

**Terraform 0.13+**

```hcl
terraform {
  required_providers {
    tfe = {
      version = "~> 0.23.0"
    }
  }
}
```

**Terraform 0.12**

```hcl
provider "tfe" {
  version = "~> 0.23.0"
  ...
}
```

Full documentation on how to pin a version of a provider can be found on the Terraform Website.

## Testing plan

All resources and APIs should be thoroughly tested to ensure no regressions with the major SDK upgrade.

---

Closes #108


website/docs/index.html.markdown
```diff
@@ -15,6 +15,8 @@ Terraform Cloud, this provider supports both Cloud and Enterprise.
 
 Use the navigation to the left to read about the available resources.
 
+Since v0.24.0, this provider requires Terraform >= 0.12
```
radditude, line 18: Tiny nitpick for a missing period:

```suggestion
Since v0.24.0, this provider requires Terraform >= 0.12.
```

radditude (approved):
Looks good, and local testing was all green!

PR #254, opened 2020-12-14
Update CHANGELOG.md with next unreleased version

lafentres:
## Description

Normally the release bot does this but it didn't this time, for whatever reason. The unreleased version is normally necessary for new releases to work properly so I'm fixing it. 

It is possible this isn't necessary anymore but either way, it is also helpful to have the next version so we can update the changelog in PRs as we go.

## Testing plan

No need to test, just a changelog update

No need to run acceptance tests, just a changelog update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.24.0 (Unreleased)
+
+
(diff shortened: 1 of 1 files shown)


chrisarcand (approved):
Weird.

PR #262, opened 2021-01-21
Add IP Ranges datasource

chrisarcand:
## Description

Adds a new `tfe_ip_ranges` datasource to fetch a list of Terraform Cloud and Enterprise's IP ranges from the [IP Ranges API](https://www.terraform.io/docs/cloud/api/ip-ranges.html).

```
» TF_ACC=1 go test -v ./tfe -timeout=30m -run TestAccTFEIPRanges
=== RUN   TestAccTFEIPRangesDataSource_basic
--- PASS: TestAccTFEIPRangesDataSource_basic (4.73s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/tfe (cached)
```

### DIFF
--- a/tfe/data_source_ip_ranges_test.go
+++ b/tfe/data_source_ip_ranges_test.go
@@ -1,0 +1,38 @@
+package tfe
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccTFEIPRangesDataSource_basic(t *testing.T) {
+	ipRegex := regexp.MustCompile(`^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\/([1-9]|[1-9][0-9])$`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccTFEIPRangesDataSourceConfig(),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttrSet("data.tfe_ip_ranges.ips", "id"),
+					resource.TestCheckResourceAttrSet("data.tfe_ip_ranges.ips", "api.0"),
+					resource.TestMatchResourceAttr("data.tfe_ip_ranges.ips", "api.0", ipRegex),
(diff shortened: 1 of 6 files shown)


website/tfe.erb
```diff
@@ -17,6 +17,10 @@
                             <a href="/docs/providers/tfe/d/agent_pool.html">tfe_agent_pool</a>
                         </li>
 
+                        <li<%= sidebar_current("docs-datasource-tfe-ip-ranges") %>>
+                            <a href="/docs/providers/tfe/d/ip_ranges.html">tfe_agent_pool</a>
```
jgrumboe, line 21:
```suggestion
                            <a href="/docs/providers/tfe/d/ip_ranges.html">ip_ranges</a>
```

jgrumboe (commented):
small fix

website/tfe.erb
chrisarcand: Whoops! Good catch.

jgrumboe: 🤦 but mine was also not completely correct 😁

tfe/data_source_ip_ranges_test.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccTFEIPRangesDataSource_basic(t *testing.T) {
+	ipRegex := regexp.MustCompile(`^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\/([1-9]|[1-9][0-9])$`)
```
radditude, line 12: Nit: I love a good complicated regex, but since we really only care that this _is_ an IP address and don't necessarily care about ensuring that it's valid (that's the API's job!), what if we made this a little fuzzier and had it just match the format of an IPv4 address?

```suggestion
	ipRegex := regexp.MustCompile(`^[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}/[\d]{1,3}$`)
```

radditude (approved):
One regex nit, but that's really a judgement call, so I'll leave it up to you!

One more thought: this API isn't available by default in our local environment, so this test will fail if someone runs the full test suite locally - do we have any prior art/recommendations about documenting tests that are expected to fail in some circumstances like that?

tfe/data_source_ip_ranges_test.go
chrisarcand: Totally! TBH I didn't even write this regex, it's just cribbed because I was lazy 😜  Yours is very reasonable and what I should have done to start 👍🏻

Solid feedback.

PR #263, opened 2021-01-21
Drop website-test make target

chrisarcand:
## Description

This make target from the website project was dropped in https://github.com/hashicorp/terraform-website/commit/6d41be434cf85392bc9de773d8a5a8d571a195ad and is causing a failing Travis job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ install:
 script:
   - make test
   - make vet
-  - make website-test
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


nfagerlund (approved):
Looks good! For posterity's sake: this was a link checking job for the provider docs, but it relied on the (unreliable, busted) old link checking tasks from terraform-website, and provider docs are no longer published to terraform.io via the terraform-website tooling. Link checking is still a worthy thing to do, but we should implement it with tooling that fits the new provider docs environment on the Registry.

radditude (approved):

PR #265, opened 2021-02-03
[DOCS] Some addition clarification on 'queue_all_runs'

chrisarcand:
## Description

Just some additional clarification on this setting and why the default here in the provider is the opposite of the API default.

### DIFF
--- a/website/docs/d/workspace.html.markdown
+++ b/website/docs/d/workspace.html.markdown
@@ -35,7 +35,9 @@ In addition to all arguments above, the following attributes are exported:
 * `auto_apply` - Indicates whether to automatically apply changes when a Terraform plan is successful.
 * `file_triggers_enabled` - Indicates whether runs are triggered based on the changed files in a VCS push (if `true`) or always triggered on every push (if `false`).
 * `operations` - Indicates whether the workspace is using remote execution mode. Set to `false` to switch execution mode to local. `true` by default.
-* `queue_all_runs` - Indicates whether all runs should be queued.
+* `queue_all_runs` - Indicates whether the workspace will automatically perform runs
+  in response to webhooks immediately after its creation. If `false`, an initial run must
+  be manually queued to enable future automatic runs.
 * `speculative_enabled` - Indicates whether this workspace allows speculative plans.
 * `ssh_key_id` - The ID of an SSH key assigned to the workspace.
 * `terraform_version` - The version of Terraform used for this workspace.
--- a/website/docs/r/workspace.html.markdown
+++ b/website/docs/r/workspace.html.markdown
@@ -70,9 +70,12 @@ The following arguments are supported:
 * `operations` - **Deprecated** Whether to use remote execution mode. When set to `false`, the workspace will 
   be used for state storage only. Defaults to `true`. This value _must not_ be provided if `execution_mode` is 
   provided.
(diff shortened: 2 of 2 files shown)


website/docs/r/workspace.html.markdown
```diff
@@ -70,9 +70,11 @@ The following arguments are supported:
...
   be used for state storage only. Defaults to `true`. This value _must not_ be provided if `execution_mode` is 
   provided.
-* `queue_all_runs` - (Optional) Whether all runs should be queued. When set
-  to `false`, runs triggered by a VCS change will not be queued until at least
-  one run is manually queued. Defaults to `true`.
+* `queue_all_runs` - (Optional) Whether all runs should be queued. When set to `false`, runs triggered by a webhook
+  (such as a commit in VCS) will not be queued until at least one run is manually queued. Although the Terraform Cloud API
+  defaults to `false` as a feature (allowing other settings to be changed before accepting webhooks), this provider
+  defaults the value to `true` as any workspace provisioned with `false` would need to then have a run manually queued out-of-band
+  before accepting webhooks.
```
nfagerlund, line 77:
```suggestion
* `queue_all_runs` - (Optional) Whether the workspace should start automatically performing
  runs immediately after its creation. When set to `false`, runs triggered by a webhook
  (such as a commit in VCS) will not be queued until at least one run has been manually
  queued. Defaults to `true`. **Note:** In the Terraform Cloud API, this defaults to `false`
  instead, since setting variables or other settings might require additional imperative steps.
```

website/docs/d/workspace.html.markdown
```diff
@@ -35,7 +35,7 @@ In addition to all arguments above, the following attributes are exported:
 * `auto_apply` - Indicates whether to automatically apply changes when a Terraform plan is successful.
 * `file_triggers_enabled` - Indicates whether runs are triggered based on the changed files in a VCS push (if `true`) or always triggered on every push (if `false`).
 * `operations` - Indicates whether the workspace is using remote execution mode. Set to `false` to switch execution mode to local. `true` by default.
-* `queue_all_runs` - Indicates whether all runs should be queued.
+* `queue_all_runs` - Indicates whether all runs should be queued immediately, or an initial run should be manually queued before runs are automatically triggered by webhooks (such as changes in VCS).
```
nfagerlund, line 38:
```suggestion
* `queue_all_runs` - Indicates whether the workspace can automatically perform runs
  (in response to webhooks) immediately after its creation. If `false`, an initial run must
  be manually queued to enable future automatic runs.
```

nfagerlund (approved):
I suggest tweaking to get rid of the "all runs should be queued" language -- it makes it sound more like an enduring behavior of the workspace, and I think we can make this clearer by emphasizing that this is really only about what happens in between creation and getting your vars set up.

nfagerlund:
oh, I'm not a "reviewer with write access." 😢

chrisarcand:
@nfagerlund Good points 👍 I did tweak yours slightly to retain what I think is an important point: to emphasize that the provider is specifically using a different default from the canonical API because it makes no sense to ever provision a workspace with the API default (the "pauses as a feature" part about setting of variables is of no consequence to the provider, and is covered in the API documentation itself too)

nfagerlund (approved):

PR #271, opened 2021-02-16
Adds description parameter to TFE Workspace

petems:
* Requires go-tfe update

## Description

Adds description to workspace 

## External links

- https://github.com/hashicorp/go-tfe/issues/170

https://www.terraform.io/docs/cloud/api/workspaces.html#request-body

data.attributes.description | string | (nothing) | A description for the workspace.
-- | -- | -- | --

Required in create and read

### DIFF
--- a/tfe/data_source_workspace_test.go
+++ b/tfe/data_source_workspace_test.go
@@ -26,6 +26,8 @@ func TestAccTFEWorkspaceDataSource_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						"data.tfe_workspace.foobar", "organization", orgName),
 					resource.TestCheckResourceAttr(
+						"data.tfe_workspace.foobar", "description", "provider-testing"),
+					resource.TestCheckResourceAttr(
 						"data.tfe_workspace.foobar", "allow_destroy_plan", "false"),
 					resource.TestCheckResourceAttr(
 						"data.tfe_workspace.foobar", "auto_apply", "true"),
@@ -67,6 +69,7 @@ resource "tfe_organization" "foobar" {
 resource "tfe_workspace" "foobar" {
   name                  = "workspace-test-%d"
   organization          = tfe_organization.foobar.id
+  description           = "provider-testing"
   allow_destroy_plan    = false
   auto_apply            = true
   file_triggers_enabled = true
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -46,6 +46,11 @@ func resourceTFEWorkspace() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"description": {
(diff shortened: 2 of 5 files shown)


tfe/resource_tfe_workspace.go
```diff
@@ -46,6 +46,12 @@ func resourceTFEWorkspace() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  false,
```
chrisarcand, line 52: `false`? I think you want `""` here.

chrisarcand (changes requested):
* Needs some test additions 💞 
* Run a `go mod tidy` to remove the 0.12.0 refs, too.

tfe/resource_tfe_workspace.go
petems: Fixed! 👍

petems:
@chrisarcand Added tests 🖖

tfe/data_source_workspace_test.go
```diff
@@ -63,6 +65,7 @@ resource "tfe_organization" "foobar" {
 resource "tfe_workspace" "foobar" {
   name                  = "workspace-test-%d"
   organization          = tfe_organization.foobar.id
+  description						= "provider-testing"
```
chrisarcand, line 68:
```suggestion
  description           = "provider-testing"
```

tfe/resource_tfe_workspace.go
```diff
@@ -46,6 +46,12 @@ func resourceTFEWorkspace() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "",
```
chrisarcand, line 52: Sorry, I know we already changed this from `false` before....I don't believe this is necessary at all, right? Descriptions can be `null`; the zero value of the field here will naturally be `""` anyway.

tfe/resource_tfe_workspace_test.go
```diff
@@ -53,6 +53,47 @@ func TestAccTFEWorkspace_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(
+						"tfe_workspace.foobar", "trigger_prefixes.#", "0"),
+					resource.TestCheckResourceAttr(
+						"tfe_workspace.foobar", "working_directory", ""),
+				),
+			},
+		},
+	})
+}
+
```
chrisarcand, line 96: I don't think this simple field merits a separate expensive test; let's move it in to `TestAccTFEWorkspace_basic` with other basic arguments.

chrisarcand (changes requested):
Hey! Sorry for the delay.

chrisarcand (approved):

PR #273, opened 2021-02-18
Provider token missing: improve error message

omarismail:
## Description

Update the error message to inform the user on what to do to fix their missing token problem. 

## Testing plan

Do not include token, and ensure the error message appears.

### DIFF
--- a/tfe/provider.go
+++ b/tfe/provider.go
@@ -214,7 +214,7 @@ func getClient(tfeHost, token string, insecure bool) (*tfe.Client, error) {
 
 	// If we still don't have a token at this point, we return an error.
 	if token == "" {
-		return nil, fmt.Errorf("required token could not be found")
+		return nil, fmt.Errorf("Required token could not be found. Please set the token in the provider configuration, or setting the TFE_TOKEN environment variable.")
 	}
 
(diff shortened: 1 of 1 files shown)


tfe/provider.go
```diff
@@ -214,7 +214,7 @@ func getClient(tfeHost, token string, insecure bool) (*tfe.Client, error) {
 
 	// If we still don't have a token at this point, we return an error.
 	if token == "" {
-		return nil, fmt.Errorf("required token could not be found")
+		return nil, fmt.Errorf("Required token could not be found. Please set the token in the provider configuration, or using the TFE_TOKEN environment variable.")
```
chrisarcand, line 217: I just realized we should probably emphasize best practice and not accidentally advise people to put their token as a string literal in the configuration 😅  How about this?

> Required token could not be found. Please set the token using an input variable in the provider configuration block by using the TFE_TOKEN environment variable.

chrisarcand (changes requested):

tfe/provider.go
omarismail: Makes sense 😄

chrisarcand (approved):

PR #277, opened 2021-03-11
Add additional workspace fields

mwhooker:
## Description

Adds support for new fields in the workspace object

## Testing plan

Added a unit test, but it won't work because this feature is not generally available yet. It should pass once it is.

The tests pass against an environment with the feature available.

## External links

- [go-tfe PR](https://github.com/hashicorp/go-tfe/pull/192)
- [PR to bump go-tfe version](#284)

```
$ make testacc

...
```

## Note:

- [x]  update `go.mod` when the above go-tfe PR is merged


tfe/data_source_workspace_readme.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	rs, err := ioutil.ReadAll(readme)
+	if err != nil {
+		return fmt.Errorf("Error reading readme: %s", err)
+	}
+
+	d.Set("raw_markdown", rs)
+
+	d.SetId(workspaceID)
```
mwhooker, line 50: Is this what i should use in `SetId`? We don't have a separate ID available for the readme at the moment.

tfe/data_source_workspace_readme_test.go
```diff
@@ -0,0 +1,8 @@
+package tfe
+
+import (
+	"testing"
+)
+
+func TestAccTFEWorkspaceReadmeDataSource_basic(t *testing.T) {
```
mwhooker, line 7: Need to figure out how to test against a workspace that already has a readme

PR #286, opened 2021-03-25
Go tfe 1.13

mwhooker:
## Description

Bump `go-tfe` and make it compile.


tfe/client_mock.go
```diff
@@ -11,11 +12,13 @@ type workspaceNamesKey struct {
 }
 
 type mockWorkspaces struct {
+	workspaceID    string
```
radditude, line 15: Question about the naming of this field! It took me a minute to figure out what exactly this field was far and why it would be singular when everything else is plural 😅 (`mockWorkspaces`, `newMockWorkspaces`) - would it make sense to call it something like `defaultWorkspaceID` instead, to make that a little clearer?

radditude (commented):
Nice! Not a big deal to have one hardcoded ID for all mock workspaces for now since we're never creating more than one mock workspace at a time - one day we may need to add support for different mock workspaces having unique IDs, but this seems like a fine workaround for now.

One question about naming, but otherwise 👍🏻

radditude (approved):

mwhooker:
someone else will have to merge this for me:

> The base branch restricts merging to authorized users.

PR #287, opened 2021-03-25
Workspace additions

mwhooker:
## Description

Adds support for new fields in the workspace object

## Testing plan

Integration tests should capture these changes

## External links

- [go-tfe PR](https://github.com/hashicorp/go-tfe/pull/192)
- [PR to bump go-tfe version](#286)

I wasn't able to get these working locally. Perhaps we can use CI


tfe/data_source_workspace_test.go
```diff
@@ -47,6 +47,10 @@ func TestAccTFEWorkspaceDataSource_basic(t *testing.T) {
 						"data.tfe_workspace.foobar", "working_directory", "terraform/test"),
 
 					resource.TestCheckResourceAttrSet("data.tfe_workspace.foobar", "external_id"),
+					resource.TestCheckResourceAttrSet("data.tfe_workspace.foobar", "resource_count"),
```
radditude, line 50: I'm curious why these use [`TestCheckResourceAttrSet`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.5.0/helper/resource#TestCheckResourceAttrSet) instead of [`TestCheckResourceAttr`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.5.0/helper/resource#TestCheckResourceAttr)! The latter would let us verify that the attribute actually has the value we expect as well as checking that it's set. 

Granted, it's a somewhat academic difference in this case, since the test workspace won't have any runs and therefore the values will all be zero 😆

radditude (commented):
Two small things, then this will be ready for ✅ and merge!

tfe/data_source_workspace_test.go
mwhooker: I think I used `TestCheckResourceAttrSet` because I didn't have a way to test that they'd be anything but the zero-value. I can change to `TestCheckResourceAttr` if you think it would be better

radditude: That makes sense! Not a huge deal either way, but I'd lean towards using `TestCheckResourceAttr` since we do know the value - unlike, say, checking the workspace's ID, where we have no way of knowing what the external ID will be before it's generated.

I went ahead and pushed up a commit with that change: https://github.com/hashicorp/terraform-provider-tfe/commit/0de74c235378f76e478a68360cac1ee8d63ba941. If you're ok with it, I'll cherry-pick that commit into this branch and then smash that merge button.

radditude (dismissed):

mwhooker:
Thanks! feel free to merge

radditude (approved):

PR #292, opened 2021-04-16
Add Workspace remote state consumer

omarismail:
## Description

This adds `global_remote_state` and `remote_state_consumer_ids` to workspaces. Read more about [remote state sharing here](https://www.terraform.io/docs/cloud/workspaces/settings.html#remote-state-sharing)

## External Links

* [go-tfe PR](https://github.com/hashicorp/go-tfe/pull/201)
* [API Docs](https://www.terraform.io/docs/cloud/api/workspaces.html#get-remote-state-consumers)
* [Remote state sharing](https://www.terraform.io/docs/cloud/workspaces/settings.html#remote-state-sharing)

```
$ make testacc
```


go.mod
```diff
@@ -2,15 +2,42 @@ module github.com/hashicorp/terraform-provider-tfe
 
 go 1.16
 
+replace github.com/hashicorp/go-tfe => /Users/omarismail/work/go-tfe
```
omarismail, line 5: Once [this](https://github.com/hashicorp/go-tfe/pull/201) is merged, I'll remove this

tfe/resource_tfe_workspace_test.go
```diff
@@ -653,6 +653,40 @@ func TestAccTFEWorkspace_operationsAndExecutionModeInteroperability(t *testing.T
...
+						"tfe_workspace.foobar", "global_remote_state", "false"),
+					resource.TestCheckResourceAttr(
+						"tfe_workspace.foobar_remote", "global_remote_state", "false"),
+					resource.TestCheckResourceAttr(
+						"tfe_workspace.foobar", "remote_state_consumer_ids", workspaceRemote.ID),
+				),
+			},
+		},
+	})
+}
```
omarismail, line 688: Adding another test

go.mod
```diff
@@ -4,7 +4,6 @@ go 1.16
 
 require (
 	github.com/google/go-querystring v1.1.0 // indirect
-	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
```
omarismail, line 7: change from `go mod tidy`

tfe/resource_tfe_workspace.go
```diff
@@ -90,6 +90,18 @@ func resourceTFEWorkspace() *schema.Resource {
 				Default:  true,
 			},
 
+			"global_remote_state": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
chrisarcand, line 96: Because this isn't a true default and configurable for Terraform Enterprise, I think `Computed` should be set here instead of a `Default`.

[Computed](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed) is a schema behavior that in this case means that the provider cannot predict the value at plan-time and the value will be provided by the platform (Which will be TFC _or_ TFE - it could default to `false`!) in the absence of it being specified by the user in the configuration.

website/docs/r/workspace.html.markdown
```diff
@@ -67,6 +67,8 @@ The following arguments are supported:
   in a VCS push. If enabled, the working directory and trigger prefixes describe a set of 
   paths which must contain changes for a VCS push to trigger a run. If disabled, any push will 
   trigger a run. Defaults to `true`.
+* `global_remote_state` - (Optional) Whether to allow remote state access to
+  this workspace. It defaults to `true`.
 * `operations` - **Deprecated** Whether to use remote execution mode. When set to `false`, the workspace will 
   be used for state storage only. Defaults to `true`. This value _must not_ be provided if `execution_mode` is 
   provided.
```
chrisarcand, line 74: Needs the associated entry for `remote_state_consumer_ids`

tfe/resource_tfe_workspace_test.go
```diff
@@ -1326,3 +1360,25 @@ resource "tfe_workspace" "foobar" {
...
+
+resource "tfe_workspace" "foobar" {
+  name               = "workspace-test"
+  organization       = tfe_organization.foobar.id
+  allow_destroy_plan = false
+  auto_apply         = true
+	global_remote_state = false
+  remote_state_consumer_ids = [tfe_workspace.foobar_remote.id]
+}`, rInt)
+}
```
chrisarcand, line 1384: We should add a slightly more thorough test, perhaps moving from global sharing to two different configs of different state consumers (testing adding/removing them in the Update function).

tfe/data_source_workspace.go
```diff
@@ -44,6 +44,11 @@ func dataSourceTFEWorkspace() *schema.Resource {
 				Computed: true,
 			},
 
+			"global_remote_state": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
```
chrisarcand, line 51: Needs the associated `remote_state_consumer_ids` schema/setting.

website/docs/r/workspace.html.markdown
chrisarcand, line 74: Needs the associated datasource documentation.

chrisarcand (changes requested):
Lookin' good, just a few more bits to add and polish up.

tfe/resource_tfe_workspace_test.go
omarismail: My bad, i intended to add more tests and then didn't. Will do so

tfe/resource_tfe_workspace.go
omarismail: Ah, okay, makes sense!

tfe/data_source_workspace.go
omarismail: added

tfe/resource_tfe_workspace_test.go
omarismail: added a few more

go.mod
```diff
@@ -4,13 +4,12 @@ go 1.16
...
 	github.com/hashicorp/go-retryablehttp v0.6.8 // indirect
 	github.com/hashicorp/go-slug v0.7.0 // indirect
 	github.com/hashicorp/go-tfe v0.14.0
 	github.com/hashicorp/go-version v1.2.1
 	github.com/hashicorp/hcl v0.0.0-20180404174102-ef8a98b0bbce
-	github.com/hashicorp/terraform-plugin-sdk/v2 v2.3.0
+	github.com/hashicorp/terraform-plugin-sdk/v2 v2.6.0
```
chrisarcand, line 12: ? Did you intend for an SDK bump in these changes?

website/docs/d/workspace.html.markdown
```diff
@@ -34,6 +34,8 @@ In addition to all arguments above, the following attributes are exported:
 * `allow_destroy_plan` - Indicates whether destroy plans can be queued on the workspace.
 * `auto_apply` - Indicates whether to automatically apply changes when a Terraform plan is successful.
 * `file_triggers_enabled` - Indicates whether runs are triggered based on the changed files in a VCS push (if `true`) or always triggered on every push (if `false`).
+* `global_remote_state` - (Optional) Whether to allow remote state access to this workspace. It defaults to `true`.
```
chrisarcand, line 37: This seems to imply that `false` means remote state access isn't allowed, but it's just for implicit org-wide sharing.

I'd just use what's in the API docs, tweaked slightly for the provider:

```suggestion
* `global_remote_state` - (Optional) Whether the workspace should allow all workspaces in the organization to access its state data during runs. If false, then only specifically approved workspaces can access its state (determined by the `remote_state_consumer_ids` argument).
```

website/docs/d/workspace.html.markdown
```diff
@@ -34,6 +34,8 @@ In addition to all arguments above, the following attributes are exported:
...
+* `remote_state_consumer_ids` - (Optional) A set of workspace IDs that will be set as the remote state consumers for the given workspace.
```
chrisarcand, line 38:
```suggestion
* `remote_state_consumer_ids` - (Optional) A set of workspace IDs that will be set as the remote state consumers for the given workspace. Cannot be used if `global_remote_state` is set to `true`.
```

website/docs/r/workspace.html.markdown
```diff
@@ -67,6 +67,8 @@ The following arguments are supported:
   in a VCS push. If enabled, the working directory and trigger prefixes describe a set of 
   paths which must contain changes for a VCS push to trigger a run. If disabled, any push will 
   trigger a run. Defaults to `true`.
+* `global_remote_state` - (Optional) Whether to allow remote state access to this workspace. It defaults to `true`.
```
chrisarcand, line 70: The other arguments here don't follow the convention (🙄) but typically you see the same description as the associated resource argument, but tweaked slightly to indicate that it's read-only:

```suggestion
* `global_remote_state` - (Optional) Whether the workspace allows all workspaces in the organization to access its state data during runs. If false, then only specifically approved workspaces can access its state (`remote_state_consumer_ids`).
```

website/docs/r/workspace.html.markdown
```diff
@@ -67,6 +67,8 @@ The following arguments are supported:
...
+* `remote_state_consumer_ids` - (Optional) A set of workspace IDs that will be set as the remote state consumers for the given workspace.
```
chrisarcand, line 71:
```suggestion
* `remote_state_consumer_ids` - (Optional) The set of workspace IDs set as explicit remote state consumers for the given workspace.
```

chrisarcand (changes requested):
Getting closer! I'll pull down this build and give it a spin now while we wrestle with tests and polish up the last little bits.

go.mod
omarismail: Yup, its to get the acceptance tests to work: https://github.com/hashicorp/terraform-plugin-sdk/blob/main/CHANGELOG.md#243-february-10-2021

chrisarcand: Ah excellent!

chrisarcand (approved):

PR #182, opened 2020-05-27
Deprecate workspace_external_id/workspace_external_ids in tfe_notification_configuration, tfe_policy_set, and tfe_run_trigger resources

lafentres:
## Description

This PR adds deprecation warnings to the `ids` attribute for data source `tfe_workspace_ids`, the `workspace_external_id` attributes for resources `tfe_notification_configuration` & `tfe_run_trigger`, and the `workspace_external_ids` attribute for resource `tfe_policy_set`. 

#### The following data sources are impacted:
   * `tfe_workspace_ids` - `ids` is deprecated and will show a deprecation warning but will not cause any errors. You should use `full_names` instead.
**NOTE:** If you use the `tfe_workspace_ids` data source, this deprecation warning won't go away until it is removed. This is due to a limitation in Terraform SDK with deprecating attributes that aren't specified via config. [See this issue for more information](https://github.com/hashicorp/terraform/issues/7569).

#### The following resources are impacted:
   * `tfe_notification_configuration` - `workspace_external_id` is deprecated and will show a deprecation warning but will not cause any errors. You should use `workspace_id` instead
   * `tfe_policy_set` - `workspace_external_ids` is deprecated and will show a deprecation warning but will not cause any errors. You should use `workspace_ids` instead
   * `tfe_run_trigger` - `workspace_external_id` is deprecated and will show a deprecation warning but will not cause any errors. You should use `workspace_id` instead

### Why are these attributes being deprecated?
After #106 migrated workspace ids to the immutable id format (`ws-<random string>`), we still had some inconsistent usages of `external_id`, `workspace_external_id`, and `workspace_external_ids`. In order to clean this up and make the provider consistent and easier to understand when working with workspace IDs, we will be replacing all usages of `workspace_external_id`/`workspace_external_ids` with `workspace_id`/`workspace_ids`.

###  Deprecation plan:
To minimize disruption and give everyone time to migrate over, we will be doing this in 3 phases. 
#### Phase 1:
You are here. That's what this PR is for :] 

**tfe_workspace_ids Data Source:** We will add a new computed attribute to the tfe_workspace_ids data source called `full_names` that will mimic the current behavior of the `ids` attribute. In this phase, both attributes will return the old human-readable workspace ID format of `<organization_name>/<workspace_name>`. We will add deprecation warnings to the `ids` attribute explaining that users should begin using `full_names` instead, as the behavior of `ids` will be changing soon.

[...]

PR #106, opened 2019-11-20
Workspace id migration

beekus:
This PR migrates the schema of workspaces to use the workspace's immutable `external_id` as the **id**, instead of the mutable `organization/workspace` identifier.

The *variables* and *team_access* resources also have a `workspace_id` field, which is migrated to the new `workspace.id` as well. Additionally, the *team_access* and *workspace* data source have been updated accordingly, though those don't require a state migration.

### Todo:
- [x] Fix the variable and team_access (and maybe workspace) importers
- [x] Figure out how to unit test the variable and team access migrations, since they have to reach out to the API.
- [x] ~Adjust the state upgraders to not upgrade if the `workspace_id` is already the `external_id`.~ Also not going to do this as it seems extremely corner casey.
- [x] Add a validation rule in all of the places where workspace ids are expected to produce a helpful error message if we detect the user is trying to hard-code the "organization/workspace" syntax
- [x] Some additional tests probably need to be updated, as well as the acceptance tests.
- [x] Update the docs
- [ ] Update the CHANGELOG.

### Notes
* Requires at least TFE v201904-1

PR #295, opened 2021-04-26
Remove deprecated external_id/external_ids attributes

lafentres:
## Description

This PR removes the deprecated `external_id` attribute from resource `tfe_workspace`, the `external_id` attribute from data source `tfe_workspace`, and the `external_ids` attribute from data source `tfe_workspace_ids`.  
**This is a breaking change. Please change your config to use `id`/`ids` before upgrading.** 

#### The following data sources are impacted:
   * `tfe_workspace` - The deprecated `external_id` attribute has been removed. If you are still using this attribute, you will get errors. You should use `id` instead.
   * `tfe_workspace_ids` - The deprecated `external_ids` attribute has been removed. If you are still using this attribute, you will get errors. You should use `ids` instead.

#### The following resources are impacted:
   * `tfe_workspace` - The deprecated `external_id` attribute has been removed. If you are still using this attribute, you will get errors. You should use `id` instead

### Why are these attributes being deprecated?
After #106 migrated workspace ids to the immutable id format (`ws-<random string>`), we still had some inconsistent usages of `external_id`, `workspace_external_id`, and `workspace_external_ids`. In order to clean this up and make the provider consistent and easier to understand when working with workspace IDs, we will be replacing all usages of `workspace_external_id`/`workspace_external_ids` with `workspace_id`/`workspace_ids`.

###  Deprecation plan:
To minimize disruption and give everyone time to migrate over, we will be doing this in 3 phases. 

#### [COMPLETE] Phase 1:
**This phase was completed in #182 and released in version 0.18.0.**

<details>
   <summary>Click to expand for phase 1 details</summary>
   
   **tfe_workspace_ids Data Source:** We will add a new computed attribute to the tfe_workspace_ids data source called `full_names` that will mimic the current behavior of the `ids` attribute. In this phase, both attributes will return the old human-readable workspace ID format of `<organization_name>/<workspace_name>`. We will add deprecation warnings to the `ids` attribute explaining that users should begin using `full_names` instead, as the behavior of `ids` will be changing soon.

   **Remainder of TFE Provider:** We will add a new attribute to impacted resources called `workspace_id` or `workspace_ids` (depending on which is required for the specific resource) which will hold the same value as the current `workspace_external_id`/`workspace_external_ids` attributes. Deprecation warnings will be added to all of these, recommending the user move to `workspace_id`/ `workspace_ids` instead.
</details>

#### [COMPLETE] Phase 2:
**This phase was completed in #253 and released in version 0.24.0.**

<details>
   <summary>Click to expand for phase 2 details</summary>

   **tfe_workspace_ids Data Source:** We will officially change the behavior of the `ids` attribute so it returns the immutable workspace ID in format `ws-<random_string>`, just as the `external_ids` attribute currently does. We will add a deprecation warning to the `external_ids` attribute, alerting them that it will be deprecated soon and they should use `ids` instead.

   **tfe_workspace Data Source:** We will add a deprecation warning to the `external_id` attribute, recommending the user move to `id` instead.

   **tfe_workspace Resource:** We will add a deprecation warning to the `external_id` attribute, recommending the user move to `id` instead.

   **Remainder of TFE Provider:** We will remove the `workspace_external_id `/ `workspace_external_ids` attributes from all the impacted resources.
</details>

#### [YOU ARE HERE] Phase 3:
**You are here. That's what this PR is for :]**

**tfe_workspace_ids Data Source:** We will remove `external_ids` from this data source entirely, leaving behind `ids` returning the immutable workspace ID in the format `ws-<random_string>`, and `full_names` returning the format `<organization_name>/<workspace_name>`.

**tfe_workspace Data Source:** We will remove the `external_id` attribute.

**tfe_workspace Resource:** We will remove the `external_id` attribute.

## BREAKING CHANGE

This PR removes `external_id `/ `external_ids` from `r/tfe_workspace`, `d/tfe_workspace`, and `d/tfe_workspace_ids`. You should change your config to use `id `/ `ids` before upgrading. 

## Testing plan

As these changes just remove certain arguments, the testing plan is mostly just to have an existing 0.24.0 configuration/state and then 'upgrade' using a built provider with this branch:

* `external_id(s)` should not be usable anymore at all on the datasource or resource.
* No deprecation warnings should be emitted anymore from d/tfe_workspace.
* Perhaps most importantly: The behavior of what happens with the existing `external_id` attribute in state should be monitored - the attributes should disappear without incident, not causing any drift or awkward half-states.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## Unreleased 0.25.0
+
+BREAKING CHANGES:
+* d/tfe_workspace: Removed deprecated `external_id` attribute, preferring `id` instead ([#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295))
+* d/tfe_workspace_ids: Removed deprecated `external_ids` attribute, preferring `ids` instead ([#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295))
+* r/tfe_workspace: Removed deprecated `external_id` attribute, preferring `id` instead ([#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295))
+
+NOTES:
+* You will need to migrate to the preferred attributes to update to the latest version of this
+  provider. More information about these deprecations can be found in the description of [#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295)
+
 ## Unreleased 0.24.1
 
 ENHANCEMENTS:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,20 +1,21 @@
 ## Unreleased 0.25.0
 
 BREAKING CHANGES:
-* d/tfe_workspace: Removed deprecated `external_id` attribute, preferring `id` instead ([#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295))
-* d/tfe_workspace_ids: Removed deprecated `external_ids` attribute, preferring `ids` instead ([#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295))
-* r/tfe_workspace: Removed deprecated `external_id` attribute, preferring `id` instead ([#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295))
-
-NOTES:
-* You will need to migrate to the preferred attributes to update to the latest version of this
-  provider. More information about these deprecations can be found in the description of [#295](https://github.com/hashicorp/terraform-provider-tfe/pull/295)
-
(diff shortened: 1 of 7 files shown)


chrisarcand (approved):
Looks great! I did an 'upgrade' on an existing workspace and everything checks out: old arguments need to be updated (of course), no more deprecation warnings, and the previously tracked state just disappears happily without incident.

Thanks again @lafentres for seeing this migration all the way to the end! Extremely appreciated.

PR #297, opened 2021-04-29
Fix remote state consumer tests

chrisarcand:
## Description

In a unfortunate twist of fate and unbelievable timing, the attribute in state that this test function was depending on was just ripped out from underneath it in the deprecation removals of #295  😆 Use `id`, not `external_id`, which is now removed.

## Testing plan

CI should pass.


omarismail (approved):
image

PR #304, opened 2021-04-30
clean up CI checks

radditude:
## Description

We've had this Travis config hanging around for a while, but all it does is run the tests (which Circle already does) and run `make vet`. Let's say our farewells to Travis and move the vet check into Circle.


chrisarcand (approved):
Yayyy! Thank you!!

PR #320, opened 2021-05-25
Add Data Source: Organizations & Organization

omarismail:
## Description

This PR adds two data sources: `Organizations` and `Organization`. 

`Organizations` returns a list of names and a map of IDs/Names.

`Organization` returns all the fields for a given organization, and requires a 'name' to query it.

## Testing plan

1. Run the tests.
2. Create a terraform config and use the `data` source for `tfe_organization` and `tfe_organizations` and `output` the values.

```
$ TFE_TOKEN=<token> TFE_HOSTNAME=<hostname> TF_ACC=1  go test -v ./tfe/...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,12 +138,13 @@ Please use the template below when updating the changelog:
 * r/tfe_resource: description of change or bug fix ([#124](link-to-PR))
 ```
 
-### Updating the documentation.
+### Updating the documentation
 
-To update the documentation for the provider, please copy the new markdown and
-paste it here: https://registry.terraform.io/tools/doc-preview
+For pull requests that update provider documentation, please help us verify that the
+markdown will display correctly on the Registry:
 
-Then copy/paste an image of the updated documentation in your pull request.
+- Copy the new markdown and paste it here to preview: https://registry.terraform.io/tools/doc-preview
+- Paste a screenshot of that preview in your pull request.
 
 ### Change categories
 
--- a/README.md
+++ b/README.md
@@ -138,6 +138,13 @@ Please use the template below when updating the changelog:
 * r/tfe_resource: description of change or bug fix ([#124](link-to-PR))
 ```
 
+### Updating the documentation.
+
+To update the documentation for the provider, please copy the new markdown and
+paste it here: https://registry.terraform.io/tools/doc-preview
+
+Then copy/paste an image of the updated documentation in your pull request.
+
 ### Change categories
 
 - BREAKING CHANGES: Use this for any changes that aren't backwards compatible. Include details on how to handle these changes.
--- a/tfe/data_source_organization.go
+++ b/tfe/data_source_organization.go
@@ -72,7 +72,6 @@ func dataSourceTFEOrganizationRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("cost_estimation_enabled", org.CostEstimationEnabled)
 	d.Set("email", org.Email)
 	d.Set("owners_team_saml_role_id", org.OwnersTeamSAMLRoleID)
-	d.Set("saml_enalbed", org.SAMLEnabled)
 	d.Set("two_factor_conformant", org.TwoFactorConformant)
 
 	return nil
--- a/tfe/data_source_organization.go
+++ b/tfe/data_source_organization.go
@@ -43,11 +43,6 @@ func dataSourceTFEOrganization() *schema.Resource {
 				Computed: true,
 			},
 
-			"saml_enalbed": {
-				Type:     schema.TypeBool,
-				Computed: true,
-			},
-
 			"two_factor_conformant": {
 				Type:     schema.TypeBool,
 				Computed: true,
--- a/tfe/data_source_organization.go
+++ b/tfe/data_source_organization.go
@@ -145,19 +145,19 @@ func dataSourceTFEOrganizationRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("cost_estimation_enabled", org.CostEstimationEnabled)
 	d.Set("email", org.Email)
 	d.Set("enterprise_plan", org.EnterprisePlan)
-	var permissions []interface{}
-	permissions = append(permissions, map[string]interface{}{
-		"can_create_team":                org.Permissions.CanCreateTeam,
-		"can_create_workspace":           org.Permissions.CanCreateWorkspace,
-		"can_create_workspace_migration": org.Permissions.CanCreateWorkspaceMigration,
-		"can_destroy":                    org.Permissions.CanDestroy,
-		"can_traverse":                   org.Permissions.CanTraverse,
-		"can_update":                     org.Permissions.CanUpdate,
-		"can_update_api_token":           org.Permissions.CanUpdateAPIToken,
-		"can_update_oauth":               org.Permissions.CanUpdateOAuth,
-		"can_update_sentinel":            org.Permissions.CanUpdateSentinel,
+	d.Set("permissions", []interface{}{
+		map[string]interface{}{
+			"can_create_team":                org.Permissions.CanCreateTeam,
+			"can_create_workspace":           org.Permissions.CanCreateWorkspace,
+			"can_create_workspace_migration": org.Permissions.CanCreateWorkspaceMigration,
+			"can_destroy":                    org.Permissions.CanDestroy,
+			"can_traverse":                   org.Permissions.CanTraverse,
+			"can_update":                     org.Permissions.CanUpdate,
+			"can_update_api_token":           org.Permissions.CanUpdateAPIToken,
+			"can_update_oauth":               org.Permissions.CanUpdateOAuth,
+			"can_update_sentinel":            org.Permissions.CanUpdateSentinel,
+		},
 	})
-	d.Set("permissions", permissions)
 	d.Set("owners_team_saml_role_id", org.OwnersTeamSAMLRoleID)
 	d.Set("saml_enalbed", org.SAMLEnabled)
 	d.Set("session_remember", org.SessionRemember)
--- a/tfe/data_source_organization.go
+++ b/tfe/data_source_organization.go
@@ -128,13 +128,11 @@ func dataSourceTFEOrganizationRead(d *schema.ResourceData, meta interface{}) err
 	tfeClient := meta.(*tfe.Client)
 
 	name := d.Get("name").(string)
-
 	log.Printf("[DEBUG] Read configuration for Organization: %s", name)
-
 	org, err := tfeClient.Organizations.Read(ctx, name)
 	if err != nil {
 		if err == tfe.ErrResourceNotFound {
-			return fmt.Errorf("Could not read organization")
+			return fmt.Errorf("Could not read organization '%s'", name)
 		}
 		return fmt.Errorf("Error retrieving organization: %v", err)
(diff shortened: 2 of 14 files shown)


tfe/data_source_organization.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		if err == tfe.ErrResourceNotFound {
+			return fmt.Errorf("Could not read organization '%s'", name)
+		}
+		return fmt.Errorf("Error retrieving organization: %v", err)
+	}
+
+	log.Printf("[DEBUG] Setting Organization Attributes")
+	d.Set("name", org.Name)
+	d.SetId(org.Name)
```
omarismail, line 142: I used the organization name as the main identifier, even though there is an external_id, because we currently use the `name` as the main identifier.

README.md
```diff
@@ -138,6 +138,13 @@ Please use the template below when updating the changelog:
 * r/tfe_resource: description of change or bug fix ([#124](link-to-PR))
 ```
 
+### Updating the documentation.
+
+To update the documentation for the provider, please copy the new markdown and
+paste it here: https://registry.terraform.io/tools/doc-preview
+
+Then copy/paste an image of the updated documentation in your pull request.
```
nfagerlund, line 146: This reads like it's instructions for updating the docs, but it's not, quite; it's just an extra thing we'd like people to do if their PR involves docs. How about this?

```suggestion
### Updating the documentation

For pull requests that update provider documentation, please help us verify that the
markdown will display correctly on the Registry:

- Copy the new markdown and paste it here to preview: https://registry.terraform.io/tools/doc-preview
- Paste a screenshot of that preview in your pull request.
```

nfagerlund (commented):
I hope to come back and test this later if I have time, but in the meantime I'll just post the readme suggestion I had.

tfe/data_source_organization.go
```diff
@@ -0,0 +1,168 @@
...
+
+			"email": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"enterprise_plan": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
chrisarcand, line 44: 🤔  I don't actually know what this is and don't see it in the API.

tfe/data_source_organization.go
```diff
@@ -0,0 +1,168 @@
...
+							Type:     schema.TypeBool,
+							Computed: true,
+						},
+
+						"can_update_sentinel": {
+							Type:     schema.TypeBool,
+							Computed: true,
+						},
+					},
+				},
```
chrisarcand, line 101: `permissions` objects aren't actually client-supported API; they aren't inherent to the object itself but the relationship of the authenticated user to that object. Their purpose is solely to provide logic for Terraform Cloud's user interface, and shouldn't be present here in the provider.

tfe/data_source_organization.go
```diff
@@ -0,0 +1,168 @@
...
+			},
+
+			"saml_enalbed": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
```
chrisarcand, line 107: This attribute is essentially deprecated. See the [internal!] serialization logic for more on what this is and why.

tfe/data_source_organization.go
```diff
@@ -0,0 +1,168 @@
...
+
+			"session_remember": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"session_timeout": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
```
chrisarcand, line 117: The two `session_` fields here, much like `permissions`, aren't valid here in the provider. There is no session!

tfe/data_source_organizations.go
```diff
@@ -0,0 +1,57 @@
...
+func dataSourceTFEOrganizations() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTFEOrganizationList,
+
+		Schema: map[string]*schema.Schema{
+			"names": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
```
chrisarcand, line 20: I wonder if this is necessary? For workspaces, `names` is a required argument you must provide; here in organizations, you're just fetching all organizations accessible by the authenticated token. 

Since `ids` (which matches the equivalent workspaces one) includes both names and IDs, this just feels like duplication, no?

tfe/data_source_organizations.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, org := range orgs.Items {
+		resourceID += org.Name
+		ids[org.Name] = org.ExternalID
+		names = append(names, org.Name)
+	}
+
+	log.Printf("[DEBUG] Setting Organizations Attributes")
+	d.Set("names", names)
+	d.Set("ids", ids)
+	d.SetId(fmt.Sprintf("%d", schema.HashString(resourceID)))
```
chrisarcand, line 54: You don't need to use `HashString` here; because the input is always `""`, the value is always `0`. For that matter you may just want to use a more meaningful sentinel value, like `organizations`

website/docs/d/organization.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+* `name` - (Required) Name of the organization.
+
+## Attributes Reference
+
+The following attributes are exported:
```
chrisarcand, line 28:
```suggestion
In addition to all arguments above, the following attributes are exported:
```

(Matches ~all~ all but one of the existing data sources)

chrisarcand (changes requested):

tfe/data_source_organization.go
omarismail: I took this from the [go-tfe Organization struct](https://github.com/hashicorp/go-tfe/blob/master/organization.go#L83), but I should've checked the API as well. You're right, it isn't in the API.

Will remove this. Also, here is a PR in go-tfe to remove the unused attribute and add the ones that aren't there.

https://github.com/hashicorp/go-tfe/pull/225

tfe/data_source_organization.go
omarismail: 👍

tfe/data_source_organization.go
omarismail: Right, I was a big gung-ho about adding all the attributes 😅

tfe/data_source_organizations.go
omarismail: I put `names` here as well because we use the organization's `name` as a primary identifier throughout the API, and I figured a common usage pattern is that users would query this data source and iterate through the names. That being said, one could also just iterate through `ids` (IDs/Names map) and do an action using the `name`.

tfe/data_source_organization.go
chrisarcand: To your credit, it looks like we have Session stuff in the client from long, long ago that definitely shouldn't be there 😬

tfe/data_source_organizations.go
radditude: Hmmm, I was actually wondering the opposite! We use `id` [elsewhere in the provider](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/organization#id) to mean the organization's name, so it seems like it could be confusing to use `id` to mean the organization's external id 😬 plus you can't really use an organization's external id for anything right now, though of course that may change in the future. 

Possibly terrible idea: what if we left `ids` out of this data source completely and had `names` as the only attribute?

chrisarcand: > That being said, one could also just iterate through ids (IDs/Names map) and do an action using the name.

Yeah that was my thought

The fact that we don't ever use external IDs for orgs is problematic, and I hope we make some enhancements to allow for either in the future and actually change things like the reference CJ linked, but I can also appreciate that it might be overoptimizing for now. I think `names` as the only attribute is just fine, actually! Maybe we could do that and leave the ID map for when (because we probably will) we migrate `r/tfe_organization.id` over 🤷‍♂️

omarismail: > The fact that we don't ever use external IDs for orgs is problematic

I'm in favor keeping `names`. The user is used to using the organization name to execute operations, so now seeing "ids" _only_ (and not `name`) will be confusing. But like CJ pointed, we set `id` [elsewhere](https://github.com/hashicorp/terraform-provider-tfe/blob/master/tfe/resource_tfe_organization.go#L92) as [name](https://github.com/hashicorp/terraform-provider-tfe/blob/master/tfe/resource_tfe_organization.go#L172) where its not actually the ID (external_id) but in fact is a name...😢.

I'm actually in favor of keeping both. `names` because the user is already used to using that as the primary "id". And keeping the `ids` map, even though it may not be immediately used, is good because we can put the pieces in place for a future migration to start using external_id as the primary identifier instead of names.

chrisarcand: Good discussion, sounds good to me!

website/docs/d/organization.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+* `email` - Admin email address.
+* `external_id` - An identifier for the organization.
+* `collaborator_auth_policy` - Authentication policy (`password`
+  or `two_factor_mandatory`). Defaults to `password`.
+* `cost_estimation_enabled` - Whether or not the cost estimation feature is enabled for all workspaces in the organization. Defaults to true. In a Terraform Cloud organization which does not have Teams & Governance features, this value is always false and cannot be changed. In Terraform Enterprise, Cost Estimation must also be enabled in Site Administration.
+* `enterprise_plan` - Represents the Enterprise Plan type.
+* `owners_team_saml_role_id` - The name of the "owners" team.
+* `permissions` - Represents the organization permissions
+* `session_timeout_minutes` - Session timeout after inactivity. Defaults to `20160`.
+* `session_remember_minutes` - Session expiration. Defaults to `20160`.
```
chrisarcand, line 40: This all needs updating yet!

website/docs/d/organizations.html.markdown
```diff
@@ -0,0 +1,29 @@
...
+## Example Usage
+
+```hcl
+data "tfe_organizations" "foo" {
+}
+```
+
+## Argument Reference
+
+No arguments are required. This retrieves the names and IDs of all the organizations.
```
chrisarcand, line 22: Maybe hint here that it may not be _all_ the organizations.

```suggestion
No arguments are required. This retrieves the names and IDs of all the organizations readable by the provided token.
```

chrisarcand (approved):
Two bits to tweak and then this is ready to roll! Also needs a CHANGELOG entry.

dennisroche:
@omarismail can a new release of `tfe` be cut to include this change?

Last release was `0.25.3` on 19 May 2021.

chrisarcand:
@dennisroche Yes indeed! There's some work being done to support the new [workspace tagging](https://www.terraform.io/docs/cloud/api/workspaces.html#get-tags) feature; once that's in, we'll go ahead and cut a 0.26.0. Thanks for your patience!

dennisroche:
Thanks @chrisarcand. We're eagerly waiting for both this and the workspace tagging.

dennisroche:
0.26.0 was released. thank-you. 🚀

PR #323, opened 2021-06-03
Add Admin Organizations to data source organizations

### LINKED ISSUE #272: data source to list all organizations in tfe
Listing organizations is supported by API

https://www.terraform.io/docs/cloud/api/admin/organizations.html#list-all-organizations

There should be a data source to get all organizations. I currently have to use a shell script custom resources to get this

### LINKED ISSUE #281: tfe_organization data source missing
there doesn't appear to be a `tfe_organization` data source. 

for example i'd like to use the example given on the `tfe_agent_token` [docs](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/agent_token) but for an existing organization like so:
```
data "tfe_organization" "test-organization" {
  name  = "my-org-name"
}

resource "tfe_agent_pool" "test-agent-pool" {
  organization = data.tfe_organization.test-organization.id
}

resource "tfe_agent_token" "test-agent-token" {
  agent_pool_id = tfe_agent_pool.test-agent-pool.id
  description   = "my-agent-token-name"
}
```

omarismail:
## Description

In #320, two a data source for listing all organizations was added. That data source returns the organizations that the user (token) has access to. 

This PR augments that data source. We add a boolean field `admin` that determines if we should retrieve the organizations from using the [Admin Organizations endpoint](https://www.terraform.io/docs/cloud/api/admin/organizations.html#list-all-organizations) or the [Organizationsendpoint](https://www.terraform.io/docs/cloud/api/organizations.html#list-organizations) will be used.

Added docs.

## Testing plan

This can only be properly tested with access to Site Admin on your Terraform Enterprise instance. 

The testing plan is as follows:

Create a terraform config using the `tfe` provider, and add this bit:
```
data "tfe_organizations" "orgs" {
  admin = true 
}

output "many_orgs" {
  value = data.tfe_organizations.orgs
}
```

1. Using your API token (assuming you have Site Admin access), add  `admin = true` to the data source in the config, run `terraform apply` and you will see the output of all the organizations.
2. Create a new organization, and create an Organization token from this new organization.
3. Remove the `admin=true` from the data source, and run `terraform apply` using this organization token.
4. Confirm that you only see one organization, and not all of the organizations as in step 1.

## External links

- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/320)

Closes #281
Closes #272

## Followup

We will add the ability to query specific organizations as a follow up to this task.

### DIFF
--- a/tfe/data_source_organizations.go
+++ b/tfe/data_source_organizations.go
@@ -26,7 +26,8 @@ func dataSourceTFEOrganizations() *schema.Resource {
 
 			"admin": {
 				Type:     schema.TypeBool,
-				Optional: false,
+				Default:  false,
+				Optional: true,
 			},
 		},
 	}
--- a/tfe/data_source_organizations.go
+++ b/tfe/data_source_organizations.go
@@ -30,20 +30,28 @@ func dataSourceTFEOrganizations() *schema.Resource {
 func dataSourceTFEOrganizationList(d *schema.ResourceData, meta interface{}) error {
 	tfeClient := meta.(*tfe.Client)
 
-	log.Printf("[DEBUG] Listing all organizations")
-	orgs, err := tfeClient.Organizations.List(ctx, tfe.OrganizationListOptions{})
-	if err != nil {
-		if err == tfe.ErrResourceNotFound {
-			return fmt.Errorf("Could not list organizations.")
-		}
-		return fmt.Errorf("Error retrieving organizations: %v.", err)
-	}
-
 	names := []string{}
 	ids := map[string]string{}
-	for _, org := range orgs.Items {
-		ids[org.Name] = org.ExternalID
-		names = append(names, org.Name)
+
+	adminOrgs, err := getAdminOrgs(tfeClient)
+	if err != nil {
+		return err
+	}
+
+	if adminOrgs != nil {
+		for _, org := range adminOrgs.Items {
+			ids[org.Name] = org.ExternalID
+			names = append(names, org.Name)
+		}
+	} else {
+		orgs, err := getOrgs(tfeClient)
+		if err != nil {
+			return err
+		}
+		for _, org := range orgs.Items {
+			ids[org.Name] = org.ExternalID
+			names = append(names, org.Name)
+		}
 	}
 
 	log.Printf("[DEBUG] Setting Organizations Attributes")
@@ -52,4 +60,30 @@ func dataSourceTFEOrganizationList(d *schema.ResourceData, meta interface{}) err
 	d.Set("ids", ids)
 
 	return nil
+}
+
+func getAdminOrgs(client *tfe.Client) (*tfe.AdminOrganizationList, error) {
+	log.Printf("[DEBUG] Listing all organizations (admin)")
+	orgs, err := client.Admin.Organizations.List(ctx, tfe.AdminOrganizationListOptions{})
+	if err != nil {
+		if err == tfe.ErrResourceNotFound {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("Error retrieving organizations: %v.", err)
+	}
+
+	return orgs, nil
+}
+
+func getOrgs(client *tfe.Client) (*tfe.OrganizationList, error) {
+	log.Printf("[DEBUG] Listing all organizations (non-admin)")
+	orgs, err := client.Organizations.List(ctx, tfe.OrganizationListOptions{})
+	if err != nil {
+		if err == tfe.ErrResourceNotFound {
+			return nil, fmt.Errorf("Could not list organizations.")
+		}
+		return nil, fmt.Errorf("Error retrieving organizations: %v.", err)
+	}
+
+	return orgs, nil
 }
--- a/tfe/data_source_organizations.go
+++ b/tfe/data_source_organizations.go
@@ -26,7 +26,7 @@ func dataSourceTFEOrganizations() *schema.Resource {
 
 			"admin": {
 				Type:     schema.TypeBool,
-				Computed: true,
+				Optional: true,
 			},
 		},
 	}
--- a/tfe/data_source_organizations.go
+++ b/tfe/data_source_organizations.go
@@ -94,7 +94,11 @@ func orgsPopulateFields(client *tfe.Client) ([]string, map[string]string, error)
 	names := []string{}
 	ids := map[string]string{}
 	log.Printf("[DEBUG] Listing all organizations (non-admin)")
-	options := tfe.OrganizationListOptions{}
+	options := tfe.OrganizationListOptions{
+		ListOptions: tfe.ListOptions{
+			PageSize: 100,
+		},
+	}
 	for {
 		orgList, err := client.Organizations.List(ctx, options)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


tfe/data_source_organizations.go
```diff
@@ -53,3 +61,29 @@ func dataSourceTFEOrganizationList(d *schema.ResourceData, meta interface{}) err
 
 	return nil
 }
+
+func getAdminOrgs(client *tfe.Client) (*tfe.AdminOrganizationList, error) {
+	log.Printf("[DEBUG] Listing all organizations (admin)")
+	orgs, err := client.Admin.Organizations.List(ctx, tfe.AdminOrganizationListOptions{})
+	if err != nil {
+		if err == tfe.ErrResourceNotFound {
+			return nil, nil
```
omarismail, line 70: Returning `nil, nil` here on `ErrResourceNotFound` is intentional because of the way the code is sequenced above. We want to return an error in `dataSourceTFEOrganizationList` if there is an actual error. If we get `ErrResourceNotFound`, then we want to tell the calling function that there aren't any "admin organizations" and there isn't an error, so that way we continue on to retrieve organizations.

chrisarcand (changes requested):
Lookin' nice! Two main points:

❌ This was a problem in the first version of this (Before admin stuff), so I'm glad we found this this time around! The result doesn't follow pagination. That is, even if you have more than 20 organizations available to you, the default page size of 20 is all that shows up. You'll need to add pagination concerns here; see [how workspaces does it as an example.](https://github.com/hashicorp/terraform-provider-tfe/blob/f1b9d6d4a2f67605cafcc6d438254cf561ad8048/tfe/data_source_workspace_ids.go#L71-L77)

❌ Needs documentation

---

Aside from that, the 'uses admin unless you aren't an admin' thing works but we may want to consider the alternative - an optional argument could be utilized, where you explicitly opt-in to using the admin API for installation-wide names/IDs. At it's simplest: `installation_wide = true` or `admin = true` (TFE only).  Besides being clear from a usage perspective, that would avoid the needless request out to the Admin API that will happen _every time_ for non-admin users; thoughts?

omarismail:
@chrisarcand 
> Aside from that, the 'uses admin unless you aren't an admin' thing works but we may want to consider the alternative - an optional argument could be utilized, where you explicitly opt-in to using the admin API for installation-wide names/IDs. At it's simplest: `installation_wide = true` or `admin = true` (TFE only). Besides being clear from a usage perspective, that would avoid the needless request out to the Admin API that will happen _every time_ for non-admin users; thoughts?

I think this is a good idea. I prefer `admin = true`. I also failed to notice that, while we don't have any query paramters for [OrganizationListOptions](https://github.com/hashicorp/go-tfe/blob/12b91a1666b30d004d170c61ee83a4ac38553cd1/organization.go#L124-L126), we do have them for [AdminOrganizationListOptions](https://github.com/hashicorp/go-tfe/blob/main/admin_organization.go#L68-L78).

Having one data source for both regular Organizations and Admin Organizations, having the prefix `admin_` would clarify. 

```
data "tfe_organizations" "orgs" {
  admin = true
  admin_query = <query>
}
```

One could also say that "if `admin=true` is set, and we see a `query` field, then we can assume that the user is intending the do add the query for admin orgs." But I think someone might just remove the `admin=true` field and keep the query and think it'll work. I think the verbosity here is clearer. 

Also, we can be a bit intelligent — while not making things confusing — by saying that "if `admin_query` is present but `admin=true` is not, then we can assume they are querying for admin orgs".

website/docs/d/organizations.html.markdown
```diff
@@ -27,3 +27,8 @@ The following attributes are exported:
 
 * `names` - A list of names of every organization.
 * `ids` - A map of organization names and their IDs.
+* `admin` - A boolean field that determines  the list of organizations that should
+  be retrieved. If it is true, then the [Admin Organizations
+endpoint](https://www.terraform.io/docs/cloud/api/admin/organizations.html#list-all-organizations)
+  will be used, otherwise if it is false or not included, then the [Organizations
+endpoint](https://www.terraform.io/docs/cloud/api/organizations.html#list-organizations) will be used.
```
omarismail, line 34: image

omarismail, line 34: image

tfe/data_source_organizations.go
```diff
@@ -23,27 +23,30 @@ func dataSourceTFEOrganizations() *schema.Resource {
 				Type:     schema.TypeMap,
 				Computed: true,
 			},
+
+			"admin": {
+				Type:     schema.TypeBool,
+				Optional: true,
```
chrisarcand, line 29: `Default: false` ?

website/docs/d/organizations.html.markdown
chrisarcand: * This should be an argument (listed above) rather than an attribute reference here. It's a user-set thing, not a piece of data retrieved. 

* I'd probably explain the difference in usage really; the user cares more about when you'd use this or not rather than which endpoint it uses (which is obvious to us); you could just add this info to what you already have, if you think the endpoint matters. You could also indicate that this is really Terraform Enterprise only, practically speaking.

> Terraform Enterprise only. If true, then the Admin Organizations endpoint will be used and the list will contain all the organization names across a Terraform Enterprise installation. If false, the Organizations endpoint will be used and all organizations readable by the token provided will be listed. Defaults to false.

tfe/data_source_organizations.go
```diff
@@ -53,3 +56,63 @@ func dataSourceTFEOrganizationList(d *schema.ResourceData, meta interface{}) err
...
 	return nil
 }
+
+func adminOrgsPopulateFields(client *tfe.Client, d *schema.ResourceData) ([]string, map[string]string, error) {
+	names := []string{}
+	ids := map[string]string{}
+	log.Printf("[DEBUG] Listing all organizations (admin)")
+	options := tfe.AdminOrganizationListOptions{}
+	for {
+		orgList, err := client.Admin.Organizations.List(ctx, options)
```
chrisarcand, line 66: This is problematic at any larger scale, sadly. I hooked this up to an installation with several thousand organizations and it took so long I canceled it after 15 minutes.

Short of a specialized endpoint for making this really performant, I think the most practical thing we can do here is to use sparse fieldsets to minimize serialization and max out the page size, which is currently twenty 😱. That is, you'll need to make this URL what is used here (and after that, might as well add it to the other endpoint, too):

```
/api/v2/admin/organizations?fields[organization]=name&page[size]=100
```

This also tells us that query filtering is undoubtedly going to be the next step here to limit results (similar to `d/tfe_workspace.names`, but smarter), but we can save that for another day.

chrisarcand (changes requested):
Nice, moving along! Biggie problem with the admin side of things sadly; with any larger amount of organizations this datasource takes a reeeeaaallly long time. That's probably already the case with `d/tfe_workspace` if you query for all workspaces in an org, but without some filtering provided here we should at least squeeze as much as we can out of minimizing the request load.

website/docs/d/organizations.html.markdown
omarismail: Updated
image

tfe/data_source_organizations.go
omarismail: I had thought it would normally default to false? I added it explicitly

chrisarcand (approved):
Works great! This datasource worked very reasonably with ~700 organizations on the TFE instance I tried it against. 🚀

PR #335, opened 2021-07-28
Add Workspace Outputs data source

omarismail:
## Description

This PR introduces a new data source for handling state outputs. `tfe_outputs ` takes an organization and workspace, and retrieves the state outputs for that workspace. 

**Example**

```
data "tfe_outputs" "foobar" {
  organization = "<org-name>"
  workspace = "<workspace-name>"
}

output "states" {
	value = data.tfe_outputs.foobar.values.<key>
}
```

The `<key>` in the `states` output can be any field in your state outputs. So if your state output looks something like this:
```
  "outputs": {
    "user": {
      "value": "9023256633839603543",
      "type": "string"
    },
    "records": {
      "value": ["hashicorp.com","terraform.io"],
      "type": ["list","string"]
    },
    "secret": {
      "value": "token",
      "type": "string",
      "sensitive": true,
    }
  }
```

Then the `<key>` can be as follows:
```
data "tfe_outputs" "foobar" {
  organization = "<org-name>"
  workspace = "<workspace-name>"
}

output "user_output" {
	value = data.tfe_outputs.foobar.values.user
}
output "records_output" {
	value = data.tfe_outputs.foobar.values.records
}
```

To show sensitive values, the user can set the boolean flag `sensitive=true`:

```
data "tfe_outputs" "foobar" {
  organization = "<org-name>"
  workspace = "<workspace-name>"
  show_sensitive = true
}

output "secret" {
	value = data. tfe_outputs.foobar.values.secret
}
```

## Testing plan

1. Pull Down the branch
2. Create a terraform configuration file that includes the tfe-provider
3. Copy the example code above and make sure the `tfe_outputs ` data source has a organization and workspace that has state outputs. If the workspace doesn't, then the `values` field of the `tfe_outputs ` will be empty.
4. Run terraform and you will see the outputs from that statefile. 
5. Change your terraform config to select a specific value from the output. Re-run terraform and you should see the value printed.

## External links

An RFC was passed internally. 

```
$ TFE_TOKEN=<token> TFE_HOSTNAME=app.terraform.io TF_ACC=1 TF_LOG_PROVIDER=DEBUG  go test ./tfe/... -v -run TestAccTFEStateOutputs
=== RUN   TestAccTFEOutputs
--- PASS: TestAccTFEOutputs (12.08s)
--- PASS: TestAccTFEOutputs_emptyOutputs (12.12s)
PASS
```


website/docs/d/state_outputs.html.markdown
```diff
@@ -0,0 +1,71 @@
+---
```
omarismail, line 1: image

tfe/data_source_state_outputs.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil, fmt.Errorf("Error downloading state: %v", err)
+	}
+
+	log.Printf("[DEBUG] Unmarshalling state output")
+	read := bytes.NewReader(rawState)
+	src, err := ioutil.ReadAll(read)
+	if err != nil {
+		return nil, fmt.Errorf("Could not read state data: %v", err)
+	}
+	rrs := &rootModule{}
```
omarismail, line 203: All the unmarshalling for the rest of this function will eventually be put in the go-client as part of the `StateVersions` interface.

main.go
```diff
@@ -1,11 +1,36 @@
...
+
 func main() {
-	plugin.Serve(&plugin.ServeOpts{
-		ProviderFunc: tfe.Provider})
+	ctx := context.Background()
+	mux, err := tfmux.NewSchemaServerFactory(
+		ctx, tfe.Provider().GRPCProvider, tfe.PluginProviderServer,
+	)
+	if err != nil {
+		log.Println(fmt.Errorf("Could not setup a NewSchemaServerFactory using the providers: %v", err))
```
paddycarver, line 25: 1) I think you're going to want to prefix this log line with `[ERROR]`, so Terraform registers it as an error log.
2) I think you could use `log.Printf` here.

tfe/data_source_state_outputs.go
```diff
@@ -0,0 +1,260 @@
...
+	var orgName string
+	var wsName string
+	var err error
+
+	config := req.Config
+	val, err := config.Unmarshal(tftypes.Object{
+		AttributeTypes: map[string]tftypes.Type{
+			"workspace":    tftypes.String,
+			"organization": tftypes.String,
+			"values":       tftypes.String,
```
paddycarver, line 133: Should this be a `tftypes.String`?

tfe/data_source_state_outputs.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		tfRawState := tfprotov5.RawState{
+			JSON: mByte,
+		}
+		newVal, err := tfRawState.Unmarshal(tfType)
+		if err != nil {
+			return nil, nil, fmt.Errorf("Could not unmarshal tftype into value: %v", err)
+		}
+		tftypesValues[name] = newVal
+		stateTypes[name] = tfType
```
paddycarver, line 256: This is very clever, and something I hadn't thought of. It makes me slightly nervous, but... well, the JSON interface of types and values _is_ something that cty and tftypes need to agree on, and could be considered their compatibility interface with each other.

Just know that you're on shaky ground with the JSON parsing of the values because while they're _supposed_ to be compatible, that's not as proven as the msgpack encoding. We really only see values as JSON during state upgrades, I think, which is a less-explored part of the protocol than PlanResourceChange/ApplyResourceChange, both of which use msgpack to encode values.

Not saying what you're doing here is wrong, or that you should change it, just know you may be the first to find an edge case if one (somehow) exists. We have pretty extensive unit testing, but....

tfe/plugin_provider.go
```diff
@@ -0,0 +1,252 @@
...
+
+	err = valMap["token"].As(&token)
+	if err != nil {
+		return meta, fmt.Errorf("Could not set the token value to string %v", err)
+	}
+	if token == "" && os.Getenv("TFE_TOKEN") == "" {
+		return meta, fmt.Errorf("Token must not be empty")
+	} else if token == "" && os.Getenv("TFE_TOKEN") != "" {
+		token = os.Getenv("TFE_TOKEN")
+	}
```
paddycarver, line 246: You should probably check these aren't null or unknown before trying to call `.As` on the `valMap` values, because they could be and I believe you'd return an error at the moment, and not a particularly helpful one at that.

paddycarver (commented):
Some minor comments, but the overall design seems okay to me.

tfe/data_source_state_outputs.go
omarismail: Nope! Will change, nice catch

brandonc (commented):
I'll circle back later after I learn how to test this change locally. For now, a couple of questions and an insignificant spelling issue

main.go
```diff
@@ -1,11 +1,35 @@
...
 
+const (
+	tfeProviderName = "registry.terraform.io/hashicorp/tfe"
+)
+
 func main() {
-	plugin.Serve(&plugin.ServeOpts{
-		ProviderFunc: tfe.Provider})
+	ctx := context.Background()
+	mux, err := tfmux.NewSchemaServerFactory(
```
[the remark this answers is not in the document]
chrisarcand: In a nutshell, outputs can be of several different types and the current SDK does not allow for that. Digging deeper into the underlying plugin package, you can make this happen with [the DynamicPseudoType primitive](https://github.com/hashicorp/terraform-plugin-go/blob/26c9b682270fc6618969eccbd21789a48c1d8251/tftypes/primitive.go#L9-L12), which requires you to mux it back in to the rest of everything built with the SDK in to this single provider. (https://github.com/hashicorp/terraform-plugin-mux)

go.mod
```diff
@@ -3,13 +3,35 @@ module github.com/hashicorp/terraform-provider-tfe
...
 	github.com/hashicorp/go-tfe v0.17.0
 	github.com/hashicorp/go-version v1.3.0
 	github.com/hashicorp/hcl v0.0.0-20180404174102-ef8a98b0bbce
-	github.com/hashicorp/terraform-plugin-sdk/v2 v2.7.0
-	github.com/hashicorp/terraform-svchost v0.0.0-20191011084731-65d371908596
+	github.com/hashicorp/hcl/v2 v2.10.0 // indirect
+	github.com/hashicorp/terraform-exec v0.14.0 // indirect
+	github.com/hashicorp/terraform-plugin-go v0.2.1
+	github.com/hashicorp/terraform-plugin-mux v0.1.1
+	github.com/hashicorp/terraform-plugin-sdk/v2 v2.5.0
```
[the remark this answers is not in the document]
chrisarcand: 🤔 I was wondering the same; I thought for a moment it could be that the explicit import was removed from main.go and the plugin/mux packages use this version as a transitive dependency but...I don't think so?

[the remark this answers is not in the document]
omarismail: Good catches! Not sure how this happened, but updating this.

tfe/data_source_workspace_outputs.go
```diff
@@ -0,0 +1,289 @@
...
+	return dataSourceWorkspaceOutputs{
+		provider: p,
+	}
+}
+
+func (d dataSourceWorkspaceOutputs) ReadDataSource(ctx context.Context, req *tfprotov5.ReadDataSourceRequest) (*tfprotov5.ReadDataSourceResponse, error) {
+	resp := &tfprotov5.ReadDataSourceResponse{
+		Diagnostics: []*tfprotov5.Diagnostic{},
+	}
+	client, err := getClient(d.provider.meta.hostname, d.provider.meta.token, false)
```
chrisarcand, line 43: I sort of wonder if there's a way to re-use the same client from the main provider instance rather than setting up a separate client for this half?

I dislike the separate clients, it seems like it would remove quite a bit of redundancy in reparsing those provider attributes. Not critical if we can't do it, but do you see a way, somehow passing it to ` tfe.PluginProviderServer` maybe?

website/docs/d/workspace_outputs.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+page_title: "Terraform Enterprise: tfe_workspace_outputs"
+sidebar_current: "docs-datasource-tfe-state-outputs"
+description: |-
+  Retrieves the State outputs per organization and workspace.
+---
+# Data Source: tfe_workspace_outputs
+
+This data source is used to retrieve the state outputs for a given workspace.
+It enables values in the outputs to be used dynamically in a terraform
+configuration.
```
chrisarcand, line 12:
```suggestion
It enables output values in one Terraform configuration to be used in another.
```

website/docs/d/workspace_outputs.html.markdown
```diff
@@ -0,0 +1,93 @@
+---
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_workspace_outputs"
+sidebar_current: "docs-datasource-tfe-state-outputs"
+description: |-
+  Retrieves the State outputs per organization and workspace.
```
chrisarcand, line 6:
```suggestion
  Get output values from another organization/workspace.
```

(Matches pattern of other page descriptions)

website/docs/d/workspace_outputs.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+      "value": "token",
+      "type": "string",
+      "sensitive": true
+    }
+  },
+  "resources": [
+    ...
+  ]
+}
+```
```
chrisarcand, line 44: Generally speaking, the state file is an internal implementation detail who's format isn't guaranteed and has defined interfaces like `terraform state show` and outputs themselves to interact with it - in other words, I wouldn't actually give the content here as an example. It's prone to being outdated, and isn't the thing that people should care about when crafting their configurations. 

Instead, I would show the user-level interface used - configurations!

Example (yes, you can have a Terraform configuration with just hardcoded outputs 😆):

---

Given the outputs defined in a configuration used at `my-org/my-workspace`:

```
output "foo" {
  value = "a"
}

output "bar" {
  value = "b"
}
```

These outputs can be used in another:

```
data "tfe_workspace_outputs" "external_outputs" {
  organization = "my-org"
  workspace = "my-workspace"
}

<rest of example omitted but you get my gist>
```

website/docs/d/workspace_outputs.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+
+The `tfe_workspace_outputs` data source can now use `identifier` and `records`
+dynamically as seen below.
```
chrisarcand, line 47: * `can now` is temporal; this won't be new forever! 
* `dynamically` isn't of any consequence to the user, the fact that the type is variable is what's expected and sadly not the case until now 😆

```suggestion
Using the `tfe_workspace_outputs` data source, the outputs `identifier` and `records` can be used as seen below:
```

website/docs/d/workspace_outputs.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+output "secret" {
+	value = data.tfe_workspace_outputs.foobar.values.secret
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `organization` - (Required) The name of the organizatin.
```
chrisarcand, line 83:
```suggestion
* `organization` - (Required) The name of the organization.
```

website/docs/d/workspace_outputs.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+* `workspace` - (Required) The name of the workspace.
+* `sensitive` - (Optional) Determines whether or not to show sensitive values.
+  Set to `true` to reveal sensitive values.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `values` - A dynamic value that can call any state outputs key and retrieve
+  its value.
```
chrisarcand, line 93:
```suggestion
* `values` - The current output values for the specified workspace.
```

chrisarcand (changes requested):
Lookin' good! I'll give it a QA roll after some discussion. My biggest overarching bit to discuss is the `sensitive` argument.

As it stands, `sensitive = true` to _reveal_ the sensitive value is the complete opposite of `sensitive = true` when _defining_ the output value in the other configuration - so I see a lot of confusion ensuing! Given the current mechanism, this will need to clarified to be `show_sensitive` or something...

...but beyond that, I wonder why `sensitive` needs to exist at all. Could you elaborate on that a bit?

chrisarcand:
Random extra thought:

There's long been a fun disconnect in what these things are called; here's a list of (IMO) some of them in Most Known to Least Known order:

* Output values (What Terraform says)
* State outputs (What TFC calls them in the API)
* Workspace outputs (what TFC calls them somewhere since workspace overview enhancements)
* Run outputs

I propose to keep the name and any references in the docs here as close to "output values" as possible, given Terraform's the main owner of them; the fact they are 'owned' by a workspace's current configuration version is inherent in the fact that you need to specify a workspace to pull them from, so: `tfe_output_values` ? `tfe_outputs`?

Let 'State Version Outputs' be an oddity of the API naming in TFC only; in the TFC UI they are just called 'outputs' anyway. 

Thoughts?

omarismail:
> Lookin' good! I'll give it a QA roll after some discussion. My biggest overarching bit to discuss is the `sensitive` argument.
> 
> As it stands, `sensitive = true` to _reveal_ the sensitive value is the complete opposite of `sensitive = true` when _defining_ the output value in the other configuration - so I see a lot of confusion ensuing! Given the current mechanism, this will need to clarified to be `show_sensitive` or something...
> 
> ...but beyond that, I wonder why `sensitive` needs to exist at all. Could you elaborate on that a bit?

I actually had `show_sensitive` to begin, but felt that was too verbose, but given the confusion you stated, it would be good to be verbose. So I'll change it to `show_sensitive`.

In terms of why does it need to exist at all. If an output value is marked as sensitive, that means the intent is for it to be sensitive and not displayed, and we should honor that intent and default to it being sensitive unless told otherwise (via `show_sensitive`)

main.go
```diff
@@ -1,11 +1,35 @@
...
+		ctx, tfe.Provider().GRPCProvider, tfe.PluginProviderServer,
```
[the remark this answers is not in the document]
omarismail: I'd prefer to add a comment to explain than to preface it with `Complex`. "Complex" doesn't tell much to the reader besides "get ready, its complicated" 😂 , which...may be a nice warning!

But I'll put a comment here and where I declare `PluginProviderServer` so that when someone looks there, they can see a full explanation.

tfe/data_source_workspace_outputs.go
omarismail: Hmm, good point. I think theres a way to do it

omarismail: I changed this up so I pass the client directly to the provider.

website/docs/d/workspace_outputs.html.markdown
omarismail: Right, will update

omarismail:
> Let 'State Version Outputs' be an oddity of the API naming in TFC only; in the TFC UI they are just called 'outputs' anyway.
> Thoughts?

Fair points. We do have many names, and I like sticking to Terraform as the default naming convention and only deviate from there if there is a good reason, and I don't see a good reason to deviate here. So I'll call them `tfe_outputs`

brandonc:
I'm a little confused about tfe_workspace_outputs. Normal tf outputs reveal sensitive variables _unless_ you specify `sensitive = true` (the inverse of this implementation), which case those outputs are not displayed in CLI output. My instinct is that we should be following the same behavior as [Output Values](https://www.terraform.io/docs/language/values/outputs.html#sensitive-suppressing-values-in-cli-output) but I'm still learning why these two use cases might different.

chrisarcand (changes requested):
Yeah the big thing left is this `sensitive` business that @brandonc alludes to. 

The provider shouldn't concern itself with providing a mechanism for opting in or out of sensitivity because that's exactly what [`sensitive()`](https://www.terraform.io/docs/language/functions/sensitive.html) and [`nonsensitive()`](https://www.terraform.io/docs/language/functions/nonsensitive.html) are for. In particular, check out the documentation on the latter - an entire set of `values` from a particular workspace is exactly the case described with `mixed_content` there.

What this means, I believe, is that if you want to continue to provide a single map of output values: _If_ an output within that set is marked as sensitive, the entire map value should be considered sensitive. It is then the responsibility of the author to fetch the right values, mark things with `nonsensitive`, etc (just as described in the docs). If it _does not_ contain a sensitive value, the entire map need not be considered sensitive.

website/docs/d/outputs.html.markdown
```diff
@@ -0,0 +1,60 @@
...
+sidebar_current: "docs-datasource-tfe-state-outputs"
+description: |-
+  Get output values from another organization/workspace.
+---
+# Data Source: tfe_outputs
+
+This data source is used to retrieve the state outputs for a given workspace.
+It enables output values in one Terraform configuration to be used in another.
+
+Any output value that is marked as sensitive is not hidden.
```
chrisarcand, line 13: This is rather whiplashing without the added context that the _other_ configuration's sensitivity context is not preserved 😆 How about:

```suggestion
Note that any output value marked as sensitive from the remote workspace configuration is not redacted in the _current_ configuration when using this datasource. It is the responsibility of the configuration using this datasource to redetermine sensitive values (e.g. using [sensitive()](https://www.terraform.io/docs/language/functions/sensitive.html) or redeclaring a sensitive output value).
```

tfe/data_source_outputs.go
```diff
@@ -0,0 +1,256 @@
...
+			return nil, fmt.Errorf("Current remote state for workspace '%s' not found.", wsName)
+		}
+		return nil, fmt.Errorf("Could not read the current state for workspace '%s' : %v", wsName, err)
+	}
+
+	log.Printf("[DEBUG] Downloading State Version")
+	rawState, err := tfeClient.StateVersions.Download(ctx, sv.DownloadURL)
+	if err != nil {
+		return nil, fmt.Errorf("Error downloading state: %v", err)
+	}
```
chrisarcand, line 191: Whoops! I missed this earlier. We'll want the ability to be able to pull output values _without_ needing the entire state version; that's one of the clear advantages of using this over `terraform_remote_state`, as with TFC you can give [explicit access to outputs only](https://www.terraform.io/docs/cloud/users-teams-organizations/permissions.html#general-workspace-permissions).

chrisarcand (changes requested):

brandonc (commented):
It worked!

Screen Shot 2021-08-13 at 9 24 15 AM

omarismail:
This PR got a bit long and messy and a few core changes have happened. Opening up a new PR

PR #344, opened 2021-08-20
Add Outputs data source

omarismail:
## Description

This PR introduces a new data source for handling state outputs. `tfe_outputs ` takes an organization and workspace, and retrieves the state outputs for that workspace. 

**Example**

```
data "tfe_outputs" "foobar" {
  organization = "<org-name>"
  workspace = "<workspace-name>"
}

output "states" {
	value = nonsensitive(data.tfe_outputs.foobar.values.<key>)
}
```

The `<key>` in the `states` output can be any field in your state outputs. So if your state output looks something like this:
```
  "outputs": {
    "user": {
      "value": "9023256633839603543",
      "type": "string"
    },
    "records": {
      "value": ["hashicorp.com","terraform.io"],
      "type": ["list","string"]
    }
  }
```

Then the `<key>` can be as follows:
```
data "tfe_outputs" "foobar" {
  organization = "<org-name>"
  workspace = "<workspace-name>"
}

output "user_output" {
	value = nonsensitive(data.tfe_outputs.foobar.values.user)
}
output "records_output" {
	value = nonsensitive(data.tfe_outputs.foobar.values.records)
}
```

## Testing plan

1. Pull Down the branch
2. Create a terraform configuration file that includes the tfe-provider
3. Copy the example code above and make sure the `tfe_outputs ` data source has a organization and workspace that has state outputs. If the workspace doesn't, then the `values` field of the `tfe_outputs ` will be empty.
4. Run terraform and you will see the outputs from that statefile. 
5. Change your terraform config to select a specific value from the output. Re-run terraform and you should see the value printed.

## External links

An RFC was passed internally. 

```
$ TFE_TOKEN=<token> TFE_HOSTNAME=app.terraform.io TF_ACC=1 TF_LOG_PROVIDER=DEBUG  go test ./tfe/... -v -run TestAccTFEStateOutputs
=== RUN   TestAccTFEOutputs
--- PASS: TestAccTFEOutputs (12.08s)
--- PASS: TestAccTFEOutputs_emptyOutputs (12.12s)
PASS
```


website/docs/d/outputs.html.markdown
```diff
@@ -0,0 +1,60 @@
+---
```
omarismail, line 1: image

tfe/data_source_outputs.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (d dataSourceOutputs) readStateOutput(ctx context.Context, tfeClient *tfe.Client, orgName, wsName string) (*stateData, error) {
+	log.Printf("[DEBUG] Reading the Workspace %s in Organization %s", wsName, orgName)
+	ws, err := tfeClient.Workspaces.Read(ctx, orgName, wsName)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading workspace: %v", err)
+	}
+
+	log.Printf("[DEBUG] Reading the current StateVersion for Workspace ID %s.", ws.ID)
+	sv, err := tfeClient.StateVersions.Current(ctx, ws.ID)
```
chrisarcand, line 168: * This defeats the intent of not needing to be able to read state versions to read outputs 😅 
* You shouldn't need to call out for the current state version anyway, it's a needless call; a user without state version access should be able to fetch the ID from the relationships member in the workspace response and use that directly.

tfe/data_source_outputs.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	for _, op := range outputs {
+		buf, err := json.Marshal(op.Value)
+		if err != nil {
+			return nil, fmt.Errorf("Could not marshal output value: %v", err)
+		}
+
+		v := ctyjson.SimpleJSONValue{}
+		err = v.UnmarshalJSON(buf)
```
chrisarcand, line 193: Wow really? I'm sort of surprised it's as easy as using JSON marshalling; I've never really considered how Terraform types and JSON types appear to be truly 1:1.

tfe/data_source_outputs_test.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	wsCleanup = func() {
+		if err := client.Workspaces.Delete(ctx, org.Name, ws.Name); err != nil {
+			t.Errorf("Error destroying workspace! WARNING: Dangling resources\n"+
+				"may exist! The full error is shown below.\n\n"+
+				"Workspace: %s\nError: %s", ws.Name, err)
+		}
+	}
```
chrisarcand, line 144: Do you need this? The organization cleanup should cascade and delete the workspace anyway.

website/docs/d/outputs.html.markdown
```diff
@@ -0,0 +1,60 @@
...
+sidebar_current: "docs-datasource-tfe-state-outputs"
+description: |-
+  Get output values from another organization/workspace.
+---
+# Data Source: tfe_outputs
+
+This data source is used to retrieve the state outputs for a given workspace.
+It enables output values in one Terraform configuration to be used in another.
+
+Note that any output value marked as sensitive from the remote workspace configuration is not redacted in the _current_ configuration when using this datasource. It is the responsibility of the configuration using this datasource to redetermine sensitive values (e.g. using [sensitive()](https://www.terraform.io/docs/language/functions/sensitive.html) or redeclaring a sensitive output value).
```
chrisarcand, line 13: I mentioned this in Slack last week but easy to get lost:

I think the best answer to the sensitive values question we were dealing with is indeed to not touch changing sensitivity from the provider (which this PR does already), with one additional bit: Given that the remote workspace we're pulling from could potentially have sensitive values, the entire value should be marked as sensitive.

We sadly can't do per-output-value marking of sensitivity or any such business, as you can't do the conditional part across the plugin bridge 😭 But Paddy agreed we should probably opt to a safer default and just mark it statically sensitive at the schema level, then, and users can still use the values as normal and even output them by redefining outputs and/or nonsensitive().

TLDR: Given functionality constraints, let's statically mark the whole map as sensitive always.

chrisarcand (changes requested):
This is looking _awesome_. Really awesome testing too. Nice work!

Couple things, and I'll wait to test drive it a bit until we resolve them:

tfe/data_source_outputs.go
omarismail: Oops, this was an artifact of the last attempt, and will change it to use the workspace response. :sigh:, stale context.

tfe/data_source_outputs_test.go
omarismail: That's true. Will remove

website/docs/d/outputs.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+  workspace = "my-workspae"
+}
+
+output "hello" {
+	value = data.tfe_outputs.foobar.values.foo
+}
+
+output "world" {
+	value = data.tfe_outputs.foobar.values.bar
+}
```
chrisarcand, line 50: You may want to use a different example, as 'declaring an output value from an output value' is more complicated than I thought, QAing this 😬

Basically, I didn't know this was a thing:

image

If you declare `sensitive = true` here, it then allows the output to be completed and present in the state file - but the CLI output is still sensitive (`<sensitive>`). So to both successfully output the values and show it in the output requires `nonsensitive()` as well. 

Given that using both of those starts to make the example look overly complicated and more about sensitivity than the actual usage of an output value from one workspace here, I'd consider doing something like...

```hcl
data "tfe_outputs" "foo" {
  organization = "my-org"
  workspace = "my-workspace"
}

resource "random_id" "vpc_id" {
  keepers = {
    # Generate a new ID any time the value of 'bar' in workspace 'my-org/my-workspace' changes.
    bar = data.tfe_outputs.foo.values.bar
  }

  byte_length = 8
}

```

chrisarcand (approved):
Just some doc changes to make, and then this is good to go! ✅

website/docs/d/outputs.html.markdown
omarismail: fair point 👍

PR #339, opened 2021-08-10
Add Workspace Tags

thrashr888:
## Description

```hcl
// adding tags
resource "tfe_workspace" "test" {
  name         = "my-workspace-name"
  organization = "my-organization"
  tag_names    = ["prod", "database", "app"]
}

// linking resources by tags
data "tfe_workspace_ids" "prod-apps" {
  tag_names    = ["prod", "app", "aws"]
  organization = "my-org-name"
}

// now you can attach resources via tag
resource "tfe_policy_set" "test" {
  name          = "my-policy-set"
  description   = "A brand new policy set"
  organization  = "my-org-name"
  workspace_ids = values(data.tfe_workspace_ids.prod-apps.ids)
}

// resources available via `workspace_ids` or looped `workspace_id`:
// - tfe_policy_set
// - tfe_variable
// - tfe_notification_configuration
// - tfe_team_access
// - tfe_run_trigger
```

## TODO

- [x] Add tests
- [x] Get tags changes from go-tfe

## Testing plan

1.  check the docs

## External links

docs and go-tfe coming soon

## Output from acceptance tests

_Please run applicable acceptance tests locally and include the output here. See TESTS.md to learn how to run acceptance tests._

_If you are an external contributor, your contribution(s) will first be reviewed before running them against the project's CI pipeline._

```
$ TESTARGS="-run TestAccTFEWorkspace" make testacc 

...
```

PR #348, opened 2021-08-31
Update changelog for v0.26.0

omarismail:
## Description

This updates the changelog in preparation for v0.26.0. 

All changes since last release: https://github.com/hashicorp/terraform-provider-tfe/compare/v0.25.3...main

There was a go mod prompt because some packages should not be `indirect` (probably a mistake from a previous PR), so ran `go mod tidy` and included changes here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 FEATURES:
 * **New Data Sources:** d/tfe_organizations, d/tfe_organization [#320](https://github.com/hashicorp/terraform-provider-tfe/pull/320)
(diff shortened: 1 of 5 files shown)


chrisarcand (commented):
🎉

I believe we're just waiting on #339 to make it in for this release, then this looks great!

chrisarcand (approved):
For posterity, we've decided to proceed with 0.26.0 because workspace tag support may require a bit more time and there's too many highly anticipated goodies in this release to delay it further. :shipit: 🎉

PR #354, opened 2021-09-03
Fix 0.26.0 configuration regressions

### LINKED ISSUE #353: v2.6.0 provider error
Only upgrading the provider version constraint to `~> 0.26.0` now shows:
```
╷
│ Error: Invalid provider configuration
│
│ Provider "registry.terraform.io/hashicorp/tfe" requires explicit
│ configuration. Add a provider block to the root module and configure the
│ provider's required arguments as described in the provider documentation.
│
╵
╷
│ Error: Error retrieving provider meta values from provider request
│
│   with provider["registry.terraform.io/hashicorp/tfe"],
│   on <empty> line 0:
│   (source code not available)
│
│ Error retrieving provider meta values from provider request the hostname
│ and token must be present.
```
This is using Terraform Cloud and `TFE_ENGINE` environment variable is set within the workspace. The runs are remote.

chrisarcand:
## Description

Closes #353 

The 0.26.0 release introduced features with #344 that required muxing the existing provider built with [terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk) and another one built with a lower level provider package, [terraform-plugin-go](https://github.com/hashicorp/terraform-plugin-go). What this means is that there are two providers which must interpret the same configuration/settings, in one. 

The SDK schema [usually provides a default hostname value with DefaultFunc](https://github.com/hashicorp/terraform-provider-tfe/blob/7bb6b0f0b1717c5bcd722d14940ca6c93f0c8874/tfe/provider.go#L54), and unfortunately this behavior was not translated to an equivalent in new lower level provider now muxed in, causing a regression where unless a hostname is explicitly set in configuration or `TFE_HOSTNAME`, the provider configuration will error. 

As an aside, `ssl_skip_verify` was also not respected by the muxed plugin provider, and has also been corrected here.

### DIFF
--- a/tfe/plugin_provider.go
+++ b/tfe/plugin_provider.go
@@ -3,7 +3,6 @@ package tfe
 import (
 	"context"
 	"fmt"
-	"os"
 
 	tfe "github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
@@ -36,8 +35,9 @@ func (e errUnsupportedResource) Error() string {
 }
 
 type providerMeta struct {
-	token    string
-	hostname string
+	token         string
+	hostname      string
+	sslSkipVerify bool
 }
 
 func (p *pluginProviderServer) GetProviderSchema(ctx context.Context, req *tfprotov5.GetProviderSchemaRequest) (*tfprotov5.GetProviderSchemaResponse, error) {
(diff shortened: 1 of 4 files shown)


omarismail:
Tests are as follows:

**Tests with no provider block**
1. Remove the `credentials` block in your `~/.terraformrc` that has the hostname `app.terraform.io`. This should cause a failure.
2. Add a credentials block, but a different hostname. This too should cause a failure.
3. Add a credentials block using the "app.terraform.io" and a valid token. This should pass.
4. Add a valid token in the credentials block in `~/.terraformrc` for `TFE_HOSTNAME=app.staging.terraform.io ` and it should work.
5. Add an invalid token in the credentials block in `~/.terraformrc` for `TFE_HOSTNAME=app.staging.terraform.io ` and it should fail.

_env variable tests_
1.  Set the `TFE_HOSTNAME=<host> TFE_TOKEN=<token>` to something different than in your `.terraformrc` (if you don't have a provider block) and this works.
2. Set only the `TFE_HOSTNAME=<host>` to something different than your `terraformrc` and it should fail saying missing token
3. Set only the `TFE_TOKEN=<token>` to something different than your `terraformrc` and different from `provider` block, and it should succeed if the token is valid and fail if the the token is invalid.

**Tests with provider block**
1. Add only a hostname `app.terraform.io` in the provider block and it should work. Assuming the `.terraformrc` has the credentials block
2. Add only a hostname that is **not** in `.terraformrc` and ensure it fails.
3. Add a new credentials block in `terraformrc`  using new valid hostname/token, and use the hostname in the provider and it should work.
4. Add a hostname and new token in the provider block and it should work.

_env variable test_
1. Set the `TFE_HOSTNAME=<host>` `TFE_TOKEN=<token>` `TFE_SSL_SKIP_VERIFY=true` (all combinations of the three) to anything and terraform should respect what is in the provider block.

lafentres (approved):
i tested a local build of this branch with a few different variations of env vars and credential blocks and they all seem to work as described and as i'd expect.

omarismail (approved):

nfagerlund (approved):
I exercised this pretty thoroughly, including using the `TERRAFORM_CONFIG` env var to grab credentials from an explicit file, and it all worked exactly as I would hope! I'll rebase onto this to continue working on the auto-discovered credentials file stuff.

PR #358, opened 2021-09-03
d/workspace_ids use AtLeastOneOf to ensure names/tag_names is present

omarismail:
## Description

In the recent [v0.26.0 release](https://github.com/hashicorp/terraform-provider-tfe/blob/main/CHANGELOG.md#0260-september-02-2021), there was a regression regarding the `names` attribute in the data source `workspace_ids`. 

Prior to v0.26.0, `names` was required at the schema level. But this was changed with the addition of `tag_names` and neither was required **at the schema level** but the user was presented [with an error](https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/data_source_workspace_ids.go#L54-L56) when the read function `dataSourceTFEWorkspaceIDs` was called. 

This PR changes the `names` field to be optional but uses the [AtLeastOneOf](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema.AtLeastOneOf) schema helper which states that
> AtLeastOneOf is a set of schema keys that, when set, at least one of the keys in that list must be specified.

_Note:_ The [documentation does not need](https://github.com/hashicorp/terraform-provider-tfe/blob/main/website/docs/d/workspace_ids.html.markdown#argument-reference) to be updated.

## Testing plan

Create a terraform config and exclude both `tag_names` and `names`

```
data "tfe_workspace_ids" "foobar" {
  organization = "<org-name>
}
```

This should return a Missing required argument error:

```
│ Error: Missing required argument
│
│   with data.tfe_workspace_ids.foobar,
│   on main.tf line 15, in data "tfe_workspace_ids" "foobar":
│   15: data "tfe_workspace_ids" "foobar" {
│
│ "names": one of `names,tag_names` must be specified
```

And then set `names`, and then `tag_names` alone, and each should work.

```
$ terraform-provider-tfe TFE_TOKEN=<redacted> TFE_HOSTNAME=app.terraform.io TF_ACC=1 TF_LOG_PROVIDER=DEBUG  go test ./tfe/... -v -run TestAccTFEWorkspaceIDs
=== RUN   TestAccTFEWorkspaceIDsDataSource_basic
--- PASS: TestAccTFEWorkspaceIDsDataSource_basic (8.85s)
=== RUN   TestAccTFEWorkspaceIDsDataSource_wildcard
--- PASS: TestAccTFEWorkspaceIDsDataSource_wildcard (8.96s)
=== RUN   TestAccTFEWorkspaceIDsDataSource_tags
--- PASS: TestAccTFEWorkspaceIDsDataSource_tags (7.77s)
=== RUN   TestAccTFEWorkspaceIDsDataSource_searchByTagAndName
--- PASS: TestAccTFEWorkspaceIDsDataSource_searchByTagAndName (5.47s)
=== RUN   TestAccTFEWorkspaceIDsDataSource_empty
--- PASS: TestAccTFEWorkspaceIDsDataSource_empty (0.91s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/tfe 32.275s
```

### DIFF
--- a/tfe/data_source_workspace_ids.go
+++ b/tfe/data_source_workspace_ids.go
@@ -14,17 +14,16 @@ func dataSourceTFEWorkspaceIDs() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"names": {
-				Type:     schema.TypeList,
-				Elem:     &schema.Schema{Type: schema.TypeString},
-				Optional: true,
-				Required: false,
+				Type:         schema.TypeList,
+				Elem:         &schema.Schema{Type: schema.TypeString},
+				Optional:     true,
+				AtLeastOneOf: []string{"names", "tag_names"},
 			},
 
 			"tag_names": {
 				Type:     schema.TypeList,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Optional: true,
-				Required: false,
 			},
 
 			"organization": {
@@ -51,10 +50,6 @@ func dataSourceTFEWorkspaceIDsRead(d *schema.ResourceData, meta interface{}) err
 	// Get the organization.
 	organization := d.Get("organization").(string)
 
-	if len(d.Get("names").([]interface{})) == 0 && len(d.Get("tag_names").([]interface{})) == 0 {
-		return fmt.Errorf("Either `names` or `tag_names` is required")
-	}
-
 	// Create a map with all the names we are looking for.
 	var id string
 	names := make(map[string]bool)
@@ -62,6 +57,7 @@ func dataSourceTFEWorkspaceIDsRead(d *schema.ResourceData, meta interface{}) err
 		id += name.(string)
 		names[name.(string)] = true
 	}
+	isWildcard := names["*"]
 
 	// Create two maps to hold the results.
 	fullNames := make(map[string]string, len(names))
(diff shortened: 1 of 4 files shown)


tfe/data_source_workspace_ids.go
```diff
@@ -14,17 +14,16 @@ func dataSourceTFEWorkspaceIDs() *schema.Resource {
...
+				Elem:         &schema.Schema{Type: schema.TypeString},
+				Optional:     true,
+				AtLeastOneOf: []string{"names", "tag_names"},
 			},
 
 			"tag_names": {
 				Type:     schema.TypeList,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Optional: true,
-				Required: false,
```
omarismail, line 27: Only one of the fields needs to have `AtLeastOneOf`. Not both

thrashr888 (approved):
ah. makes sense to me. 👍🏻

nikolay:
@omarismail There's another bug with tags, too: https://github.com/hashicorp/terraform-provider-tfe/issues/359

chrisarcand (changes requested):
Looks good, but there's one more bit of divergent behavior from the previous release. Given the following:

```hcl
data "tfe_workspace_ids" "foobar" {
  organization = tfe_organization.test.name
  names = []
}
```

In 0.25.x, the result would be empty, requiring you to use `["*"]` to fetch all names. Now, the result is all of them, because of how tags was added. 

Although the expected behavior of `name = []` is somewhat under specified previously, an internal team seems to rely on this behavior and we should probably preserve it.

chrisarcand (approved):
Nice work, this LGTM 👍🏻

PR #360, opened 2021-09-04
Improve discovery and loading of credentials from Terraform config files

nfagerlund:
## Description

This provider is somewhat special, because the API token it uses is also used
by Terraform CLI itself. We can give a nicer experience on the local CLI by
auto-discovering Terraform's copy of the token (if one isn't provided in the
provider config or the `TFE_TOKEN` environment variable).

However, we weren't quite matching Terraform's credential-finding behavior:

- We were respecting the `TERRAFORM_CONFIG` environment variable for the main
  config file location, but not `TF_CLI_CONFIG_FILE`. (Terraform will use
  either, but prefers the latter since it matches the other environment variable
  names it uses.)
- We weren't ever consulting the credentials file
  (`~/.terraform.d/credentials.tfrc.json`, on unix), which is written to by
  `terraform login` and/or credential helper scripts. The HCL library
  understands the credentials JSON, so you could work around this by pointing
  `TERRAFORM_CONFIG` at the credentials file, but if there was anything else
  important in the config file you also had to set `TF_CLI_CONFIG_FILE` so that
  Terraform (but not the provider) could still find it.

This commit fixes both of those: we'll accept a non-default config file location
via either environment variable (preferring `TF_CLI_CONFIG_FILE`), and we'll
_also_ consult the credentials file if present.

If both sources include a token for the desired hostname, the provider will
follow Terraform's behavior and prefer the token from the CLI config file.

## Testing plan

1.  Create a basic config that configures the `tfe` provider and maybe manages one resource. In the provider config, provide a hostname (if you aren't using app.terraform.io) but not a token. 
2. Ensure that your environment doesn't have any values set for `TFE_TOKEN`, `TF_CLI_CONFIG_FILE`, or `TERRAFORM_CONFIG`.
3. Ensure that neither your `~/.terraformrc` or `~/.terraform.d/credentials.tfrc.json` files include a token for the configured TFE host, to start out with.
4. Then:

Note: for the environment variable columns, "x" means "unset"; read any other value as "points to a loadable .terraformrc config file containing `_____` for the configured hostname".

.terraformrc | credentials json | `TF_CLI_ CONFIG_FILE` | `TERRAFORM_ CONFIG` | **Outcome**
-|-|-|-|-
valid token | no token | x | x | Works (finds default config file)
no token | valid token | x | x | Works (finds credentials file, merges it with config file)
valid token | bad token | x | x | Works (config file beats credentials file)
no token | no token | valid token | x | Works (new env var works)
no token | no token | x | valid token | Works (old env var works)
bad token | no token | valid token | x | Works (env var beats config file)
no token | bad token | valid token | x | Works (env var beats credentials file) 
no token | no token | valid token | bad token  | Works (new env var beats old)
valid token | no token | no token | x | Fails (env var replaces config file instead of merging with it)
no token | valid token | no token | x | Works (credentials file is still merged with config file, even if config file is non-default)

See CI output in the checks below.


website/docs/index.html.markdown
```diff
@@ -56,14 +56,17 @@ If you are using this provider on your local command line without remote operati
...
+Terraform on your local command line, it can automatically discover the credentials generated by
+[`terraform login`](https://www.terraform.io/docs/commands/login.html).
+- **Set a `credentials` block in your CLI config file (.terraformrc):** See
 the [CLI Configuration File documentation](/docs/commands/cli-config.html).
+    - If you used the `TF_CLI_CONFIG_FILE` environment variable to specify a
+      non-default location for .terraformrc, the provider will also use that
+      location.
+    - If you want the provider to use credentials from a different config file
+      than Terraform itself uses, set the `TERRAFORM_CONFIG` environment
+      variable to the file path the provider should use.
```
nfagerlund, line 69:
```suggestion
If you used the `TF_CLI_CONFIG_FILE` environment variable to specify a
non-default location for .terraformrc, the provider will also use that location.
```

Bah, I forgot to re-update the docs after re-discovering the history of those two environment variables. We should return `TERRAFORM_CONFIG` to its resting place in obscurity!

tfe/provider.go
```diff
@@ -282,36 +282,71 @@ func getClient(tfeHost, token string, insecure bool) (*tfe.Client, error) {
...
-	config := &Config{}
-
-	// Detect the CLI config file path.
-	configFilePath := os.Getenv("TERRAFORM_CONFIG")
+	mainConfig := &Config{}
+	combinedConfig := &Config{}
+
+	// There might be credentials in the main CLI config file (manually entered)
+	// AND/OR the credentials file (auto-configured by terraform login or a
+	// credentials helper), so we need to consult both. As per the behavior of
```
chrisarcand, line 290: This doesn't work with a credentials helper, no? 

FWIW we could support them, the logic is here: https://github.com/hashicorp/terraform-svchost/blob/f050f53b97348772833a8f7fc1b297ec71f9fc7c/auth/helper_program.go

But this PR is already 💯 without all that, for sure 👍🏻 I'd just amend this comment.

chrisarcand (approved):
This LGTM, though admittedly I don't have the time to properly go through the testing plan at the moment. Really really great PR!

tfe/provider.go
nfagerlund: Oh! I thought credentials helpers used the credentials file as a cache, though now I've forgotten where I would have gotten that idea. 

I'll just amend the comment to only refer to login. 👍🏼

tfe/provider.go
```diff
@@ -282,36 +282,71 @@ func getClient(tfeHost, token string, insecure bool) (*tfe.Client, error) {
...
+	// Terraform itself, the main config file wins if both sources have
+	// credentials for a given host.
```
nfagerlund, line 292:
```suggestion
	// There might be credentials in the main CLI config file (manually entered)
	// AND/OR the credentials file (auto-configured by terraform login), so we
	// need to consult both. As per the behavior of Terraform itself, the main
	// config file wins if both sources have credentials for a given host.
```

omarismail (changes requested):
Functionally this is great! My main concern is just adding tests.

I put an example here:

https://github.com/hashicorp/terraform-provider-tfe/commit/ad40849a148691621f8ac8b75483b475dff60e2e

tfe/provider.go
```diff
@@ -282,36 +282,70 @@ func getClient(tfeHost, token string, insecure bool) (*tfe.Client, error) {
...
+	// need to consult both. As per the behavior of Terraform itself, the main
+	// config file wins if both sources have credentials for a given host.
+
+	// To find the main CLI config file, follow Terraform's own logic: try
+	// TF_CLI_CONFIG_FILE, then try TERRAFORM_CONFIG, then try the default
+	// location.
+	configFilePath := os.Getenv("TF_CLI_CONFIG_FILE")
+	if configFilePath == "" {
+		configFilePath = os.Getenv("TERRAFORM_CONFIG")
+	}
```
omarismail, line 299: What do you think about refactoring this a tad bit for readability? I added an example here https://github.com/hashicorp/terraform-provider-tfe/commit/ad40849a148691621f8ac8b75483b475dff60e2e

tfe/provider_test.go
```diff
@@ -125,6 +125,155 @@ func TestProvider_versionConstraints(t *testing.T) {
...
+	}
+	defer reset()
+
+	// Summary of fixtures: the credentials file and .terraformrc file each have
+	// credentials for two hosts, and they both have credentials for
+	// app.terraform.io
+	hasCredentials := "test-fixtures/env/home"
+	noCredentials := "test-fixtures/env/empty"
+	terraformrc := "test-fixtures/env/terraformrc"
+	noTerraformrc := "test-fixtures/env/nothing"
```
omarismail, line 214: I know i'm the one who made this directory path, so I blame my older self 😆 . I think if I was looking at the directory path, i'd be confused what "env" referred to. Could we change these paths to be `test-fixtures/tf-env-files/*`

omarismail: And nit comment, could we be more specific for the `empty` and `nothing`? So like `test-fixtures/tf-env-files/no-terraformrc` and `test-fixtures/tf-env-files/no-credentials`

tfe/provider_test.go
```diff
@@ -125,6 +125,155 @@ func TestProvider_versionConstraints(t *testing.T) {
...
+	fromTerraformrc := "something.atlasv1.prod_rc_file"
```
omarismail, line 215: Nit as well, `s/fromTerraformrc/tokenFromTerraformrc`

omarismail (approved):
A few minor nits, but overall amazing job! :shipit: once tests pass

tfe/provider_test.go
nfagerlund: 👍🏼 !

I went with `test-fixtures/cli-config-files/*` for the path name.

Issue #317, opened 2021-05-19
resource/tfe_variable: the parameter `sensitive` should be `ForceNew` when removing `sensitive`


gdavison:
A variable cannot be converted from `sensitive = true` to `sensitive = false` in-place. The API returns the error

>Sensitive cannot be changed from true to false on saved records. To make this variable readable, destroy this record and create a new one with the same key value.

A `CustomizeDiff` function could be used to set `ForceNew` on `sensitive` if changing from `true` to `false`

chrisarcand:
Indeed, good call!

fatbasstard:
Ah, just got hit with this error... Pretty annoying, would be great if this can be fixed/implemented

PR #394, opened 2021-12-02
Forcenew sensitive

uturunku1:
## Description

This PR addresses this github issue: https://github.com/hashicorp/terraform-provider-tfe/issues/317

## Testing plan

1.  _the conditions_

## External links

```

...
```


tfe/resource_tfe_variable.go
```diff
@@ -85,9 +84,25 @@ func resourceTFEVariable() *schema.Resource {
...
+		Destroy and add a new resource when:
+		1. the parameter key changed and the param sensitive is set to true
+		2. the parameter sensitive changed from true to false
+	*/
+	return func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+		wasSensitiveVar := d.HasChange("sensitive") && !(d.Get("sensitive").(bool))
+
+		if wasSensitiveVar {
+			d.ForceNew("sensitive")
+		} else if d.Get("sensitive").(bool) {
```
sebasslash, line 102: should the check here be `d.HasChange("key") && d.Get("sensitive").(bool)` ? I'm basing this off the comment you left... feel free to clarify here.

tfe/resource_tfe_variable.go
```diff
@@ -85,9 +84,25 @@ func resourceTFEVariable() *schema.Resource {
...
+	/*
+		Destroy and add a new resource when:
+		1. the parameter key changed and the param sensitive is set to true
+		2. the parameter sensitive changed from true to false
+	*/
+	return func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+		wasSensitiveVar := d.HasChange("sensitive") && !(d.Get("sensitive").(bool))
+
+		if wasSensitiveVar {
+			d.ForceNew("sensitive")
```
sebasslash, line 101: You should `return d.ForceNew("sensitive")` as that function returns an error and you want to capture that if it happens.

tfe/resource_tfe_variable.go
uturunku1: @sebasslash Sure! I would not mind adding `else if d.HasChange("key") && d.Get("sensitive").(bool)` instead of `else if d.Get("sensitive").(bool) ` if that help with readability.  But otherwise checking for `d.HasChange("key")` is not needed because d.ForceNew("key") is already doing that check for me. Let me know if I am reading this right:

https://github.com/hashicorp/terraform-plugin-sdk/blob/56627b161e2cf04e9350e77fc9c303ecbae9c5bc/helper/schema/resource_diff.go#L341

sebasslash: When `ForceNew()` is called and the key specified has not changed it will throw an error which is not something we want. However if this is intended behavior (which I doubt), we could add a comment then for readability, something like: `// We don't explicitly check "key" param has changed because ForceNew() takes care of this check`.

My money is on preventing this error from being thrown in the first place so I would go with the preemptive check.

tfe/resource_tfe_variable.go
uturunku1: Indeed!

sebasslash (approved):

PR #404, opened 2022-01-06
Fix doc bug in tfe_policy_set resource example

richard-russell:
## Description

The example usage for the tfe_policy_set resource contains the following ```oauth_token_id     = tfe_oauth_client.test.id```. This doesn't work, as ```id``` refers to the oauth client id, not the token id. The correct working code is ```oauth_token_id     = tfe_oauth_client.test.oauth_token_id```.

## Testing plan

I've included the output from the doc preview tool below.

Documentation preview:
image


nfagerlund (approved):
You're right! Thanks for noticing and fixing this.

brandonc:
I reopened this as #406, rebased on the fixed test suite. Thanks for the fix!

PR #408, opened 2022-01-14
Import Terraform Versions using version number

sebasslash:
## Description

This is a very simple PR that introduces the ability to use a Terraform version number as an alternative import ID when importing Terraform versions. I'm not particularly proud of the implementation to search for a tool version that matches the version specified, but it's currently an API limitation that doesn't allow querying/filtering Terraform versions by version. It effectively loops through the [List Terraform Versions](https://www.terraform.io/cloud-docs/api-docs/admin/terraform-versions#list-all-terraform-versions) endpoint until it finds a match. If no match is found, the import will fail.

## Testing plan

1.  Ensure you have a `.tfrc` file that overrides the Terraform provider to your locally compiled binary. It should look something like this:

```
provider_installation {
    dev_overrides {
        "hashicorp/tfe" = "path/to/bin" # Usually $GOPATH/bin
    }

    direct {}
}
```
2. Create a sample Terraform configuration, I'm omitting the `tfe` provider block:

```hcl
resource "tfe_terraform_version" "1_1_2" {}
```

3. Import, ensure you override `TF_CLI_CONFIG_FILE` var! 

```sh
TF_CLI_CONFIG_FILE=path/to/my/tfrc terraform import tfe_terraform_version.1_1_2 1.1.2
```

To run the acceptance test

```
$ TESTARGS="-run TestAccTFETerraformVersion_import" make testacc 
```

### DIFF
--- a/tfe/resource_tfe_terraform_version.go
+++ b/tfe/resource_tfe_terraform_version.go
@@ -3,6 +3,7 @@ package tfe
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	tfe "github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -15,7 +16,7 @@ func resourceTFETerraformVersion() *schema.Resource {
 		Update: resourceTFETerraformVersionUpdate,
 		Delete: resourceTFETerraformVersionDelete,
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			State: resourceTFETerraformVersionImporter,
 		},
 
 		Schema: map[string]*schema.Schema{
@@ -133,4 +134,22 @@ func resourceTFETerraformVersionDelete(d *schema.ResourceData, meta interface{})
 	}
 
 	return nil
+}
+
+func resourceTFETerraformVersionImporter(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	tfeClient := meta.(*tfe.Client)
+
+	// Splitting by '-' and checking if the first elem is equal to tool
+	// determines if the string is a tool version ID
+	s := strings.Split(d.Id(), "-")
+	if s[0] != "tool" {
+		versionID, err := fetchTerraformVersionID(d.Id(), tfeClient)
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving terraform version %s: %w", d.Id(), err)
+		}
+
+		d.SetId(versionID)
+	}
+
+	return []*schema.ResourceData{d}, nil
 }
--- a/tfe/tool_helpers.go
+++ b/tfe/tool_helpers.go
@@ -1,0 +1,46 @@
+package tfe
+
+import (
+	"fmt"
+
+	tfe "github.com/hashicorp/go-tfe"
+)
+
+// fetchTerraformVersionID returns a Terraform Version ID for the given Terraform version number
+func fetchTerraformVersionID(version string, client *tfe.Client) (string, error) {
+	versionID := ""
+	pageNum := 0
+
+found:
+	for {
+		versions, err := client.Admin.TerraformVersions.List(ctx, tfe.AdminTerraformVersionsListOptions{
+			ListOptions: tfe.ListOptions{
+				PageNumber: pageNum,
+				PageSize:   20,
+			},
+		})
+		if err != nil {
+			return "", fmt.Errorf("error reading Terraform versions: %w", err)
+		}
+
+		// we've run out of versions to search
+		if len(versions.Items) == 0 {
+			break
+		}
+
+		for _, v := range versions.Items {
+			if v.Version == version {
+				versionID = v.ID
+				break found
+			}
+		}
+
+		pageNum++
+	}
+
+	if versionID == "" {
+		return "", fmt.Errorf("terraform version not found")
+	}
+
+	return versionID, nil
+}
--- a/website/docs/r/terraform_version.html.markdown
+++ b/website/docs/r/terraform_version.html.markdown
@@ -39,10 +39,14 @@ The following arguments are supported:
 
 ## Import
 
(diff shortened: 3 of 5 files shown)


tfe/tool_helpers.go
```diff
@@ -0,0 +1,42 @@
+package tfe
+
+import (
+	"fmt"
+
+	tfe "github.com/hashicorp/go-tfe"
+)
+
+// fetchTerraformVersionID returns a Terraform Version ID for the given Terraform version number
+func fetchTerraformVersionID(version string, client *tfe.Client) (string, error) {
```
brandonc, line 10: Is it an error when the specified version is not found? I'm thinking it should be but it's not reflected in the code. (The caller does not check for empty id)

tfe/tool_helpers.go
```diff
@@ -0,0 +1,42 @@
...
+	versionID := ""
+	pageNum := 0
+
+found:
+	for {
+		versions, err := client.Admin.TerraformVersions.List(ctx, tfe.AdminTerraformVersionsListOptions{
```
brandonc, line 16: since this requires admin access, it might not be a general purpose helper. I'd suggest just keeping this function within tfe/resource_tfe_terraform_version.go

website/docs/r/terraform_version.html.markdown
```diff
@@ -39,10 +39,14 @@ The following arguments are supported:
...
 
-Terraform versions can be imported; use `<TERRAFORM VERSION ID>` as the import ID. For example:
+Terraform versions can be imported; use `<TERRAFORM VERSION ID>` or `<TERRAFORM VERSION NUMBER>` as the import ID. For example:
 
 ```shell
 terraform import tfe_terraform_version.test tool-L4oe7rNwn7J4E5Yr 
 ```
 
+```shell
+terraform import tfe_terraform_version.test 1.1.2
```
brandonc, line 49: Version numbers do not uniquely identify terraform versions, unfortunately, and people use them in weird ways. What do you think about an additional import string like:

`terraform import tfe_terraform_version.test official/1.1.2`

...at least then we can extend this in the future, and further narrow the potential matches by official attribute == true

tfe/tool_helpers.go
brandonc, line 10: it might be worth returning the entire tool version struct since you have it

brandonc (changes requested):

website/docs/r/terraform_version.html.markdown
brandonc: this prefix string also fixes the brittleness of splitting on "." (version numbers may not even need this!)

Screen Shot 2022-01-14 at 9 41 48 AM

sebasslash: So that was my original implementation using `official/<VERSION NUMBER>` for `offical = true` and `internal/<VERSION NUMBER>` for `false`, but my understanding was that the `version` attribute is unique.

tfe/tool_helpers.go
sebasslash: What would the struct be used for? When a resource is imported, the provider fetches the resource using the given ID.  The goal of a custom importer function is to retrieve/parse the ID the provider will use to fetch the resource (even though you're technically fetching twice)

website/docs/r/terraform_version.html.markdown
brandonc: I just checked, uniqueness is enforced by rails and I was mistaken! but splitting on "." thing is an issue.

tfe/tool_helpers.go
sebasslash: My hesitation with this is that it deviates from the rest of the code organization... even though moving the function to another file changes nothing (since it's the same module). Alternatively I can rename `tool_helpers.go` to `admin_helpers.go` which better indicates these are helpers for admins ? WDYT?

Since I think this is the first "Admin-Only" resource we've added to the provider,  it's open ended on how best to structure this.

website/docs/r/terraform_version.html.markdown
sebasslash: Yep, I was iffy on splitting by `.` (was not aware versions didn't have to follow semver!) and am open to alternative solutions. We also have to keep in mind the practitioner can supply either a tool version external ID or the version number itself so I have to have a reliable way of distinguishing between the two. I guess what we can do is split on `-` and if the first elem is `tool` then it's an external ID and passthrough.

tfe/resource_tfe_terraform_version.go
```diff
@@ -134,3 +135,21 @@ func resourceTFETerraformVersionDelete(d *schema.ResourceData, meta interface{})
...
 	return nil
 }
+
+func resourceTFETerraformVersionImporter(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	tfeClient := meta.(*tfe.Client)
+
+	// Splitting by '-' and checking if the first elem is equal to tool
+	// determines if the string is a tool version ID
+	s := strings.Split(d.Id(), "-")
+	if s[0] != "tool" {
```
brandonc, line 145: This is much less brittle 👍

tfe/resource_tfe_terraform_version.go
```diff
@@ -134,3 +135,21 @@ func resourceTFETerraformVersionDelete(d *schema.ResourceData, meta interface{})
...
+		versionID, err := fetchTerraformVersionID(d.Id(), tfeClient)
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving terraform version %s: %v", d.Id(), err)
```
brandonc, line 148: 💅 use %w when wrapping other errors with fmt.Errorf

brandonc (approved):

tfe/resource_tfe_terraform_version.go
sebasslash: Huh was not aware of this, always defaulted to %v! Will add thanks!

PR #415, opened 2022-01-21
fix can't rename workspace while run is being performed

brandonc:
## Description

Test fixes only: These names were used as proxies for a successful applies, I think, so I just moved the strings to the description field


uturunku1 (approved):
🏁🐎

PR #417, opened 2022-01-21
Optimize Terraform Version Importer func

sebasslash:
## Description

This is a simple PR that makes use of the newly added query params to filter a Terraform version when importing a Terraform version by version number. This will drastically reduce the calls needed to find the tool version ID. 

This PR also bumps the go-tfe version to `0.24` which includes the query params as list opts. 

## External links

- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/408)


uturunku1 (approved):
Sebastian gave me some context to understand these changes, and seems like these changes are safe and it's more of a small complement to other PRs 🎉

Issue #171, opened 2020-05-10
Unclear error when attempting to add teams to free accounts


tedivm:
If you attempt to add a team to a free account it gives an error about not being able to find the organization. This was easily resolved by upgrading the account, but it wasn't clear that that was the needed step

koikonom:
Hi @tedivm ,

Unfortunately the messages are coming straight from TFC, we cannot really from the provider side if the error is caused because of an account limitation or because the organization is not really there.

tedivm:
Isn't this project maintained by Hashicorp, the company that makes Terraform Cloud?

koikonom:
Apologies, I should have elaborated. You are correct, this project is maintained by HashiCorp but there are different teams with different backlogs and roadmaps handling the two sides (one side being the provider and the other being the TFC platform). 

I have already created an internal issue for the TFC team to prioritize and address, but until the service API can send a more complete error message, there is no other way for the provider to detect this behavior and respond accordingly.

magzim21:
`Error: Error creating team owners for organization <org>: missing entitlements to create teams`. 
Activating standard plan solved this.

PR #418, opened 2022-01-24
adds entitlment error message for creating Teams

Mfhodges:
## Description

This PR resolves [Issue 171](https://github.com/hashicorp/terraform-provider-tfe/issues/171). Previously when a User on the Free Tier tried to create a team using the tfe provider, the CLI would return an error message like : `Error: Error creating team my-team-name for organization raoul-staging: resource not found`. This error does not explain to the user that they cannot create a team because the Free tier does not have the proper entitlements. 

These changes check entitlements after the `ErrResourceNotFound` error and if there is an error surfaces an error message like: `Error creating team my-team-name for organization raoul-staging: missing entitlements to create teams` 

## Testing plan

1.  Run tfe provider locally against staging.
1.  Create an Organization and set the Feature Set to Free tier.
1. Using the CLI run `terraform apply` to create a team in the Org - This should surface an error due to entitlements 
1. Set the Organization to Business and re-run the `apply` - This should not surface any error.

### DIFF
--- a/tfe/resource_tfe_team.go
+++ b/tfe/resource_tfe_team.go
@@ -107,6 +107,9 @@ func resourceTFETeamCreate(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		if err == tfe.ErrResourceNotFound {
 			entitlements, _ := tfeClient.Organizations.Entitlements(ctx, organization)
+			if entitlements == nil {
+				return fmt.Errorf("Error creating team %s for organization %s: %v", name, organization, err)
+			}
 			if !entitlements.Teams {
 				return fmt.Errorf("Error creating team %s for organization %s: missing entitlements to create teams", name, organization)
 			}
--- a/tfe/resource_tfe_team.go
+++ b/tfe/resource_tfe_team.go
@@ -105,8 +105,13 @@ func resourceTFETeamCreate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[DEBUG] Create team %s for organization: %s", name, organization)
 	team, err := tfeClient.Teams.Create(ctx, organization, options)
 	if err != nil {
-		return fmt.Errorf(
-			"Error creating team %s for organization %s: %v", name, organization, err)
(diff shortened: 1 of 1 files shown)


sebasslash (approved):
I've pulled your changes locally and have tested they work as intended. Heck yeah 🔥

tfe/resource_tfe_team.go
```diff
@@ -105,8 +105,13 @@ func resourceTFETeamCreate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[DEBUG] Create team %s for organization: %s", name, organization)
 	team, err := tfeClient.Teams.Create(ctx, organization, options)
 	if err != nil {
-		return fmt.Errorf(
-			"Error creating team %s for organization %s: %v", name, organization, err)
+		if err == tfe.ErrResourceNotFound {
+			entitlements, _ := tfeClient.Organizations.Entitlements(ctx, organization)
+			if !entitlements.Teams {
```
brandonc, line 110: a panic might happen here if `entitlements` is nil, so it may be best to check for the error return value here before dereferencing entitlements

It may be interesting to eventually pull entitlements up front so that all resources could short circuit with a helpful error message some day. I didn't know about this API. I'll look into it!

brandonc (changes requested):

tfe/resource_tfe_team.go
sebasslash: It could be something as simple as the API returns a `500` or a `503`. If an organization has zero entitlements, that would be a huge problem and beyond the scope of the provider.

I don't think an explicit test is necessary; simply having a block that handles the error and returns a clear error msg is sufficient.

brandonc: I would just fall through and let the 404 error be returned

Mfhodges: gotcha! that makes a lot of sense. thanks for clarifying that 😊. Just pushed changes for this feedback

brandonc (approved):
Thanks for submitting this!

PR #422, opened 2022-01-25
Org Membership return Username

Xorima:
## Description

- Updates the Org Membership resource and data source to return the
  username associated with the email address

All other team resources require usernames, so not being able to map a
user's email to their username adds complexity when wanting to add users
to a team

## Testing plan

create a minimum viable TF Code:

```
data "tfe_organization_membership" "test" {
  organization  = "ENTER YOUR ORG HERE"
  email = "Enter your email here"
}

output "output" {
    value = data.tfe_organization_membership.test
}
```
You should see your username in the output. 

## External links

```

...
```

### DIFF
--- a/tfe/data_source_organization_membership.go
+++ b/tfe/data_source_organization_membership.go
@@ -2,6 +2,7 @@ package tfe
 
 import (
 	"fmt"
+
 	"github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -22,6 +23,11 @@ func dataSourceTFEOrganizationMembership() *schema.Resource {
 			},
 
 			"user_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"username": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
--- a/tfe/data_source_organization_membership_test.go
+++ b/tfe/data_source_organization_membership_test.go
@@ -25,6 +25,7 @@ func TestAccTFEOrganizationMembershipDataSource_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						"data.tfe_organization_membership.foobar", "organization", orgName),
 					resource.TestCheckResourceAttrSet("data.tfe_organization_membership.foobar", "user_id"),
+					resource.TestCheckResourceAttrSet("data.tfe_organization_membership.foobar", "username"),
 				),
 			},
 		},
--- a/tfe/resource_tfe_organization_membership.go
+++ b/tfe/resource_tfe_organization_membership.go
@@ -34,6 +34,11 @@ func resourceTFEOrganizationMembership() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"username": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -86,6 +91,7 @@ func resourceTFEOrganizationMembershipRead(d *schema.ResourceData, meta interfac
 	d.Set("email", membership.Email)
 	d.Set("organization", membership.Organization.Name)
 	d.Set("user_id", membership.User.ID)
+	d.Set("username", membership.User.Username)
 
 	return nil
 }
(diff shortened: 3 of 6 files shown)


Xorima:
I'm having issues with how to run the tests for this as it wants to use the example@hashicorp.com account however I do not know the username nor do I know if this is mock data or not from how this works. Assistance would be highly appreciated.

brandonc:
This makes sense to me since it is useful for team resources. I smoke tested this and it works fine for both data sources and resources but would you mind amending `tfe/resource_tfe_organization_membership_test.go` and `tfe/data_tfe_organization_membership_test.go` with the new attribute (similar to the existing attribute test for user_id) ? Then I can trigger the test run in a way that our CI system will allow with no further action required. Thank you very much for this!

Xorima:
done @brandonc, do you have any links to documentation around these tests so I can understand them more for future reference?

brandonc:
@Xorima the full test environment is not exactly accessible to you because we use acceptance tests against a running Terraform Cloud test environment with a full feature set. Some tests can be run against a Terraform Enterprise installation, but this requires a ton of setup. You can read about configuring it here https://github.com/hashicorp/terraform-provider-tfe/blob/main/TESTS.md

The CI environment is only accessible to Hashicorp employees, and it won't run when triggered from forked repos. I pushed your branch to this repo which will make the CI environment start reporting status checks in this PR. I'll let you know if anything else is needed. Thanks again for the test changes.

brandonc (approved):

Xorima:
Great, can you let me know when this is released as well as I have a module pending on this 

Thanks for the review

brandonc:
@Xorima unfortunately, your test changes are the the only thing that failed 😅 You just need to add the attributes into the function that generates the test terraform config (in both test files) one is called `testAccTFEOrganizationMembershipDataSourceConfig` and the other `testAccTFEOrganizationMembership_basic` Would you mind doing that? If not, I don't mind taking this over but I'm feeling confident this will make it pass ✅ 

Merging this PR won't release a new version of the provider, but we are actually interested in releasing the next minor version later this week.

Xorima:
I'll aim to get this done tomorrow as it's on my high priority list, expect something ready for you when you login for work :)

Xorima:
Hey @brandonc  could you make the required changes please? I am struggling to understand how to update the config for an attribute that is computed. As the attribute I am setting is similar to `user_id` but I cannot see it in the `testAccTFEOrganizationMembershipDataSourceConfig` I am unsure of the required changes. Would be great if you could explain them as well for the next time I issue a PR

brandonc:
@Xorima I took a closer look at this to fix the test issue and it struck me that the username would be an empty string if the user is created as a result of the organization membership resource being created.

In that case, it would be broken to try and create a team_member or team_members resource using one or more empty strings, which I think is why it's preferred to use the `tfe_team_organization_member` [resource](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/team_organization_member) which accepts an organization_membership_id

I think it's safe to say that others have probably run into this issue in the past for the same reason which is why tfe_team_organization_member exists. Does this resource fit your use case? I'd rather not add confusion by extending the usage of an optional field like username

Xorima:
@brandonc my usecase is the following:

1. Users are added to TFC using sso, so I cannot use the `tfe_organization_membership` resource, only the data source
2. Users can have random usernames but email is a well known property
3. I add them to a team individually, as unfortunately other users could be added manually (not best practice but it is a fact)

To add them to a team the following modules exist:
- `tfe_team_organization_member` which requires a `organization_membership_id` which I cannot find by the data source for org membership if the only known property I have is an email address
- `tfe_team_members` which required a `username`, which I cannot predetermine as these are user controlled instead of emails which are centrally controlled 
- `tfe_team_member` which required a `username`, which I cannot predetermine as these are user controlled instead of emails which are centrally controlled 

The reason for wanting email address so strongly besides who configures them is when a user leaves the company it is easy to search and remove their email from a well known file, the username may or may not be easily identifiable which will cause considerable additional efforts. 

I hope this helps explain the reasoning

brandonc:
@Xorima Yes, I think in this case it would be best to use the tfe_organization_membership data source to fetch the [id attribute](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/data-sources/organization_membership#attributes-reference) using the email argument, which you can plug into a tfe_team_organization_member resource.

Xorima:
Confirmed working via suggestions, closed pr

PR #429, opened 2022-02-03
Add example for Organization Access

Xorima:
Extends the examples to cover adding org access

### DIFF
--- a/website/docs/r/team.html.markdown
+++ b/website/docs/r/team.html.markdown
@@ -21,6 +21,18 @@ resource "tfe_team" "test" {
 }
 ```
 
+Organization Permission usage:
+
+```hcl
+resource "tfe_team" "test" {
+  name         = "my-team-name"
+  organization = "my-org-name"
+    organization_access {
+      manage_vcs_settings = true
+    }
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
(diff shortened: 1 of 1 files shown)


website/docs/r/team.html.markdown
```diff
@@ -21,6 +21,18 @@ resource "tfe_team" "test" {
...
 
+Organization Permission usage:
+
+```hcl
+resource "tfe_team" "test" {
+  name         = "my-team-name"
+  organization = "my-org-name"
+    organization_access {
+      manage_vcs_settings = true
+    }
```
brandonc, line 32: I think this just need a de-intent

```suggestion
  organization_access {
    manage_vcs_settings = true
  }
```

brandonc (approved):
Thanks for improving the docs!

Xorima:
One small thing to note here, we are unable to directly permission the private registry modules & providers as the upstream golang module doesn't support this from what I can tell

Xorima:
So I no longer have access as I just left my previous employer to send you a screenshot, but if on a fully licensed TFC account you go into a team and look at organization permissions in the UI you will see as well something like:
- Manage Registry
  - Manage Modules
  - Manage Providers

This all relates to the private TF registry you get in TFC, it was frustrating that I could not perm this via code itself and has to be manually done on relevant teams.

PR #442, opened 2022-03-07
Migrate the provider to go-tfe v1.0.0

sebasslash:
This PR will be the working branch for migrating the changes as we take go-tfe from pre-release to release!


tfe/data_source_agent_pool.go
```diff
@@ -36,7 +36,7 @@ func dataSourceTFEAgentPoolRead(d *schema.ResourceData, meta interface{}) error
 	options := tfe.AgentPoolListOptions{}
 
 	for {
-		l, err := tfeClient.AgentPools.List(ctx, organization, options)
+		l, err := tfeClient.AgentPools.List(ctx, organization, &options)
```
uturunku1, line 39: Why not pass `nil` in here and other places where ListOptions are not assigning any particular values?

uturunku1: nvm! I just noticed that further in the code we need to assign a value to `options.PageNumber`

PR #445, opened 2022-03-18
implement tfe_oauth_clients datasource

arminaaki:
## Description

Adds the new data source `tfe_oauth_clients` to retrieve a list of TerraformCloud/Enterprise's oauth clients in a given organization. This data source will ultimately allow us to lookup `oauth_client_id`s across different organizations. 
## Testing plan
* Ran acceptance tests locally against my TerraformCloud account.
```
Running tool: /usr/local/opt/go/libexec/bin/go test -timeout 30s -run ^TestAccTFEOAuthClientsDataSource_basic$ github.com/hashicorp/terraform-provider-tfe/tfe

ok  	github.com/hashicorp/terraform-provider-tfe/tfe	5.980s
```

## External links
- https://www.terraform.io/cloud-docs/api-docs/oauth-clients#list-oauth-clients

### DIFF
--- a/tfe/data_source_oauth_clients.go
+++ b/tfe/data_source_oauth_clients.go
@@ -1,0 +1,138 @@
+package tfe
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceTFEOauthClients() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTFEOauthClientList,
+
+		Schema: map[string]*schema.Schema{
+			"organization": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"oauth_clients": {
+				Computed: true,
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeMap},
+			},
+		},
+	}
+}
+
+func dataSourceTFEOauthClientList(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	organization := d.Get("organization").(string)
+
+	log.Printf("[DEBUG] Setting OauthClients Attributes for organization %s", organization)
+	err := oauthClientsPopulateFields(tfeClient, organization, d)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func oauthClientsPopulateFields(client *tfe.Client, organization string, d *schema.ResourceData) error {
+	var oauthClients []map[string]interface{}
+
+	log.Printf("[DEBUG] Listing oauth clients")
+	options := tfe.OAuthClientListOptions{
+		ListOptions: tfe.ListOptions{
+			PageSize: 100,
+		},
+	}
+	for {
+		oauthClientList, err := client.OAuthClients.List(ctx, organization, options)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve oauth Clients from organization %s: %v", organization, err)
+		}
+
+		for _, client := range oauthClientList.Items {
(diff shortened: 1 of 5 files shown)


tfe/data_source_oauth_clients.go
```diff
@@ -0,0 +1,138 @@
...
+
+		// Update the page number to get the next page.
+		options.PageNumber = oauthClientList.NextPage
+	}
+
+	d.SetId("oauth-clients")
+	return d.Set("oauth_clients", oauthClients)
+}
+
+func flattenOauthClient(in *tfe.OAuthClient) (map[string]interface{}, error) {
```
sebasslash, line 82: I'm curious about this one. Is there a particular reason you wish to flatten the `OAuthClient` object? I'm wondering about the implications if a user expects to fetch an attribute that was omitted due to flattening.

sebasslash: Also, this isn't a pattern that I've seen applied to other data sources, so I'm hesitant to introduce this.

tfe/data_source_oauth_clients_test.go
```diff
@@ -0,0 +1,68 @@
...
+						"data.tfe_oauth_clients.foobar", "oauth_clients.0.service_provider", "github"),
+					resource.TestCheckResourceAttr(
+						"data.tfe_oauth_clients.foobar", "oauth_clients.1.service_provider", "github"),
+				),
+			},
+		},
+	})
+}
+
+func testAccTFEOAuthClientsDataSourceConfig(rInt int) string {
```
sebasslash, line 39: I would also write a test/config for when there are no oauth clients to fetch. Tests should capture multiple configuration scenarios other than 'basic' (even though this DS is quite simple), the reason being users have a knack for breaking things! 😆

tfe/data_source_oauth_clients_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccTFEOAuthClientsDataSource_basic(t *testing.T) {
+	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			if GITHUB_TOKEN == "" {
```
sebasslash, line 17: What do you think about moving checking the existence of `GITHUB_TOKEN` to the top of the method body to trigger an early exit and avoid a few method calls?

tfe/data_source_oauth_clients_test.go
```diff
@@ -0,0 +1,68 @@
...
+		Steps: []resource.TestStep{
+			{Destroy: false,
+				PreventPostDestroyRefresh: true,
+				Config:                    testAccTFEOAuthClientsDataSourceConfig(rInt),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"data.tfe_oauth_clients.foobar", "oauth_clients.#", "2"),
+					resource.TestCheckResourceAttr(
+						"data.tfe_oauth_clients.foobar", "oauth_clients.0.service_provider", "github"),
+					resource.TestCheckResourceAttr(
```
sebasslash, line 31: We should probably sanity check a few more attributes!

sebasslash (changes requested):
Thank you for your contribution, Armin! 👍 Some few things to mention...

tfe/data_source_oauth_clients_test.go
sebasslash: Alternatively, we've typically have `full` and `basic` acceptance tests, where `TestAcc*_full` validates all attributes and `TestAcc*_basic` validates *required* attributes, or in this case attributes that must be present.

barrettclark:
I've market this PR as `stale` since it hasn't had a response in 4 months. I'll close it in a week unless there are additional comments or changes posted.

PR #451, opened 2022-03-24
Changelog for v0.29.0 (unreleased)

sebasslash:
The changelog! 🕺

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.29.0 (Unreleased)
+
+BUG FIXES:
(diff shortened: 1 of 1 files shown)


laurenolivia (commented):
Should "Update go-tfe dependency to version ..." read "r/team: Update go-tfe dependency to version..." ? Referencing how you noted the other changes.

sebasslash:
When you see `r/some_resource` or `d/some_data_source` prefix in the changelog it means a change specific to that item. Since the go-tfe upgrade is a dependency change, and not specific to any resource/datasource, it's not prefixed. We typically don't mention dependency upgrades in the changelog, but considering this dependency is the backbone of this provider we want to communicate the API client is now stable.

laurenolivia (approved):
Thanks for explaining.

Issue #391, opened 2021-11-11
Supporting Variable Sets


mbevc1:
#### Use-cases

New feature released: https://www.hashicorp.com/blog/terraform-cloud-variable-sets-beta-now-available

#### Attempted Solutions

/

#### Proposal

It seems it's already part of API: https://www.terraform.io/docs/cloud/api/variable-sets.html
I recognize it's still in beta, but it would be nice have a resource in TFE for configuring it.

madtank:
Thanks for opening this feature request. What we would need is ability to add variable sets to a workspace. Creating and updating the variable sets would be nice option too.

elocke:
Thanks for submitting this, variable sets solve a problem very elegantly for us but cannot be used easily without the provider supporting them.

dkirrane:
Any ETA for this?

madtank:
> Any ETA for this?

I was told April 2022 timeframe.

rorychatterton:
A couple of items that would be useful for customers like myself when implementing this provider:

### Seperate Set from Assignment objects
Can you please ensure that we have separate objects to handle the creation of the variable set, away from the assignment of that variable set. For example:

_Not this:_
```
resource "tfe_variable_set" "my_set" {
   tfe_workspaces = [...]
}
```

But rather this:
```
resource "tfe_variable_set" "my_set" {}
resource "tfe_variable_set_assignment" "my_set_assignment_ws_1" {
    variable_set = ""
    tfe_workspaces = ""
}

```

We would like this to be able to decouple the creation of global variable sets, from the assignment to workspaces. This particular pain point exists with the policy workspace, and it would be great if we could avoid it here!

### Handle Complex Terraform Objects
The current terraform variable class doesn't handle complex objects. If you want to output a complex object into a terraform managed variable (such as a map of objects), you must convert that object into a string (we normally use json encode). 

This limits the usefulness of the variable provider as you are not able to do type validation on the consuming workspace side. 

I have raised an [issue](https://github.com/hashicorp/terraform-provider-tfe/issues/433) here for resolution in the context of standard variables.

Satak:
This would be very useful 🤞

ryanisnan:
Plus one on this feature!

morremeyer:
That's what a thumbs up reaction is for.

In the future, please don't spam people with notifications just to express your support of an idea.

mbevc1:
[...]

PR #452, opened 2022-03-24
Manage Variable Sets

rexredinger:
## Description

Add support for variable sets

TODO: Get go-tfe [updates](https://github.com/hashicorp/go-tfe/pull/305) in a tagged release for reference.

## Testing plan

1.  Run specs
2. Try to provision a global varset
3. Try to provision a varset assigned to particular workspaces
4. Add variables
5. Read the variable sets from datasources
6. destroy everything

## External links

- [API documentation](https://www.terraform.io/cloud-docs/api-docs/variable-sets)
- [Related PR](https://github.com/hashicorp/go-tfe/pull/305)

```
🍔_alex: terraform-provider-tfe $ TESTARGS="-run TestAccTFEVariable" envchain tfe_provider make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEVariable -timeout 15m
=== RUN   TestAccTFEVariableSetsDataSource_basic
--- PASS: TestAccTFEVariableSetsDataSource_basic (8.39s)
=== RUN   TestAccTFEVariablesDataSource_basic
--- PASS: TestAccTFEVariablesDataSource_basic (7.69s)
=== RUN   TestAccTFEVariableSet_basic
--- PASS: TestAccTFEVariableSet_basic (6.55s)
=== RUN   TestAccTFEVariableSet_update
--- PASS: TestAccTFEVariableSet_update (10.49s)
=== RUN   TestAccTFEVariableSet_import
--- PASS: TestAccTFEVariableSet_import (7.06s)
--- PASS: TestAccTFEVariable_basic (6.55s)
--- PASS: TestAccTFEVariable_basic_variable_set (6.36s)
--- PASS: TestAccTFEVariable_update (12.05s)
--- PASS: TestAccTFEVariable_update_key_sensitive (11.97s)
--- PASS: TestAccTFEVariable_import (7.53s)
PASS
PASS	github.com/hashicorp/terraform-provider-tfe/tfe	85.186s
PASS
...
```


tfe/resource_tfe_variable_set_test.go
```diff
@@ -0,0 +1,264 @@
...
+  name         = "variable_set_test"
+	description  = "a test variable set"
+	global       = false
+	organization = tfe_organization.foobar.id
+}
+
+resource "tfe_variable_set" "applied" {
+  name         = "variable_set_applied"
+	description  = "a test variable set"
+	workspaces   = [tfe_workspace.foobar.id]
```
sebasslash, line 234: Isn't workspaces a computed attribute? We should omit computed attributes from our configuration, what do you think?

website/docs/d/variable_sets.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_variable_set"
+sidebar_current: "docs-datasource-tfe-variable-set"
+description: |-
+  Get information on organization variable sets.
+---
+
+# Data Source: tfe_variable_set
+
+This data source is used to retrieve all variables defined in a specified workspace
```
sebasslash, line 11: Missing period. 😆 (sorry)

tfe/resource_tfe_variable_set.go
```diff
@@ -0,0 +1,181 @@
...
+				ConflictsWith: []string{"workspaces"},
+			},
+
+			"organization": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"workspaces": {
```
sebasslash, line 48: Rename to `workspace_ids` ?

website/docs/r/variable_set.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+data "tfe_workspace_ids" "prod-apps" {
+  tag_names    = ["prod", "app", "aws"]
+  organization = tfe_organization.test.id
+}
+
+resource "tfe_variable_set" "test" {
+  name         = "Tag Based Varset"
+  description  = "Variable set applied to workspaces based on tag."
+  organization = tfe_organization.test.id
+  workspaces   = tfe_workspace_ids.prod-apps.ids
```
sebasslash, line 102: Ditto on computed attributes showing up in the configuration (we should only have arguments)

sebasslash: I may be wrong here and computed attributes that are also arguments are totally valid.

sebasslash (changes requested):
_Phew_, ok... this PR is a juicy one and I will definitely need to give it a second pass (and have a second pair of eyes). For now, I simply wanted to give back enough to chew on while I thoroughly re-review. 

All in all, awesome work and thank you so much for this! 🔥 Feel free to debate any of my points as I may have left some feedback without the proper context.

tfe/data_source_variable_set_test.go
```diff
@@ -0,0 +1,73 @@
...
+					resource.TestCheckResourceAttrSet("data.tfe_variable_set.foobar", "id"),
+					resource.TestCheckResourceAttr(
+						"data.tfe_variable_set.foobar", "name", fmt.Sprintf("varset-foo-%d", rInt)),
+					resource.TestCheckResourceAttr(
+						"data.tfe_variable_set.foobar", "description", "a description"),
+					resource.TestCheckResourceAttr(
+						"data.tfe_variable_set.foobar", "global", "false"),
+					resource.TestCheckResourceAttr(
+						"data.tfe_variable_set.foobar", "organization", orgName),
+					resource.TestCheckResourceAttr(
```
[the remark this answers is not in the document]
rexredinger: Since the ids are determined at run time how can I reference them from inside the test func declaration? Looking for prior art and coming up short.

tfe/data_source_variable_set.go
```diff
@@ -0,0 +1,116 @@
...
+				d.Set("name", vs.Name)
+				d.Set("description", vs.Description)
+				d.Set("global", vs.Global)
+
+				//Only now include vars and workspaces to cut down on request load.
+				readOptions := tfe.VariableSetReadOptions{
+					Include: &[]tfe.VariableSetIncludeOpt{tfe.VariableSetWorkspaces, tfe.VariableSetVars},
+				}
+
+				vs, err = tfeClient.VariableSets.Read(ctx, vs.ID, &readOptions)
```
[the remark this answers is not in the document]
rexredinger: My thinking here was to not put the extra load of looking up and serializing all the vars and workspaces until we know which one we care about. There are realistic scenarios where I believe the List call could become quite slow if asked to collect all that extra data. Happy to chat more about it if that seems unfounded.

tfe/data_source_variables_test.go
```diff
@@ -57,23 +66,57 @@ resource "tfe_variable" "envbar" {
...
 		tfe_variable.envbar
   ]
 }
 
-output "variables" {
-	value = data.tfe_variables.foobar.variables[0]["name"]
+data "tfe_variables" "variable_set_foobar" {
+	variable_set_id = tfe_variable_set.foobar.id
+	depends_on = [
+    tfe_variable.terrfoo,
```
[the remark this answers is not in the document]
rexredinger: Agreed 😸
It does not look like that on my local machine 🤔

tfe/resource_tfe_variable_set.go
rexredinger: The thinking here is that other references, like `organization`, also take an identifier as an argument, but the label is for the whole object(s) being referenced. But I see other code using worksace_ids, so I will update this to match.

tfe/data_source_variables_test.go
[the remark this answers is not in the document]
rexredinger: sneaky autoformatted tabs

tfe/data_source_variables.go
```diff
@@ -62,14 +62,26 @@ func dataSourceTFEWorkspaceVariables() *schema.Resource {
 				},
 			},
 			"workspace_id": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				ExactlyOneOf: []string{"workspace_id", "variable_set_id"},
```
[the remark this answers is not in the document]
rexredinger: Do you recommend this for the `resource*` definitions as well?

tfe/resource_tfe_variable.go
```diff
@@ -107,13 +121,19 @@ func forceRecreateResourceIf() schema.CustomizeDiffFunc {
...
+
 	tfeClient := meta.(*tfe.Client)
 
 	// Get key and category.
 	key := d.Get("key").(string)
 	category := d.Get("category").(string)
 
-	// Get the workspace
+	// Get the workspace if workspace_id present
 	workspaceID := d.Get("workspace_id").(string)
```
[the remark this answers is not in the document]
rexredinger: If this code gets exercised, `workspace_id` is guaranteed to be present, by way of schema constraints. The update to this comment is confusing. I will revert it.

rorychatterton:
Referencing my post in this thread: https://github.com/hashicorp/terraform-provider-tfe/issues/391#issuecomment-1036929310

By coupling the variable set creation logic, with the assignment logic, this resource becomes much harder to consume.

This means that you have to create the variable set and assign it to child workspaces, all in the same workspace. 

Is it possible to further split `tfe_variable_set` resource further to decouple the logic?

tfe/resource_tfe_variable_set.go
```diff
@@ -0,0 +1,181 @@
...
+				ConflictsWith: []string{"workspace_ids"},
+			},
+
+			"organization": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"workspace_ids": {
```
rorychatterton, line 48: Referencing my post in this thread: https://github.com/hashicorp/terraform-provider-tfe/issues/391#issuecomment-1036929310

By coupling the variable set creation logic, with the assignment logic, this resource becomes much harder to consume.

This means that you have to create the variable set and assign it to child workspaces, all in the same workspace.

Is it possible to further split tfe_variable_set resource further to decouple the logic?

sebasslash: This makes sense to me, I definitely think the `workspace_ids` attribute being both a computed property and an argument can make it harder to reason about your configuration, especially if that variable set is attached to a workspace outside your configuration. Your config might look like:

```hcl
resource "tfe_variable_set" "foo" {
  # other attributes here
  workspace_ids = ["ws-ID1"]
}
```
and your state file: 
```json
"workspace_ids": ["ws-ID1", "ws-ID2", "ws-ID3"] 
```

_(which I think it's weird 'cause it creates a disjunction between intent and reality)_ 

Expressing relationships as resources has been done before, the examples I can think of are: `tfe_organization_module_sharing` and `tfe_organization_membership`. Doing so for variable sets, we can at least be explicit about what VS->Wkspace relationships are managed by Terraform, while keeping `workspace_ids` as the source of truth for _all_ relationships (including those not managed by Terraform).

sebasslash (approved):
I'm going to hit approve on this one 🥷 . While there is an interesting discussion to be had regarding how best to represent object relationships, the team implementing this has notified me there have already been internal discussions regarding this and have decided a change like this is beyond scope.

Truthfully, this is part of a larger underlying issue with provider (and not just this resource) and should be addressed in future releases. I apologize to those wanting the decoupling!

tfe/resource_tfe_variable_set.go
rorychatterton: Appreciate this PR is closed. Now that you have explained it, I can understand why that could be difficult to work with.

The Vault Provider deals with this with the use of an `explicit = true` and `external_policies = true` flag, as seen in these two resources:
[identity_group_policies](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/identity_group_policies)
[identity_group_policies](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/identity_group)

This tells Terraform to ignore any items that weren't previously managed under state to avoid the disfunction you spoke of earlier. I haven't looked closely enough at the provider code to understand how that works under the hood. 

You mentioned you want to decouple the conversation. Please advise where you would like to track that going forth if looking for further input :)

PR #466, opened 2022-03-31
Update Variable & Variable Sets Documentation

zahiar:
## Description

Fixed incorrect workspace resource reference, should be `name` not `id`. And added docs to show how to import variable sets and variables that are part of a variable set.

PR #469, opened 2022-04-01
Update changelog to v0.30.2 (unreleased)

sebasslash:


uturunku1 (approved):
🤞

zahiar:
@sebasslash Looks to be missing a fair few changes from #466 like the import sections on both variable & variable_set resources 😢

zahiar:
> @sebasslash Looks to be missing a fair few changes from #466 like the import sections on both variable & variable_set resources 😢

Actually, looking at previous Changelog PRs this is fine, I got caught out by the additional `organization = tfe_organization.test.name` 🙈

sebasslash:
No worries, your changes are there! This was simply to verify the changelog was correctly worded. I noticed a few organization attributes were being referenced by `id` instead of `name` so I changed them. Although in practice `name` and `id` are interchangeable for organizations.

zahiar:
Cool, we actually found the opposite in our case, `id` & `name` were not interchangeable - it led to errors as `id` was some generated value.

sebasslash:
That's very interesting and I'll keep that in mind for our future effort to bring the provider to 1.0.

PR #471, opened 2022-04-06
Migrate the release process for the provider

sebasslash:


terraform-registry-manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "version": 1,
+    "metadata": {
+        "protocol_versions": [
+            "6.0"
```
bflad, line 5: You'll want 5.0 for now, since this provider is being developed using terraform-plugin-sdk/v2. Otherwise, this provider will not show up for Terraform CLI 0.12-0.15 discovery. 👍 

```suggestion
            "5.0"
```

bflad (approved):
Two things for consideration, otherwise looks good to me 🚀

barrettclark (approved):

PR #473, opened 2022-04-07
Fix example creating variable set that applies workspace_ids data

marcel-dias:
## Description

Fix example creating a variable set that applies workspace_ids data

`data.tfe_workspace_ids.this.ids`  returns a `map[WS name] = ws-id` 
so to use that as input for `tfe_variable_set.workspace_ids` only values are needed 

## Testing plan

documentation update

Documentation website should be updated and new example showed


sebasslash (approved):
Great catch 🔥

PR #477, opened 2022-04-15
allow TF_TOKEN_... variables to specify host-specific credentials

brandonc:
## Description

This is a companion PR to a change to terraform that adds host specific env svc creds. It should be possible for tfe to use the same method when specifying credentials for a particular tfc host.

## Testing plan

0. Build this code `go build -o terraform-provider-tfe` and update your CLI config to override it (see below)
1. Don't specify a token in your tfe provider configuration block
2. Remove or modify your CLI configuration credentials for the test host
3. Run terraform using an env var  with the correct token value (see below)
4. This should work on Terraform 1.1 with local state. It would also work in latest Terraform built from source

I used this config:

```
provider "tfe" {
  hostname = "YOURHOST.com"
}

resource "tfe_organization" "test-organization" {
  name  = "my-org-name"
  email = "bcroft@hashicorp.com"
}

  tag_names    = ["test", "app"]
}
```

with this command:

`TF_CLI_CONFIG_FILE=.terraformrc TF_TOKEN_YOURHOST_COM="MY_REAL_TOKEN" terraform apply`

with this CLI config:

```
provider_installation {
  # Use /home/developer/tmp/terraform-null as an overridden package directory
  # for the hashicorp/null provider. This disables the version and checksum
  # verifications for this provider and forces Terraform to look for the
  # null provider plugin in the given directory.
  dev_overrides {
    "hashicorp/tfe" = "/Users/brandonc/hashicorp/terraform-provider-tfe"
  }

  direct {}
}
```

## External links

- [Related PR](https://github.com/hashicorp/terraform/pull/30878)

### DIFF
--- a/tfe/credentials.go
+++ b/tfe/credentials.go
@@ -1,0 +1,101 @@
+package tfe
+
+import (
+	"log"
+	"os"
+	"strings"
+
+	svchost "github.com/hashicorp/terraform-svchost"
+	"github.com/hashicorp/terraform-svchost/disco"
+)
+
+func collectCredentialsFromEnv() map[svchost.Hostname]string {
+	const prefix = "TF_TOKEN_"
+
+	ret := make(map[svchost.Hostname]string)
+	for _, ev := range os.Environ() {
+		eqIdx := strings.Index(ev, "=")
+		if eqIdx < 0 {
+			continue
+		}
+		name := ev[:eqIdx]
+		value := ev[eqIdx+1:]
+		if !strings.HasPrefix(name, prefix) {
+			continue
+		}
+		rawHost := name[len(prefix):]
+
+		// We accept double underscores in place of hyphens because hyphens are not valid
+		// identifiers in most shells and are therefore hard to set.
+		// This is unambiguous with replacing single underscores below because
+		// hyphens are not allowed at the beginning or end of a label and therefore
+		// odd numbers of underscores will not appear together in a valid variable name.
+		rawHost = strings.ReplaceAll(rawHost, "__", "-")
+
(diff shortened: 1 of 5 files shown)


annawinkler (approved):
🌟 Tested and verified locally and it works!

tfe/credentials.go
```diff
@@ -0,0 +1,101 @@
...
+	svchost "github.com/hashicorp/terraform-svchost"
+	"github.com/hashicorp/terraform-svchost/disco"
+)
+
+func collectCredentialsFromEnv() map[svchost.Hostname]string {
+	const prefix = "TF_TOKEN_"
+
+	ret := make(map[svchost.Hostname]string)
+	for _, ev := range os.Environ() {
+		eqIdx := strings.Index(ev, "=")
```
sebasslash, line 17: Just curious here on readability: Would it be better to `strings.Split()` on `=` and compare each index rather than using ranges ?

sebasslash (changes requested):
Code looks fuego 🔥 👍  -- Last thing would be to update the `Authentication` section in this file: https://github.com/hashicorp/terraform-provider-tfe/blob/main/website/docs/index.html.markdown

tfe/credentials.go
brandonc: Split might be slightly more readable, but this version doesn't do any array allocation and is shared with terraform already. I think I'd like to keep it.

brandonc:
@sebasslash I added the docs. Good call

sebasslash (approved):
LGTM 👍

PR #478, opened 2022-04-19
changes to support releases-api launch

modrake:
This reflects the changes that accompany the releases-api anticipated to be launched in early May. Included is:

 *  Version bump of the terraform provider release gh action
 *  Variables added for interfacing with releases-api
 *  hc-releases syntax modified for releases-api use
 *  metadata file added for ingesting with releases-api


sebasslash:
We do have some changes awaiting to be released, would we need to hold off on creating new releases until the new releases-api is launched?

modrake:
> We do have some changes awaiting to be released, would we need to hold off on creating new releases until the new releases-api is launched?

No, this change will not be merged until after the launch so y'all can stay on your current release tooling. We have dozens of repos to prep for launch day so we are doing these ahead of time so that we can quickly merge and get folks back on track after the launch.

jeanneryan (approved):

PR #487, opened 2022-05-05
Add run_tasks permission to Team Access

glennsarti:
## Description

Run Tasks has added a new `run_tasks` custom permission to Team
Workspace access.  This commit adds this new permission to the
Team Access object (both resource and data) and adds acceptance
tests for both objects.

## Testing plan

N/A  Added acceptance tests

## External links

- [go-tfe documentation](https://github.com/hashicorp/go-tfe/blob/main/team_access.go#L101)

N/A It'll be in the CI tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 BREAKING CHANGES:
 * **Removed Authentication Method**: Host-specific TF_TOKEN_... environment variable can no longer be used for token authentication. This method of authentication is incompatible with the Terraform Cloud remote execution model. Please use the TFE_TOKEN environment variable.
+* r/tfe_workspace: Default value of the `file_triggers_enabled` field is changed to `false`. This will align the
+  `file_triggers_enabled` field default value with the default value for the same field in the
+  [TFC API](https://www.terraform.io/cloud-docs/api-docs/workspaces).
+  If the value of the `file_triggers_enabled` field was not explicitly set and either of the fields `working_directory`
+  (not an empty string) or `trigger_prefixes` was used - to keep the behavior unchanged, the `file_trigger_enabled`
+  field should now explicitly be set to `true`. ([#510](https://github.com/hashicorp/terraform-provider-tfe/pull/510/files))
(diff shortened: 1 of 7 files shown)


glennsarti:
This is a breaking change because custom access permissions will _require_ the run_tasks parameter

e.g.

```
resource "tfe_team_access" "foobar" {
  permissions {
    runs = "apply"
    variables = "read"
    state_versions = "read-outputs"
    sentinel_mocks = "none"
    workspace_locking = false
    run_tasks = false           <------ THIS IS NOW REQUIRED
  }
 ```

sebasslash (changes requested):
Code looks good, missing the provider documentation + Wondering how our docs should reflect this breaking change? Would a simple `-> **Note**` suffice?

glennsarti:
@sebasslash   I've updated the docs and changelog. I have no idea how to surface the breaking changen-ess besides bumping a major version number.

brandonc:
@glennsarti I think we reserve the right to break the API in reasonable ways until release 1.0. The best thing to do is just note it in CHANGELOG, as you've done.

I've already forgotten why we couldn't default this setting, but I'm wondering how to avoid similar breaking changes after 1.0...

brandonc (approved):

glennsarti:
For ref - Private chat about why it is - https://hashicorp.slack.com/archives/C01BR9FR8KZ/p1651525482312299

> The problem therein lies: access and permissions behave both as arguments and computed properties, but their behavior depends on whether the other is marked as an argument or as a computed property. The reason for this weird behavior is due to the [Team Access API](https://www.terraform.io/cloud-docs/api-docs/team-access#add-team-access-to-a-workspace): you cannot set custom permissions unless the access field is set to custom so we can't treat both attributes as arguments because if permissions is set, access is computed to "custom" and if access is set, permissions is computed to whatever values are returned by the API.
Setting default values however breaks this behavior because the intent is muddled, are we using access to set the permissions or the default values of permissions ?

glennsarti:
@sebasslash Re-requesting a review based on the changes I made.

brandonc (approved):
@sebasslash is out of the office but I think we are good to go. Thanks for adding context about the required permission arguments!

Issue #326, opened 2021-06-07
Excessive API calls from tfe_team and tfe_organization_membership data sources


ericnorris:
#### Terraform Cloud/Enterprise version

Terraform Cloud

#### Terraform version

```plaintext
Terraform v0.15.5
on darwin_amd64
```

#### Terraform Configuration Files

```terraform
data "tfe_team" "team_lookup" {
  for_each = to_set(var.some_list)

  organization = "some-organization"
  name         = each.value
}

data "tfe_organization_membership" "membership_lookup" {
  for_each = to_set(var.some_other_list)

  organization = "some-organization"
  email        = each.value
}

```

#### Debug Output

N/A

#### Expected Behavior

`tfe_team` and `tfe_organization_membership` data sources won't potentially cause excessive API lookups by paginating through all of the Teams and Organization Memberships in the organization.

#### Actual Behavior

Each `tfe_team` and `tfe_organization_membership`  data source paginates through potentially all Teams and Organization Memberships in the organization, which could lead to performance issues or potentially throttling on the API side.

#### Additional Context

We're using the `tfe` provider heavily to configure our Terraform Cloud teams and their access to workspaces. We will likely have hundreds, and potentially thousands, of `tfe_organization_membership` and `tfe_team` data source lookups - since each one appears to list _every_ organization member and team (albeit exiting early when it finds a match), we may end up issuing tens of thousands of API calls in a short burst when executing a plan.

While we haven't seen issues yet as we're still on-boarding folks to our organization, but we're concerned with the risk of causing excessive API lookups and slowing down our Terraform Cloud runs. I would be open to submitting a PR that adds some sort of caching, but wanted to raise an issue first to get your opinion.

Alternative solutions could be changing the TFE API to allow for `GET` requests for a particular team, rather than having to list every team, and similar for the `tfe_organization_membership` data source.

chrisarcand:
Indeed, this is a very valid request and thank you for raising it. I do think the most appropriate answer may involve API changes as you mentioned.

brandonc:
I think we forgot to close this back in May. Thanks for reporting it.

PR #491, opened 2022-05-06
Replace brute search on `d/tfe_team` and `d/tfe_organization_membership` 

sebasslash:
## Description

Previously, the [List Organization Members](https://www.terraform.io/cloud-docs/api-docs/organization-memberships#list-memberships-for-an-organization) and [List Teams](https://www.terraform.io/cloud-docs/api-docs/teams#list-teams) API did not support any way to retrieve the desired object(s) through the use of query params. Therefore, the current implementation of `d/tfe_team` and `d/tfe_organization_membership` relies on brute searching through the paginated API until a match was found. 

This PR makes use of the new `filter[email]` (for org members) and `filter[names]` (for teams) to reduce the number of API calls to fetch the desired resource from N to 1. This resolves issue #326 

## Testing plan

1.  In order to run the tests:
```sh
TESTARGS="-run TestAccTFEOrganizationMembershipDataSource_basic" make testacc
TESTARGS="-run TestAccTFETeamDataSource_basic" make testacc
```

## External links

- [go-tfe changes](https://github.com/hashicorp/go-tfe/pull/393)

---
  - https://app.asana.com/0/0/1202192449507534


tfe/data_source_organization_membership.go
```diff
@@ -40,29 +40,18 @@ func dataSourceTFEOrganizationMembershipRead(d *schema.ResourceData, meta interf
...
+	}
 
-		// Update the page number to get the next page.
-		options.PageNumber = organizationMembershipsList.NextPage
+	if len(oml.Items) == 0 {
+		return fmt.Errorf("Could not find organization membership for organization %s and email %s", organization, email)
 	}
 
-	return fmt.Errorf("Could not find organization membership for organization %s and email %s", organization, email)
+	d.SetId(oml.Items[0].ID)
```
brandonc, line 55: What if this is used against a TFE version before this filter did anything? I think the brute search needs to be a fallback?

tfe/data_source_team.go
```diff
@@ -36,30 +36,43 @@ func dataSourceTFETeamRead(d *schema.ResourceData, meta interface{}) error {
...
-	for {
-		l, err := tfeClient.Teams.List(ctx, organization, options)
-		if err != nil {
-			return fmt.Errorf("Error retrieving teams: %v", err)
-		}
+	switch len(tl.Items) {
+	case 0:
+		return fmt.Errorf("Could not find team %s/%s", organization, name)
+	case 1:
+		d.SetId(tl.Items[0].ID)
```
brandonc, line 50: similar problem as above

brandonc (changes requested):

tfe/data_source_organization_membership.go
```diff
@@ -40,28 +40,44 @@ func dataSourceTFEOrganizationMembershipRead(d *schema.ResourceData, meta interf
 	// Create an options struct.
 	options := &tfe.OrganizationMembershipListOptions{
 		Include: []tfe.OrgMembershipIncludeOpt{tfe.OrgMembershipUser},
+		Emails:  []string{email},
```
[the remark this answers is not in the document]
sebasslash: Yes and it would behave as expected, i.e letting the user know that the org user with an empty email does not exist -- either by returning zero org members or by brute searching against all org members and checking if `orgMember.Email == ""` (obviously will never return a match)

go.sum
```diff
@@ -210,12 +210,18 @@ github.com/hashicorp/go-plugin v1.4.1 h1:6UltRQlLN9iZO513VveELp5xyaFxVD2+1OVylE+
 github.com/hashicorp/go-plugin v1.4.1/go.mod h1:5fGEH17QVwTTcR0zV7yhDPLLmFX9YSZ38b18Udy6vYQ=
 github.com/hashicorp/go-retryablehttp v0.7.0 h1:eu1EI/mbirUgP5C8hVsTNaGZreBDlYiwC1FZWkvQPQ4=
 github.com/hashicorp/go-retryablehttp v0.7.0/go.mod h1:vAew36LZh98gCBJNLH42IQ1ER/9wtLZZ8meHqQvEYWY=
+github.com/hashicorp/go-retryablehttp v0.7.1 h1:sUiuQAnLlbvmExtFQs72iFW/HXeUn8Z1aJLQ4LJJbTQ=
+github.com/hashicorp/go-retryablehttp v0.7.1/go.mod h1:vAew36LZh98gCBJNLH42IQ1ER/9wtLZZ8meHqQvEYWY=
 github.com/hashicorp/go-safetemp v1.0.0 h1:2HR189eFNrjHQyENnQMMpCiBAsRxzbTMIgBhEyExpmo=
 github.com/hashicorp/go-safetemp v1.0.0/go.mod h1:oaerMy3BhqiTbVye6QuFhFtIceqFoDHxNAB65b+Rj1I=
 github.com/hashicorp/go-slug v0.8.0 h1:h7AGtXVAI/cJ/Wwa/JQQaftQnWQmZbAzkzgZeZVVmLw=
 github.com/hashicorp/go-slug v0.8.0/go.mod h1:Ib+IWBYfEfJGI1ZyXMGNbu2BU+aa3Dzu41RKLH301v4=
+github.com/hashicorp/go-slug v0.8.1 h1:srN7ivgAjHfZddYY1DjBaihRCFy20+vCcOrlx1O2AfE=
```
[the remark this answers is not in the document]
sebasslash: Done 👍

tfe/data_source_organization_membership.go
```diff
@@ -40,28 +40,48 @@ func dataSourceTFEOrganizationMembershipRead(d *schema.ResourceData, meta interf
...
-		for _, organizationMembership := range organizationMembershipsList.Items {
-			if organizationMembership.User.Email == email {
-				d.SetId(organizationMembership.ID)
-				return resourceTFEOrganizationMembershipRead(d, meta)
-			}
+	switch len(oml.Items) {
+	case 0:
+		return fmt.Errorf("Could not find organization membership for organization %s and email %s", organization, email)
+	case 1:
+		if oml.Items[0].User.Email != email {
```
brandonc, line 55: Just one suggestion... this case (and the team case) has become odd but you and I both know it's totally reasonable. Maybe just comment about how this and the default case are to support versions of TFE that do not support the email filter option?

brandonc (approved):

PR #497, opened 2022-05-14
Add keepers argument for token resources

Eric-Jckson:
## Description

Add support to recreate tokens with an argument named keepers. Adding this functionality to recreate token if resources change/update or can be used with the terraform resource time_rotating.

## Testing plan
Example: This will rotate the token after 2 minutes.
```
resource "time_rotating" "token" {
  rotation_minutes = 2
}

resource "tfe_team_token" "manage_workspaces" {
  team_id = "xxxxxx"

  keepers = {
    time_rotation = time_rotating.token.id
  }
}
```
Screen Shot 2022-05-13 at 8 56 05 PM

Currently do not have a test organization for TFC setup. 

Please let me know if any information is missing or if changes need to me made.

### DIFF
--- a/tfe/resource_tfe_agent_token.go
+++ b/tfe/resource_tfe_agent_token.go
@@ -30,6 +30,12 @@ func resourceTFEAgentToken() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+			"keepers": {
+				Description: "Arbitrary map of values that, when changed, will trigger recreation of resource.",
+				Type:        schema.TypeMap,
+				Optional:    true,
+				ForceNew:    true,
+			},
 		},
 	}
 }
--- a/tfe/resource_tfe_organization_token.go
+++ b/tfe/resource_tfe_organization_token.go
@@ -35,6 +35,13 @@ func resourceTFEOrganizationToken() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+
+			"keepers": {
+				Description: "Arbitrary map of values that, when changed, will trigger recreation of resource.",
+				Type:        schema.TypeMap,
+				Optional:    true,
+				ForceNew:    true,
+			},
 		},
 	}
 }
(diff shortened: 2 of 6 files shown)


Eric-Jckson:
Hey @uturunku1 !

**Use Case**
Ability to rotate tokens with terraform based on a time duration. If we have a token that needs to be rotated every 30 days then we can set the resource "time_rotating" to recreate the token resource at the end of that time period.

To achieve this same operation today we would need to do a terraform target destroy on just the token resource.


**Possible Solutions**
I believe that this new terraform option below would be able to achieve the same results that i was planning in the PR.
`replace_triggered_by` is a new lifecycle argument for managed resources which triggers replacement of an object based on changes to an upstream dependency.

uturunku1:
> Hey @uturunku1 !
> 
> **Use Case** Ability to rotate tokens with terraform based on a time duration. If we have a token that needs to be rotated every 30 days then we can set the resource "time_rotating" to recreate the token resource at the end of that time period.
> 
> To achieve this same operation today we would need to do a terraform target destroy on just the token resource.
> 
> **Possible Solutions** I believe that this new terraform option below would be able to achieve the same results that i was planning in the PR. `replace_triggered_by` is a new lifecycle argument for managed resources which triggers replacement of an object based on changes to an upstream dependency.

Thanks for the quick response. `replace_triggered_by` is exactly the option we also thought it could serve your needs! https://www.terraform.io/language/meta-arguments/lifecycle
You would have to upgrade terraform 1.2 and then do something similar to this:

```
resource tfe_team_token "some_team" {
  # ...some tf code
  lifecycle {
    replace_triggered_by = [time_rotating.token.id]
  }
}
```

Are you able to upgrade to 1.2 or is this a limitation for you?

brandonc:
Closing this PR because it's over 6 months old. Let us know if lifecycle replace_triggered_by didn't help your situation!

Issue #492, opened 2022-05-06
`tfe_workspace` state is updated even when the API call to update it fails


ericnorris:
#### Terraform Cloud/Enterprise version

Terraform Cloud

#### Terraform version

```plaintext
Terraform v1.1.9
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v4.17.0
+ provider registry.terraform.io/hashicorp/googleworkspace v0.6.0
+ provider registry.terraform.io/hashicorp/random v3.1.2
+ provider registry.terraform.io/hashicorp/tfe v0.30.2
```

#### Terraform Configuration Files

```terraform
resource "tfe_workspace" "workspace" {
  # ...
  vcs_repo {
    identifier     = var.vcs_repo.identifier
    branch         = local.computed_vcs_branch
    oauth_token_id = var.vcs_repo.oauth_token_id
  }
}
```

#### Debug Output

I don't have debug output, and it'd be difficult to get considering the size of our workspace.

#### Expected Behavior

Updates to the `branch` attribute should not persist in the Terraform State if the API call to update the `branch` attribute errored.

#### Actual Behavior

The `branch` attribute was persisted in the Terraform State after the run completed even though the API call to update the `branch` attribute failed.

#### Additional Context

We received the following error:

```
Error: Error updating workspace ws-REDACTED: invalid attribute Branch doesn't exist
with module.workspace_factory["REDACTED"].tfe_workspace.workspace["REDACTED"]
on ../../modules/tfcloud-workspace/main.tf line 13, in resource "tfe_workspace" "workspace":
```

...but on a subsequent run (with `refresh=false`), no change was detected. When looking at the Terraform state, it appears that the state file was updated even though the run errored out.

Obviously if we ran with `refresh=true` this would have worked. That said, I still think it's a bug that the state is updated when the API call fails.

sebasslash:
Closed. Solved with #498

PR #498, opened 2022-05-16
Use partial flag to prevent updating state on api failure `tfe_workspace.vcs_repo`

sebasslash:
## Description

This quick fix stems from #492 where arguments in `vcs_repo` where being written to state even when the API call failed. Typically this behavior is abnormal, as Terraform will refresh each resource on each operation (plan, apply). The SDKv2 library provides a handy function `(d *ResourceData).Partial(on bool)`:

_Partial is a legacy function that was used for capturing state of specific attributes if an update only partially worked. Enabling this flag without setting any specific keys with the now removed SetPartial has a useful side effect of preserving all of the resource's previous state. Although confusing, it has been discovered that during an update when an error is returned, the proposed config is set into state, even without any calls to d.Set._

_In practice this default behavior goes mostly unnoticed since Terraform refreshes between operations by default. The state situation discussed is subject to further investigation and potential change. Until then, this function has been preserved for the specific usecase._

## Testing plan

1.  Make sure to `go install` the provider with these changes and to setup your overrides file. 
1. Create a configuration `main.tf` like so:
```hcl
provider "tfe" {}

resource "tfe_oauth_client" "test" {
  name = "test-oauth-client"
  organization = "awesome-org"
  api_url = "https://api.github.com"
  http_url = "https://github.com"
  oauth_token = "<MY GITHUB PAT>"
  service_provider = "github"
}

  name = "my-wkspace"
  organization = "awesome-org"

  vcs_repo {
    identifier = "my-user/my-repo"
    # make sure this branch exists on your repo
    branch = "main" 
    oauth_token_id = tfe_oauth_client.test.oauth_token_id
  }
}
```

You will need to create a VCS repo in Github and a Personal Access Token (PAT) if you have not done so already.
1. Run `TF_CLI_CONFIG_FILE=/path/to/overrides.tfrc terraform init && TF_CLI_CONFIG_FILE=/path/to/overrides.tfrc terraform apply`
1.  Once applied successfully, go to your configuration and change `branch = "main"` to `branch = "nonexistent"`.
1. Repeat the apply and expect a failure.
1. Open `terraform.tfstate` and ensure that the value of `branch` is `main` not `nonexistent`
1. Great success!


tfe/resource_tfe_workspace.go
```diff
@@ -482,6 +483,7 @@ func resourceTFEWorkspaceUpdate(d *schema.ResourceData, meta interface{}) error
 		if !ok {
 			_, err := tfeClient.Workspaces.RemoveVCSConnectionByID(ctx, id)
 			if err != nil {
+				d.Partial(true)
```
brandonc, line 486: I'm wondering if we should also use this below for error cases within changes to ssh_key_id, tag_names, remote_state_consumer_ids ?

brandonc (commented):
❓👇

sebasslash:
I think the discrepancy here is that whenever you update a VCS repo you actually have to remove the previous VCS connection entirely first -- which would explain why Terraform is failing to refresh properly (?)

@omarismail feel free to correct me if I'm wrong

brandonc (approved):

tfe/resource_tfe_workspace.go
annawinkler: @sebasslash is there an answer to this question? 🤔

PR #507, opened 2022-05-31
Update logging

uturunku1:
## Description

This issue was reported by Support and there is an internal ticket associated to this PR.
 
When a user of the tfe provider elevates logging to trace or debug mode, we will print out to the screen the API request and response data with the help of this library/method httputil.DumpRequestOut. 
Exposing details about these requests and responses is okay most of the time. In the cases of responses, Atlas redacts some of their output and the response do not show any potential sensitive data. When it comes to requests, GET requests, like to read a resource, will not contain a payload with potential sensitive so printing details of this request onto the screen is safe too. The issue seems to be when we print POST request, for example to create a variable. In this case, debug/trace logs will show all details about the payload that is being sent into that POST request:

exposed

I think is tfe provider's responsibility to be a little more discreet in their usage of httputil.DumpRequestOut and avoid dumping all details when a resource has been marked with `sensitive: true`

My solution is to suppress the exposure of the whole body data when we see that the data has been marked with the attribute `sensitive:true`. At first I considered only redacting the value field but then I went for redacting the whole body for 3 reasons:
1.Customers could still prefer to request for other fields (such as description and field name) to not be exposed as they are all reflecting details on that resource that was marked as sensitive.
2.Showing limited information for Request details would match the behavior seeing around Response details, where we only show headers.
3.For troubleshooting purposes we may only need to info about the headers, status code but not details on the body.

I am removing the exposure of the body for both requests and responses. This is overkill for the responses, but it also doesn't hurt to have that check there.

## Testing Plan

Setup any TFE environment for which you have a user token. Use the TFE provider with this token and TF_LOG set to debug or trace. Here is an example of a config that is creating a variable resource:

```
terraform {
  required_providers {
    tfe = {
      source = "hashicorp/tfe"
      version = "0.31.0"
    }
  }
}

provider "tfe" {
  token    = your-token
  hostname = "app.staging.terraform.io"
}

  name         = "some-name"
  organization = "your-staging-orgname"
}

resource "tfe_variable" "test" {
  key          = "my_key_name"
  value        = "blugblug"
  category     = "terraform"
  workspace_id = tfe_workspace.test.id
  description  = "a useful description"
  sensitive = true
}
```

After running `terraform init` run `terraform init` 

Notice that after saying "yes" to the prompt for creating the resources, you'll see the POST request for the sensitive data showing up like on my screenshot on top.

To test for my changes, do git checkout on my branch and the run `go install`

Then go through the same process of creating a variable resource using my config sample. You can just add a new variable to the same config, but change the value and the key name.

Then instead of running `terraform apply`, run `TF_LOG=TRACE TF_CLI_CONFIG_FILE=/path-to-your/dev.tfrc terraform apply` so you can point to the new built binary for tfe provider.

The output in the CLI should not exposed the body:

redacted

_Describe why you're making this change._

_Remember to:_

## Testing plan

1.  _the conditions_

## External links

```

...
```


tfe/resource_tfe_organization_token.go
```diff
@@ -35,6 +35,13 @@ func resourceTFEOrganizationToken() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+
+			"keepers": {
```
brandonc, line 39: I don't think this belongs in this PR

brandonc (changes requested):
I don't think the `keepers` stuff belongs in this PR. Did you base on main?

Otherwise this looks like it could work.

tfe/resource_tfe_organization_token.go
uturunku1: ugh, sorry. I branched off the previous PR I was reviewing. I am going to rebase.

tfe/logging.go
```diff
@@ -33,10 +35,21 @@ func IsDebugOrHigher() bool {
...
+	includeBody := true
+
+	bodyHasSensitiveVals, err := findSensitiveValues(req)
+	if err != nil {
+		return &http.Response{}, err
+	}
+
+	if bodyHasSensitiveVals {
+		includeBody = false
+	}
```
brandonc, line 47: OK this is just an optional simplification I have in mind but what if you ignored errors from findSensitiveValues and returned false in those circumstances? Then all this could become:

```suggestion
	includeBody = !hasSensitiveValues(req)
```

Much more readable and very low risk since the regex pattern is literal. You could add a panic for not being able to read the request body.

brandonc (approved):

tfe/logging.go
uturunku1: I like this suggestion!

brandonc: Sorry, you are correct. That part is not returning an error and already panics if it doesn't compile.

PR #510, opened 2022-06-02
Change default for `file_triggers_enabled` to false

matejrisek:
## Description

Change default for `file_triggers_enabled` to false.

Why?
Current default is wrong and we're having a clause on the TFC side to account for this behavior.
New trigger types, e.g. tag based triggering, will depend on this field being set to `false`. We'll introduce the conflict between the new fields and `file_triggers_enabled`.

_Remember to:_

## Testing plan

1. Define a new workspace without `file_triggers_enabled`
2. Newly created workspace should have the field set to `false`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ BUG FIXES:
 FEATURES:
 * r/team, d/team: Add manage_run_tasks to the tfe_team organization_access attributes ([#486](https://github.com/hashicorp/terraform-provider-tfe/pull/486))
 
+BREAKING CHANGES:
+* r/tfe_workspace: Default value of the `file_triggers_enabled` field is changed to `false`. 
+  If the value of the `file_triggers_enabled` field was not explicitly set and either of the fields `working_directory`
+  (not an empty string) or `trigger_prefixes` was used - to keep the behavior unchanged the `file_trigger_enabled` 
+  field should now explicitly be set to `true`. 
+
 ## 0.31.0 (April 21, 2022)
 
 BUG FIXES:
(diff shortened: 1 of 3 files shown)


omarismail:
@matejrisek Any reason to wait on updating the documentation in this PR

annawinkler (approved):
🚀

brandonc (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ BUG FIXES:
 FEATURES:
 * r/team, d/team: Add manage_run_tasks to the tfe_team organization_access attributes ([#486](https://github.com/hashicorp/terraform-provider-tfe/pull/486))
 
+BREAKING CHANGES:
+* r/tfe_workspace: Default value of the `file_triggers_enabled` field is changed to `false`.
```
omarismail, line 10: We should reference that `false` is the default value in the TFC API.

omarismail (approved):
two comments, otherwise lgtm. ping for another ✅

sebasslash (changes requested):
Looks good 🔥 . Last thing would be to update the relevant provider documentation as well:

https://github.com/hashicorp/terraform-provider-tfe/blob/6f515c30a7cf5281addf417dd40fe730d9716ee5/website/docs/r/workspace.html.markdown?plain=1#L72-L75

brandonc (approved):

matejrisek:
> Looks good 🔥 . Last thing would be to update the relevant provider documentation as well:
> 
> https://github.com/hashicorp/terraform-provider-tfe/blob/6f515c30a7cf5281addf417dd40fe730d9716ee5/website/docs/r/workspace.html.markdown?plain=1#L72-L75

Good catch @sebasslash, addressed in 21d20d4c4145149701f3fb280657aedad235f2c2

PR #514, opened 2022-06-08
Revert TF_TOKEN_... authentication method

brandonc:
## Description

This new feature was ill-advised because it can't work in the terraform cloud remote execution model because 1) if you remove it from the agent env, the provider will not be able to use it as a means of authentication and 2) if you don't remove it, the build agent can't be authenticated by the run token that it needs.

## Testing plan

This is a revert commit of [PR 477](https://github.com/hashicorp/terraform-provider-tfe/pull/477)

## External links

- [Credentials documentation](https://www.terraform.io/cli/config/config-file#credentials-1)


tfe/plugin_provider_test.go
```diff
@@ -67,7 +67,7 @@ func TestPluginProvider_providerMeta(t *testing.T) {
 		}
 
 		if tc.hostname == "" && meta.hostname != "" {
-			t.Fatalf("Test %s: hostname was not set in config and meta hostname should be empty in this moment (in retrieveProviderMeta). It is parsed later in within the `getClient` function", name)
+			t.Fatalf("Test %s: hostname was not set in config and meta hostname should be empty in this moment (in retrieveProviderMeta). It is parsed later in wihtin the `getClient` function", name)
```
annawinkler, line 70:
```suggestion
			t.Fatalf("Test %s: hostname was not set in config and meta hostname should be empty in this moment (in retrieveProviderMeta). It is parsed later in within the `getClient` function", name)
```

annawinkler (approved):
🥇 Thank you for investigating this and the revert commit.

PR #520, opened 2022-06-16
Fix flaky TerraformVersion tests with large blast radius

nfagerlund:
## Description

- Newly created workspaces default to the latest available Terraform version if
  you don't explicitly request one, with "latest" determined by comparing
  version numbers.

- `TerraformVersions` are global to the TFC instance, not scoped to org or anything.

- There's nothing preventing unrelated processes from creating workspaces on the
  test TFC instance while these tests are running, and in fact, our CI happens
  to use the same test instances for concurrent tests in multiple projects.

- These test `TerraformVersions` are unusable (since their URLs and SHAs are
  garbage), so any test whose workspace is accidentally set to one will fail.

- TFC will not allow you to delete an existing `TerraformVersion` if any
  workspace's `terraform_version` attribute is set to its exact `version`
  string, so if any unrelated test was destroyed due to getting a bad version,
  the test that owns that version will also fail when its cleanup destroy gets
  prevented.

- If a bad and high-numbered `TerraformVersion` sticks around due to its cleanup
  destroy getting blocked, subsequent tests on the same TFC instance will
  continue failing, a _long_ way away from where the original problem occurred.

The solution is to make sure the bogus Terraform versions we create for these
tests will never be selected as the latest Terraform version. We can do that by
ensuring our random version numbers are always lower than the most recent
official versions, which are reliably present in a stock TFC/TFE instance.

## Testing plan

1.  Tests pass, and we hopefully end up with fewer total flakes over time, both here and in go-tfe.

## External links

- go-tfe PR TBA

```
$ ENABLE_TFE=1 TESTARGS="-run TestAccTFETerraformVersion" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFETerraformVersion -timeout 15m
=== RUN   TestAccTFETerraformVersion_basic
--- PASS: TestAccTFETerraformVersion_basic (5.32s)
=== RUN   TestAccTFETerraformVersion_import
--- PASS: TestAccTFETerraformVersion_import (6.43s)
=== RUN   TestAccTFETerraformVersion_full
--- PASS: TestAccTFETerraformVersion_full (4.95s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	16.855s
```

I'm not going to paste the entire results of re-running the tests with `TF_LOG=trace`, but I will note that if you spy on the HTTP request bodies, it's sure enough using expected version values like `"version": "1.0.8849921612919164019"` (instead of e.g. 8496220000669343191.8496220000669343191.8496220000669343191).

### DIFF
--- a/tfe/resource_tfe_terraform_version_test.go
+++ b/tfe/resource_tfe_terraform_version_test.go
@@ -19,8 +19,7 @@ func TestAccTFETerraformVersion_basic(t *testing.T) {
 
 	tfVersion := &tfe.AdminTerraformVersion{}
 	sha := genSha(t, "secret", "data")
-	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
-	version := genVersion(rInt)
+	version := genSafeRandomTerraformVersion()
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -46,8 +45,7 @@ func TestAccTFETerraformVersion_basic(t *testing.T) {
 
 func TestAccTFETerraformVersion_import(t *testing.T) {
 	sha := genSha(t, "secret", "data")
-	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
-	version := genVersion(rInt)
+	version := genSafeRandomTerraformVersion()
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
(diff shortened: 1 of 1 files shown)


tfe/resource_tfe_terraform_version_test.go
```diff
@@ -252,6 +249,18 @@ func genSha(t *testing.T, secret, data string) string {
...
+// `1.0.<RANDOM>`, which TFC won't ever select as the latest available
+// Terraform. (At the time of writing, a fresh TFC instance will include
+// official Terraforms 1.2 and higher.) This is necessary because newly created
+// workspaces default to the latest available version, and there's nothing
+// preventing unrelated processes from creating workspaces during these tests.
+func genSafeRandomTerraformVersion() string {
+	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
+	// Avoid colliding with an official Terraform version. Highest 1.0 was
+	// 1.0.11, so add a little padding and call it good.
+	for rInt < 20 {
```
brandonc, line 262: 😁

brandonc (approved):

Issue #528, opened 2022-06-21
Latest provider v0.32.0 seems to revert all team accesses to organization owner team


jcmcken:
#### Terraform Cloud/Enterprise version

```
v202112-2
```
#### Terraform version

```plaintext
Terraform v1.2.2
on linux_amd64
```

#### Terraform Configuration Files

```terraform
resource "tfe_team_access" "foo" {
  access = "read"
  team_id = data.tfe_team.foo.id
  workspace_id = tfe_workspace.workspace.id
}
```

Here, ``data.tfe_team.foo.id`` is an additional team we created. The provider seems to ignore this setting and set ``team_id`` to the organization owner team's ID instead.

#### Expected Behavior

If you revert to 0.31.0, our plans show no changes. 

#### Actual Behavior

Once we upgrade to 0.32.0, every single team access in our environment gets updated to point to the organization owner team regardless of what the code says.

brandonc:
@jcmcken Thanks for the report! I think I've found the bug within the new release and will release a patch shortly.

PR #530, opened 2022-06-21
fix(d/team): the wrong team can be picked within the "filter by name" fallback

brandonc:
## Description

This fallback method for selecting a team by name will pick the wrong team when doing a linear search

## Testing plan

Fixes https://github.com/hashicorp/terraform-provider-tfe/issues/528 - see issue details for example config


sebasslash (approved):
👍 👍

nfagerlund (approved):

PR #538, opened 2022-06-27
Log warnings on 404 responses

sebasslash:
## Description

This PR stems from API behaviors returning a 404 when a token is invalid. In this case, the customer was reporting a "successful" deletion of a registry module when using an organization token. Even though Terraform recognizes a successful apply, the module was not deleted.  The culprit?

https://github.com/hashicorp/terraform-provider-tfe/blob/3804563b40eac5cec533fff11058227eebfc900d/tfe/resource_tfe_registry_module.go#L170-L175

The error check `if err == ErrResourceNotFound { return nil }` is quite common throughout the provider, the assumption being the 404 explicitly states the resource no longer exists. This is not the case because if we look at the [API Docs](https://www.terraform.io/cloud-docs/api-docs/private-registry/modules#delete-a-module), we find we can receive a 404 if I am not authorized -- in the customer's case the organization token can not be used to manage private registry modules. 

The issue isn't entirely apparent to a customer who may not have read the fine print of the API docs. Hopefully to provide some guidance on 404s, we should log that the requested resource was not found and that this may indicate drift or an invalid token. 

### To reproduce
1. Create a new Terraform configuration as follows:
```hcl
provider "tfe" {}

resource "tfe_oauth_client" "test" {
  name             = "github-oauth-tf-tst"
  organization     = "my-awesome-org"
  api_url          = "https://api.github.com"
  http_url         = "https://github.com"
  oauth_token      = "<YOUR-GITHUB-PAT>"
  service_provider = "github"
}

resource "tfe_registry_module" "test-registry-module" {
  vcs_repo {
    display_identifier = "<username/repo>"
    identifier = "<username/repo>"
    oauth_token_id     = tfe_oauth_client.test.oauth_token_id
  }
}
```
**Note**: Make sure to substitute the values in `<>` accordingly! and that your `token` or `TFE_TOKEN` points to a user token. 

2. Run `terraform init && terraform apply`. The apply should succeed. 

3. Change your provider configuration so that `token` or `TFE_TOKEN` points to an **organization** token.  

4.  Comment out the resources in the configuration and run `TF_CLI_CONFIG_FILE=/path/to/overrides.tfrc TF_LOG=WARN terraform apply`. You should see a warning like so:
```sh
provider.terraform-provider-tfe: [WARN] The requested resource at POST /api/v2/registry-modules/actions/delete/my-awesome-org/example could not be found. 
Please ensure no drift occurred by attempting to import the desired resource. It may also be that your token is invalid. 
```

This will work for all 404 responses, so feel free to test against drift or using an invalid token. You should also see the warning if the logging is set to `TF_LOG=TRACE`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ENHANCEMENTS:
 * d/agent_pool: Improve efficiency of reading agent pool data when the target organization has more than 20 agent pools ([#508](https://github.com/hashicorp/terraform-provider-tfe/pull/508))
+* Added warning logs for 404 error responses ([#538](https://github.com/hashicorp/terraform-provider-tfe/pull/538))
 
 ## 0.33.0 (July 8th, 2022)
 
--- a/tfe/logging.go
+++ b/tfe/logging.go
@@ -26,10 +26,16 @@ const (
 // header values should be redacted from logs
 var redactedHeaders = []string{"authorization:", "proxy-authorization:"}
 
-// IsDebugOrHigher returns whether or not the current log level is debug or trace
-func IsDebugOrHigher() bool {
+// logLevelSet reads the TF_LOG level and ensures it is valid
+func logLevelSet() bool {
 	level := strings.ToUpper(os.Getenv(EnvLog))
-	return level == "DEBUG" || level == "TRACE"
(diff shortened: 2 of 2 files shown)


laurenolivia (approved):
Phew, ok! This works as expected and the refactor is clear ✅

Issue #529, opened 2022-06-21
Applying tags with uppercase letters to a workspace causes the workspace to be modified each time a plan is run.


Josh-Tracy:
#### Terraform Cloud/Enterprise version

Terraform Cloud

#### Terraform version

```plaintext
Terraform v1.2.3
on darwin_arm64
+ provider registry.terraform.io/hashicorp/tfe v0.32.0
```

#### Terraform Configuration Files

I'm calling a module one directory down
```terraform
module "workspacer" {
  source = "../.."

  organization   = "ise-joshua-tracy"
  workspace_name = "workspacer-test"
  workspace_desc = "Alex's worksapcer test"
  workspace_tags = ["azure", "infra", "test", "Alex"]
}
```

The modules workspace resource
```terraform
resource "tfe_workspace" "ws" {
  organization                  = var.organization
  name                          = var.workspace_name
  description                   = var.workspace_desc
  execution_mode                = var.execution_mode
  agent_pool_id                 = var.agent_pool_id
  auto_apply                    = var.auto_apply
  terraform_version             = var.terraform_version
  working_directory             = var.working_directory
  global_remote_state           = var.global_remote_state
  remote_state_consumer_ids     = var.remote_state_consumer_ids
  structured_run_output_enabled = var.structured_run_output_enabled
  ssh_key_id                    = var.ssh_key_id
  allow_destroy_plan            = var.allow_destroy_plan
  tag_names                     = var.workspace_tags
```

#### Expected Behavior

1.  Terraform plan and apply creates the workspace with the workspace_tags
2. Each additional plan should show no modifications if the tags are not modified

#### Actual Behavior

1. Terraform plan and apply is creating the workspace with the workspace_tags as expected
2. Terraform is modifying the uppercase letters in the Alex tag to lowercase when adding it to the workspace
3. Each additional plan is attempting to replace the lowercase tag on the existing resource with the uppercase one, causing modifications. 

```terraform
  # module.workspacer.tfe_workspace.ws will be updated in-place
  ~ resource "tfe_workspace" "ws" {
        id                            = "ws-i65QPpVEqBRG94LM"
        name                          = "workspacer-test"
      ~ tag_names                     = [
          + "Alex",
          - "alex",
            # (3 unchanged elements hidden)
        ]
        # (14 unchanged attributes hidden)
    }
```

#### Additional Context

[...]

PR #540, opened 2022-06-28
Log warning when workspace tag names contain uppercased characters

sebasslash:
## Description

This PR stems from #529 where the user was reporting a constant modified plan for the workspace whenever the tags in the workspace configuration contained uppercase characters, ex:
```hcl
  name = "my-workspace"
  organization = "my-org"
  tag_names = ["Foo", "Bar", "foobar"]
}
```
Would result in a planned change from `"Foo" -> "foo"` and `"Bar" -> "bar"` and only happens after the workspace is created. The reason for this odd behavior stems from the API auto-downcasing tag names while at the same time accepting tag names that are not case sensitive. So your configuration will always have `Foo`, because the API accepts this value and state thinks it needs to store `foo` because the API returns this value during a refresh operation. 

There isn't an immediate solution to fix this API conflict, so I'm resolving to logging a warning mentioning the would-be affected tag names and updating the documentation to mention lowercase characters specifically. 

## Considered Solutions

* Preemptively lowering the casing when reading the attribute from the configuration. This works well for `Create` but would fail to work for `Update`. `Update` relies on a method `d.HasChange()` to determine whether this attribute needs to be updated. The problem lies with the fact `d.HasChange()` compares the values from your configuration with a refreshed view of state (and of course the provider cannot modify your configuration to match it). 

* An API change to either store non-case sensitive tag names or throw an error if the request contains a tag name that needs to be downcased. Regardless, this route requires further consideration & more internal discussion, so will be out of scope for the foreseeable future. 
 
## Testing plan

1.  Using the latest provider version, create a workspace configuration, with `tag_names` containing at least one string with an uppercase character. 
2. `apply` your changes. It should succeed! 
3. Without making any configuration changes, run `terraform apply` again. This time you should see a planned change to update the tag name(s) that contains an uppercase character. 
4. Now pull these changes and override your provider. This time, we'll add the flag `TF_LOG=WARN`:

```sh
TF_CLI_CONFIG_FILE=/path/to/overrides.tfrc TF_LOG=WARN terraform apply
```

5. You should see some log output after the confirmation phase of the run:

```sh
2022-06-28T14:20:15.154-0400 [WARN]  provider.terraform-provider-tfe: [WARN] The tag "Foo" contains uppercased characters that will be downcased by the API. Please update your configuration to avoid conflicts with state. 
```


tfe/workspace_helpers.go
```diff
@@ -104,3 +104,12 @@ func readWorkspaceStateConsumers(id string, client *tfe.Client) (bool, []interfa
 
 	return false, remoteStateConsumerIDs, nil
 }
+
+func tagContainsUppercase(tag string) bool {
+	for _, c := range tag {
+		if c < 'a' || c > 'z' {
+			return true
+		}
```
brandonc, line 112: 💅 
```suggestion
		if unicode.IsUpper(c) {
			return true
		}
```

tfe/resource_tfe_workspace.go
```diff
@@ -669,3 +676,8 @@ func resourceTFEWorkspaceImporter(d *schema.ResourceData, meta interface{}) ([]*
 
 	return []*schema.ResourceData{d}, nil
 }
+
+// Warns the user that a workspace tag containing uppercase characters will be downcased.
+func warnWorkspaceTagsCasing(ctx context.Context, tag string) {
+	log.Printf("[WARN] The tag \"%s\" for contains uppercased characters that will be downcased by the API. Please update your configuration to avoid conflicts with state.", tag)
```
brandonc, line 682:
```suggestion
	log.Printf("[WARN] The tag \"%s\" contains uppercase characters that will be transformed by the API. Please update your configuration to lowercase tags in order to avoid conflicts with state.", tag)
```

brandonc (changes requested):

go.mod
```diff
@@ -77,3 +77,5 @@ require (
 	google.golang.org/genproto v0.0.0-20210319143718-93e7006c17a6 // indirect
 	google.golang.org/grpc v1.36.0 // indirect
 )
+
+require github.com/hashicorp/terraform-plugin-log v0.4.1
```
brandonc, line 81: I think adding this was not intentional? You can `go mod tidy` it out.

brandonc (changes requested):

go.mod
sebasslash: Good catch, thanks!

brandonc (approved):

PR #508, opened 2022-06-01
Use filtering to improve performance of agent pool data source

JarrettSpiker:
## Description

Once agent pool filtering is enabled (currently off in production behind the `remove-agent-pool-limit` feature flag) we can filter agent pool list requests by agent pool name. This improves performance of the agent pool data source when the organization has >20 agent pools by reducing the number of API calls we need to make in order to loop through all of the pages of agent pools

## Testing plan

1.  Set up a TFC organization with >20 agent pools, and a terraform config with a data source matching one of those agent pools
1.  Verify that with the `remove-agent-pool-limit` feature flag off, the data source resolves the agent pool
2. Verify that with the `remove-agent-pool-limit` feature flag on, the data source resolves the agent pool even faster
   -  The difference for me, against an org with ~1100 agent pools, is 6s w/ the feature flag vs. 1m12s w/o the feature flag

## External links

- API documentation: In progress...
- [go-tfe PR](https://github.com/hashicorp/go-tfe/pull/417)

## Output from acceptance tests

```
 TESTARGS="-run  TestAccTFEAgentPoolDataSource_" make testacc
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run  TestAccTFEAgentPoolDataSource_ -timeout 15m
?   	github.com/hashicorp/terraform-provider-tfe	[no test files]
=== RUN   TestAccTFEAgentPoolDataSource_basic
--- PASS: TestAccTFEAgentPoolDataSource_basic (8.57s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	8.798s
?   	github.com/hashicorp/terraform-provider-tfe/version	[no test files]

...
```

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202293596899902

PR #561, opened 2022-07-20
Allow agent pools to be imported by name

JarrettSpiker:
## Description

Allow agent pools to be imported by `org name/agent pool name`

## Testing plan

1.  Check that agent pool data sources still works
2. Check that agent pools can still be imported by ID
3. 2. Check that agent pools can  be imported by org/name
4. Check that invalid import ID formats fail (eg a/b/c)
5. Check for error if org name is not found
6. Check for error if pool name is not found

## External links

- [Related Earlier PR](https://github.com/hashicorp/terraform-provider-tfe/pull/508)
- [JIRA](https://hashicorp.atlassian.net/browse/TF-414)

```
❯ go test github.com/hashicorp/terraform-provider-tfe/tfe  -run TestAccTFEAgentPool_import -v
=== RUN   TestAccTFEAgentPool_import
--- PASS: TestAccTFEAgentPool_import (104.15s)
PASS
ok
...
```

### DIFF
--- a/tfe/agent_pool_helpers.go
+++ b/tfe/agent_pool_helpers.go
@@ -6,7 +6,7 @@ import (
 	tfe "github.com/hashicorp/go-tfe"
 )
 
-func fetchtAgentPoolId(orgName string, poolName string, client *tfe.Client) (string, error) {
+func fetchtAgentPoolID(orgName string, poolName string, client *tfe.Client) (string, error) {
 	// to reduce the number of pages returned, search based on the name. TFE instances which
 	// do not support agent pool search will just ignore the query parameter
 	options := tfe.AgentPoolListOptions{
--- a/tfe/agent_pool_helpers.go
+++ b/tfe/agent_pool_helpers.go
@@ -6,7 +6,7 @@ import (
 	tfe "github.com/hashicorp/go-tfe"
 )
 
-func fetchtAgentPoolID(orgName string, poolName string, client *tfe.Client) (string, error) {
+func fetchAgentPoolID(orgName string, poolName string, client *tfe.Client) (string, error) {
 	// to reduce the number of pages returned, search based on the name. TFE instances which
 	// do not support agent pool search will just ignore the query parameter
 	options := tfe.AgentPoolListOptions{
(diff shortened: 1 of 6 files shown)


tfe/resource_tfe_agent_pool.go
```diff
@@ -109,3 +110,27 @@ func resourceTFEAgentPoolDelete(d *schema.ResourceData, meta interface{}) error
...
+		org := s[0]
+		poolName := s[1]
+		poolID, err := fetchtAgentPoolID(org, poolName, tfeClient)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"error retrieving agent pool with name %s from organization %s %w", poolName, org, err)
+		}
+
+		d.SetId(poolID)
+	}
```
SwiftEngineer, line 133: should we handle cases where users accidentally throw an extra slash or two in there? something like `/my-org/my-agent-pool/`

SwiftEngineer (commented):
One note about a case that may or may not deserve some additional handling, other than that this looks great! 🎉

tfe/resource_tfe_agent_pool.go
JarrettSpiker: That's an interesting questions.... if they put slashes anywhere other than the front or end (eg `my/org/my-agent-pool`) it will error out and tell the user the correct format

I guess if they add extra slashes that dont break our parsing...do we care? Im not really sure what the correct behaviour there should even be

sebasslash: Currently all custom importer funcs will error if the import ID doesn't follow the exact format specified. To maintain consistency, we should leave it as is :+1:

tfe/agent_pool_helpers.go
```diff
@@ -0,0 +1,38 @@
...
+		// Exit the loop when we've seen all pages.
+		if l.CurrentPage >= l.TotalPages {
+			break
+		}
+
+		// Update the page number to get the next page.
+		options.PageNumber = l.NextPage
+	}
+
+	return "", fmt.Errorf("Could not find agent pool %s/%s", orgName, poolName)
```
sebasslash, line 37: (Non Blocker)
The error that wraps this returned error will already mention `Failed to retrieve agent pool <name> in organization <name>` so we can omit formatting `%s/%s`. Perhaps this should just be:
```suggestion
	return "", tfe.ErrResourceNotFound
```

sebasslash (changes requested):
Looks good to me :+1: ! Minor grammatical error and a non blocker below :arrow_down:

brandonc (approved):
I tested it and it worked!

sebasslash (approved):

PR #563, opened 2022-07-21
Add possibility to add team members via organization membership ID

relusc:
## Description

It is possible to add users to teams in TFE by using their organization membership ID. Here's the reference to the `go-tfe` package: [go-tfe](https://github.com/hashicorp/go-tfe/blob/392b6391d911646ef772a84df95a7504f9538c78/team_member.go#L52)

This is useful for us as we provision org memberships and team memberships in one run, however usernames are not available until the user did his/her initial login to TFE.

We could use `tfe_team_organization_member` but we manage a lot of users and teams and the resource can only handle the membership of one user to one team. This produces a lot of resources to manage and with this change, this can be nicely reduced.

I did not update the tests as apparently they are skipped anyway. If you want me to update them nevertheless, just let me know :)

The same change can also be implemented for the `tfe_team_member` resource, however I focused on `..._members` as this fits our use case better. I can also update it for the other resource if you want.

This is my first contribution here, so apologies if I missed something.

_Remember to:_

I did not run the acceptance tests to be honest, as I don't have my personal TFE setup. However I compiled the provider locally and tested it this way and everything worked fine.


nfagerlund:
@relusc Thank you for opening this PR and starting a design conversation about this! 

I'd like to understand your use case a bit better. It sounds like you've already got (or can easily acquire) a list of tfe_organization_membership IDs, and want to manage them as a unit. I did a quick experiment, and it turns out that's already quite practical with the existing resource using a `for_each` argument: 

```hcl
locals {
  user_emails = [
    "dog@example.com",
    "panda@example.com",
    "cornsnake@example.com",
  ]
}

resource "tfe_team" "test" {
  name         = "my-team-name"
  organization = "my-org-name"
}

resource "tfe_organization_membership" "test" {
  for_each = toset(local.user_emails)
  organization = "my-org-name"
  email = each.key
}

resource "tfe_team_organization_member" "test" {
  for_each = toset(local.user_emails)
  team_id = tfe_team.test.id
  organization_membership_id = tfe_organization_membership.test[each.key].id
}
```

Under the hood this creates many resource instances, but in the human-facing configuration it's only one nicely contained resource block to manage.

Would this `for_each` approach work for your use case today? If not, what's the limiting factor that makes a single resource more attractive?

relusc:
@nfagerlund Thanks for your reply :) 

So in fact your example is exactly the way that we currently manage the memberships ;) The only difference is that we have 100+ TFE teams in our setup. So the `for_each` approach works, but generates a lot of resources (the `tfe_team_organization_member ` resources make approx. 25% of the state that is managed by our "TFE organization management automatism",as one resource is required for each team membership of each user).

My thought was that with this PR we only have one single resource that manages the team memberships whereas with `tfe_team_organization_member` we only need one resource per team

nfagerlund:
Hmm, okay! One more question — when you were experimenting with your branch here, did you verify that using a combined resource significantly improves run times? (Or alternately, validate via benchmarking or something that the _resource overhead_ was a significant problem?)

If you can demonstrate that the resource overhead is really that much of a problem, we'll take that as a good justification that we need a combined resource type for these relationships! But if we go that way, we'll want a brand new `tfe_team_organization_memberships` resource, rather than bolting new semantics onto the by-username version. That's because:

1. The version of the plugin SDK we're using doesn't always behave reliably with mutually-exclusive mode-switching attributes like that. In the past, we've seen that pattern result in strange one-run-delay bugs, bogus planned changes, and other pathological behaviors. 
2. Simple resources with a single job are easier to understand, test, and maintain over time.

relusc:
Hey @nfagerlund,

I tested my branch locally and replicated the TFE setup we currently have.

Current: 
- ~120 `tfe_team` resources and ~1500 `tfe_team_organization_member` resources. 
- plan duration after freshly creating all resources was about 15min

Updated with my changes:
- ~120 `tfe_team` resources and the same amount of`tfe_team_members` resources (1 per team as written in the PR)
- plan duration after freshly creating all resources was about 4-5m (my updated provider was buggy though, as it showed changes that were actually already made, maybe this is what you meant in your last comment)

I'm fine with creating a new resource for this, I actually had some problems using the resource because of managing both `usernames` and `organization_membership_ids` in parallel (exactly what you mentioned in your comment).

I will wait for your approval of the general concept of my changes until I do updates to this PR. I would also be fine with if we closed this PR and created a new one (for better traceability/history)

brandonc:
@relusc Hey! Thanks for accepting the feedback that a new resource is the direction we'd like to go for this optimization. I've been contemplating a way to introduce this amid the three other methods of team member assignment:

- tfe_team_organization_member (manages one membership id per team)
- tfe_team_member (manages one username per team)
- tfe_team_members (manages many usernames per team)

And it would seem that a `tfe_team_organization_members` would create a symmetry. Indeed, it's probably what `tfe_team_organization_member` ought to have been because the [TeamMembers API](https://www.terraform.io/cloud-docs/api-docs/team-members#sample-payload) accepts multiple membership IDs (or usernames) for the create and delete actions. Duplicating that resource and modifying `organization_membership_id` to be a set should be pretty straightforward by lifting code from both `tfe_team_organization_member` and `tfe_team_members`. Implementing import using team ID alone would also seem to make sense.

Before releasing this new resource, it would be worthwhile for HashiCorp to take another holistic pass at the website docs to recommend the use of `tfe_team_organization_members` or to deprecate or settle any confusion around the many competing resources. We can take on that part!

We do think this is a worthwhile change because we'd like to see the performance of planning org memberships improve. If you are not up to contributing it, I understand, and you would probably see this added by HashiCorp sooner or later! Let me know what you think.

relusc:
Hey @brandonc thanks for your comment.

I was on vacation over the last three weeks so sorry for my late reply. I'm totally fine with what you wrote and would be happy to contribute this. You can expect a PR in the next few days ;)

relusc:
@brandonc @nfagerlund just created https://github.com/hashicorp/terraform-provider-tfe/pull/617

Issue #552, opened 2022-07-15
tfe_workspace issue with trigger_patterns and trigger_prefixes


henryxuteck:
I am using `Terraform Enterprise workspace A` to manage a bunch of Terraform Enterprise workspaces.

I was trying to revert a manual change in terraform enterprise workspace by rerunning `Terraform Enterprise workspace A`.

Everything works fine with tfe provider version `0.32.0`, but I got below error with `0.33.0`, and I don't even have `trigger_prefixes` and `trigger_patterns` in my code. 
``'trigger_prefixes' attribute is deprecated, and cannot be used in conjunction with 'trigger_patterns'. Use the latter only.'``

Correct me if I am wrong, will below logic assign `[]` to `trigger_prefixes` and `trigger_patterns` if those attributes are not provided?
https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/resource_tfe_workspace.go#L269
https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/resource_tfe_workspace.go#L277

Has `trigger_prefixes` been deprecated already? this doesn't match the [doc](https://registry.terraform.io/providers/hashicorp/tfe/0.32.0/docs/resources/workspace#trigger_prefixes)

#### Terraform Cloud/Enterprise version

Terraform Enterprise: `v202205-1`

#### Terraform version

```
1.1.9
```

#### Terraform Configuration Files

```terraform
resource "tfe_workspace" "default" {
  organization                  = var.tfe.organization.name
  name                          = var.tfe.workspace.name
  working_directory             = var.git.repository.directory
  terraform_version             = var.tfe.workspace.version
  agent_pool_id                 = var.tfe.workspace.agent_pool_id
  execution_mode                = var.tfe.workspace.execution_mode
  auto_apply                    = var.tfe.workspace.auto_apply
  ssh_key_id                    = var.tfe.workspace.ssh_key_id
  allow_destroy_plan            = true
  file_triggers_enabled         = true
  global_remote_state           = false
  queue_all_runs                = true
  speculative_enabled           = true
  structured_run_output_enabled = var.tfe.workspace.structured_run_output_enabled
  tag_names                     = var.tfe.workspace.tag_names

  vcs_repo {
    oauth_token_id     = var.tfe_oauth
    identifier         = format("%s/%s", var.git.organization.name, var.git.repository.name)
    branch             = var.git.repository.branch
    ingress_submodules = true
  }
}
```

[...]

PR #564, opened 2022-07-21
Fix/do not use computed

matejrisek:
## Description

Addresses issue [described in this GitHub issue](https://github.com/hashicorp/terraform-provider-tfe/issues/552).
It should prevent sending trigger-patterns and trigger-prefixes in the same request, which is currently causing failures for the Terraform Enterprise with version `v202205-1`. The issues is occurring because this version of TFE checks for an invariant that `trigger-patterns` and `trigger-prefixes` should NOT be present at the same time. The latest Atlas codebase (and newer versions of TFE) is checking that these values (arrays) are not populated rather for the presence, meaning the issues is not present in newer versions. 

The above fix is using raw configuration reading, and for that we had to bump up the SDK version to `v2.8.0`

This PR also addresses some other quirks we were experiencing because the fields `trigger-patterns` and `trigger-prefixes` were marked as `computed`. The `computed` trait has been removed for these fields and we introduced a more comprehensive test suite.

_Remember to:_

## Testing plan

1. Either test against the `TFE` version `v202205-1` or checkout the Atlas codebase with the same git tag
2. Create a workspace that has `trigger-prefixes` set
3. The creation should NOT fail because of the presence of both `trigger-prefixes` and `trigger-patterns`

## External links

- [GitHub Issue](https://github.com/hashicorp/terraform-provider-tfe/issues/552)

```

...
```


omarismail (approved):
amazing set of tests! LGTM

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ENHANCEMENTS:
 * d/agent_pool: Improve efficiency of reading agent pool data when the target organization has more than 20 agent pools ([#508](https://github.com/hashicorp/terraform-provider-tfe/pull/508))
 * Added warning logs for 404 error responses ([#538](https://github.com/hashicorp/terraform-provider-tfe/pull/538))
+* r/tfe_workspace: Fix `trigger-prefixes` could not be updated because of the conflict with `trigger-patterns` in some cases - as described in this [GitHub Issue](https://github.com/hashicorp/terraform-provider-tfe/issues/552) ([#564](https://github.com/hashicorp/terraform-provider-tfe/pull/564/))
```
brandonc, line 6: Can you add this to a new Bug Fixes heading?

brandonc (approved):
I smoke tested this on the latest TFC and it behaved nicely, with and without the beta feature enabled.

CHANGELOG.md
matejrisek: 👍🏻  1a89a9bd46df04ff5eb97226488f1610445b8195

Issue #557, opened 2022-07-20
Sensitive value in output can't be read using tfe_outputs


pj:
#### Terraform Cloud/Enterprise version
Terraform Cloud

#### Terraform version

```plaintext
Terraform v1.2.5                                                                                                                                                                     
on linux_amd64                                                                                                                                                                       
+ provider registry.terraform.io/hashicorp/tfe v0.30.2
```

#### Terraform Configuration Files

In workspace 1:
```terraform
resource "random_password" "password" {
  length = 32
  special = true
}

output "password" {
  value = random_password.keycloak_password.result
  sensitive = true
}
```

In workspace 2:
```terraform
data "tfe_outputs" "asdf" {
  organization = "my-org"
  workspace = "project-1"
}

output "password" {
  value = data.tfe_outputs.asdf.values
  sensitive = true
}
```

#### Expected Behavior
Running `terraform output -json` in workspace 2 should show the password from workspace 1.

#### Actual Behavior
Output of workspace 2:
```plaintext
> terraform output -json
{
  "password": {
    "sensitive": true,
    "type": [
      "object",
      {
        "password": "dynamic"
      }
    ],
    "value": {
      "password": null
    }
  }
}
```

#### Additional Context
This was also reported at https://github.com/hashicorp/terraform/issues/31234, there is more detail there. @spanktar

annawinkler:
Thank you for submitting this issue! We'll take a look and get back to you about this. ✨

annawinkler:
We have verified that this a bug and we're working on a fix. Thank you again for reporting this issue.

rafitadiaz:
Maybe I'm not getting it but shouldn't you call be like this?

```
value = data.tfe_outputs.asdf.values.password
```

to qualify which output do you want to use?

CJSmith-0141:
I'm also seeing this issue. Outputs marked sensitive in the source workspace are coming up as null in the workspace using `tfe_outputs`

ispirals:
same here, switching to `terraform_remote_state` seems to work correctly to me but it is not what is desired

brandonc:
I do believe this was fixed in [v0.36.0](https://github.com/hashicorp/terraform-provider-tfe/blob/main/CHANGELOG.md#v0360-august-16th-2022)

PR #565, opened 2022-07-22
Fix: read workspace state version outputs for sensitive values

annawinkler:
## Description

It was [reported](https://github.com/hashicorp/terraform-provider-tfe/issues/557) (thank you!) that a sensitive value in an output was not being read via tfe_outputs. Investigating the issue, we were originally getting state version outputs via reading a workspace and including outputs. This [API](https://www.terraform.io/cloud-docs/api-docs/workspaces#show-workspace) call does not include sensitive values.

The fix is to instead read the outputs from the state version itself.

🤔 Question for reviewers: I noticed that in the go-tfe test for StateVersion.ListOutputs, we have a [call](https://github.com/hashicorp/go-tfe/blob/ce0f146febc7a8ad6aef26d56a0d0ebc9f242991/state_version_integration_test.go#L484) to wait for the state version outputs. I'm wondering if we need to do something similar here? Any thoughts?

## Testing plan

I created a simple example to show the error (based on what was reported in the two issues). Replace `YOUR_ORG_NAME` with the name of the organization you wish to use to test this out. I'm also using tfcdev to run a local tfc to make sure I'm using the version of `terraform-provider-tfe` built from this branch.

The first workspace, `uno_workspace` has the following config:

```
terraform {
  required_version = ">= 1.1.0"
  cloud {
    hostname     = "app.terraform.io"
    organization = "YOUR_ORG_NAME"

    workspaces {
      name = "uno-workspace"
    }
  }

required_providers {
  tfe = {
    version = "~> 0.33.0"
  }
  random = {
    source  = "hashicorp/random"
    version = "~> 3.3.1"
  }
}
}

resource "random_password" "uno_password" {
  length  = 32
  special = true
}

output "uno_password" {
  value     = random_password.uno_password.result
  sensitive = true
}
```

The second workspace, `dos_workspace` has the following terraform config:

```
terraform {
  cloud {
    hostname     = "app.terraform.io"
    organization = "YOUR_ORG_NAME"

    workspaces {
      name = "dos-workspace"
    }
  }
  required_providers {
    tfe = {
      version = "~> 0.33.0"
    }
    random = {
      source  = "hashicorp/random"
      version = "~> 3.3.1"
    }
  }
}

data "tfe_outputs" "remote_state" {
  organization = "YOUR_ORG_NAME"
  workspace = "uno-workspace"
}

output "dos_password" {
  sensitive = true
  value = data.tfe_outputs.remote_state.values.uno_password
}
```

1. Run `terraform apply` for `uno-workspace`
2. Run `terraform plan` for `dos-workspace` and you'll see an error:

```
│ Error: Unsupported attribute
│
│   on main.tf line 29, in output "dos_password":
│   29:   value = data.tfe_outputs.remote_state.values.uno_password
│     ├────────────────
│     │ data.tfe_outputs.remote_state.values has a sensitive value
│
│ This object does not have an attribute named "uno_password".
```

3. Check out this branch, run `make build`. Update your `.terraformrc` to point `"hashicorp/tfe"` to the newly-built `terraform-provider-tfe` binary (if it's not in the tfe provider directory, it might be in `~/go/bin`).

```
provider_installation {
  dev_overrides {
    "hashicorp/tfe" = "/Users/YOURNAME/dev/terraform-provider-tfe"
  }

  direct {}
}
```

4. Then run `terraform apply` in `dos_worskpace` and it should work.

## External links

- [Reported issue](https://github.com/hashicorp/terraform-provider-tfe/issues/557)
- [Related issue](https://github.com/hashicorp/terraform/issues/31234)
- [Docs on permissions](https://www.terraform.io/cloud-docs/users-teams-organizations/permissions)

```
$ TESTARGS="-run TestAccTFEOutputs" make testacc

--- PASS: TestAccTFEOutputs_emptyOutputs (11.47s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	22.432s
```

### DIFF
--- a/tfe/data_source_outputs.go
+++ b/tfe/data_source_outputs.go
@@ -172,6 +172,14 @@ func (d dataSourceOutputs) readStateOutput(ctx context.Context, tfeClient *tfe.C
 	}
 
 	for _, op := range ws.Outputs {
+		if op.Sensitive {
+			sensitiveOutput, err := tfeClient.StateVersionOutputs.Read(ctx, op.ID)
+			if err != nil {
+				return nil, fmt.Errorf("Could not read sensitive output: %w", err)
+			}
+			op.Value = sensitiveOutput.Value
+		}
+
 		buf, err := json.Marshal(op.Value)
 		if err != nil {
 			return nil, fmt.Errorf("Could not marshal output value: %w", err)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 FEATURES:
 * r/tfe_organization_run_task, d/tfe_organization_run_task: Add `description` attribute to organization run tasks. ([#585](https://github.com/hashicorp/terraform-provider-tfe/pull/585))
 
+BUG FIXES:
+* d/tfe_outputs: Fix referencing sensitive values in tfe_outputs ([#565](https://github.com/hashicorp/terraform-provider-tfe/pull/565))
+
 ## 0.35.0 (July 27th, 2022)
 
 BREAKING CHANGES:
--- a/tfe/data_source_outputs_test.go
+++ b/tfe/data_source_outputs_test.go
@@ -86,7 +86,7 @@ func TestAccTFEOutputs_emptyOutputs(t *testing.T) {
 						"data.tfe_outputs.foobar", "organization", orgName),
 					resource.TestCheckResourceAttr(
 						"data.tfe_outputs.foobar", "workspace", wsName),
-					// This is relies on test-fixtures/state-versions/terraform-empty-outputs.tfstate
+					// This relies on test-fixtures/state-versions/terraform-empty-outputs.tfstate
 					testCheckOutputState("state_output", &terraform.OutputState{
 						Value: map[string]interface{}{},
 					}),
--- a/tfe/test-fixtures/state-versions/terraform.tfstate
+++ b/tfe/test-fixtures/state-versions/terraform.tfstate
@@ -15,7 +15,8 @@
     },
     "test_output_string": {
       "value": "9023256633839603543",
-      "type": "string"
+      "type": "string",
+      "sensitive": true
     },
     "test_output_tuple_number": {
       "value": [


tfe/data_source_outputs.go
```diff
@@ -159,19 +159,28 @@ type outputData struct {
...
+	}
+	sv, err := tfeClient.StateVersions.ReadCurrentWithOptions(ctx, ws.ID, svOpts)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading state version: %w", err)
+	}
+
+	outputs, err := tfeClient.StateVersions.ListOutputs(ctx, sv.ID, nil)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading state version outputs: %w", err)
+	}
```
sebasslash, line 178: This will error for configurations that are managed by [someone that does not have permissions](https://www.terraform.io/cloud-docs/users-teams-organizations/permissions#general-workspace-permissions) to read a SV or its outputs. So we should attempt a fallback before throwing an error.

sebasslash (changes requested):
Overall the fix looks good :+1: The caveat being what if my token doesn't have permissions to read a SV or its outputs?

tfe/data_source_outputs.go
```diff
@@ -159,19 +159,28 @@ type outputData struct {
...
-	ws, err := tfeClient.Workspaces.ReadWithOptions(ctx, orgName, wsName, opts)
+	ws, err := tfeClient.Workspaces.Read(ctx, orgName, wsName)
 	if err != nil {
 		return nil, fmt.Errorf("Error reading workspace: %w", err)
 	}
 
+	svOpts := &tfe.StateVersionCurrentOptions{
+		Include: []tfe.StateVersionIncludeOpt{tfe.SVoutputs},
+	}
+	sv, err := tfeClient.StateVersions.ReadCurrentWithOptions(ctx, ws.ID, svOpts)
```
brandonc, line 170: If you are listing the outputs in the next step, you probably don't need to include the outputs in the sv result.

tfe/data_source_outputs.go
brandonc: Which fallback did you have in mind? If you can't read outputs, this data source will be completely nonfunctional. One possibility is that if you can't read state, but you can read outputs, you could use tfeClient.WorkspaceOuputs.ReadCurrent but this is a newish feature of the API. I'm thinking that if you don't have at least workspace admin permission you're not going to get very far with the provider!

tfe/data_source_outputs.go
```diff
@@ -159,19 +159,28 @@ type outputData struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("Error reading state version: %w", err)
+	}
+
+	outputs, err := tfeClient.StateVersions.ListOutputs(ctx, sv.ID, nil)
```
brandonc, line 175: Nonblocking but useful feedback: This API reveals sensitive outputs but it's on the list of things to deprecate in the future because of this behavior.

One [alternative possibility](https://github.com/hashicorp/terraform/blob/main/internal/cloud/state.go#L106-L123) is to fetch individual outputs if they are sensitive.

brandonc: Blocking feedback: this endpoint is paginated so we have to consume all the pages.

tfe/data_source_outputs.go
annawinkler: I was under the impression that we are OK if there is an error if someone does not have permissions for the outputs. 🤔

brandonc (changes requested):

tfe/data_source_outputs.go
sebasslash: So thinking about this further, I'm wrong in my suggestion here. We should not have a fallback and talk two endpoints with varying permissions and perhaps we may be better served with a separate datasource that requires permissions to read state versions. There is value, I think, in being able to fetch outputs w/o having permissions to read a state version for a given workspace: See the [yellow box](https://www.terraform.io/language/state/remote-state-data#the-terraform_remote_state-data-source)

Merging these changes as is would strip the ability for me to read outputs from a different workspace (of which I might not be a workspace admin for).

tfe/data_source_outputs.go
annawinkler: Decided to avoid the API that is deprecated and instead load an individual output if it is sensitive. Thanks for the suggestion!

tfe/data_source_outputs.go
annawinkler: Thanks for the comment @sebasslash. I decided to change the approach - let me know what you think!

brandonc (approved):
Smoke tested this and it's behaving nicely now.

sebasslash (approved):

Issue #439, opened 2022-03-01
We should be able to rename teams.


mw-root:
For some reason the ForceNew flag is set on the name attribute in the 
[team resource
](https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/resource_tfe_team.go)
#### Use-cases

Creating a team using tfe provider and updating to rename team.

#### Attempted Solutions
N/A

#### Proposal
It appears that the go-tfe client supports it, I can't think of any reason why ForceNew shouldn't be removed from the name attribute.

sebasslash:
Closing. Resolved in #566

PR #566, opened 2022-07-25
Remove force new modifier on name attribute

sebasslash:
## Description

The issue was reported [here](https://github.com/hashicorp/terraform-provider-tfe/issues/439).

Effectively, the `ForceNew` modifier was set in the `name` attribute of the `r/tfe_team` schema causing the recreation of the team resource whenever the name was updated.

This PR removes this as there isn't any API specific behavior that requires the team to be recreated upon a name change. 

## Testing plan

In order to smoke test this, we'll need a simple configuration that provisions a team (ensure your organization is not on the free tier):
```hcl
resource "tfe_team" "foo" {
  name = "the-foo-fighters"
  organization = "rocknroll"
}
```
Create the team first using `terraform apply` and then simply modify the name, running `terraform apply` again. Overriding the provider to a local build including my changes, you should see that the team is no longer recreated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 BUG FIXES:
 * Removed nonworking example from `tfe_variable_set` docs ([#562](https://github.com/hashicorp/terraform-provider-tfe/pull/562))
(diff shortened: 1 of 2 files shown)


brandonc (approved):
It's always been possible to update a team name! 👍

Issue #570, opened 2022-07-27
tfe_organization broken in 0.34.0


lra:
#### Terraform Cloud/Enterprise version

Terraform Cloud

#### Terraform version

```plaintext
Terraform v1.2.5
on darwin_arm64
+ provider registry.terraform.io/hashicorp/tfe v0.34.0
```

#### Terraform Configuration Files

```terraform
resource "tfe_organization" "SOME_ORG" {
  name                     = "SOME_ORG"
  email                    = "SOME_EMAIL"
  collaborator_auth_policy = "two_factor_mandatory"
}
```

#### Debug Output

A `GET /api/v2/admin/organizations/SOME_ORG` call returns 404.

#### Expected Behavior

```
$ terraform plan
No changes. Your infrastructure matches the configuration.
```

#### Actual Behavior

```
$ terraform plan
╷
│ Error: resource not found
│ 
│   with tfe_organization.seismic,
│   on organization.tf line 1, in resource "tfe_organization" "seismic":
│    1: resource "tfe_organization" "seismic" {
│ 
╵
```

#### Additional Context

We have this simple resource:
```
resource "tfe_organization" "SOME_ORG" {
  name                     = "SOME_ORG"
  email                    = "SOME_EMAIL"
  collaborator_auth_policy = "two_factor_mandatory"
}
```
It was clean up to 0.33.0, but since 0.34.0 we get the above error message.

tmatilai:
Looking at the debug log, a `GET /api/v2/admin/organizations/ICEYE` call returns 404.

The change seems to be #521 (which even put the changelog item into the previous release). Based on PR description the feature only exists in TFE, so maybe it should't be even tried in Terraform Cloud?

lra:
There is no mention of https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/organization being not available to tfcloud, and it was the case before.

Not questioning it, but it should be documented if it's the case, and can we still get the data source https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/data-sources/organization in tfcloud as the org id is needed in a lot of resources?

tmatilai:
No, I meant the admin feature added by #521.

And that indeed is documented:
> * [`admin_settings`](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/organization#admin_settings) - Settings that are only available in Terraform Enterprise.

lra:
[...]

PR #575, opened 2022-07-27
Remove `admin_settings` from `tfe_organization`

sebasslash:
## Description

This PR stems from #570 where a user with lower privileges would encounter a `404` error when managing a `tfe_organization` resource. 

The reason for this error is that `tfe_organization` makes a call to the [Admin Organization API](https://www.terraform.io/cloud-docs/api-docs/admin/organizations) to read the admin settings without accounting for the possibility the client `token` has lower privileges. Regardless, it's generally a bad idea to manage two different APIs that require different privileges within one resource.

This attribute will be removed in favor of creating a new resource `r/tfe_admin_organization` which will solely manage the admin settings for a given organization. Thus separating the concerns between a site admin and a regular user.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 0.35.0 (July 27th, 2022)
 
+BREAKING CHANGES:
+* `r/tfe_organization`: `admin_settings` attribute was removed after being released prematurely in 0.34.0, breaking existing configurations due to requiring a token with admin privileges ([#573](https://github.com/hashicorp/terraform-provider-tfe/pull/573))
+
 BUG FIXES:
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.35.0 (Unreleased)
+
+BUG FIXES:
+* **Removed Attribute** in `tfe_organization`: `admin_settings` was released prematurely, breaking existing configurations due to requiring a token with admin privileges ([#573](https://github.com/hashicorp/terraform-provider-tfe/pull/573))
```
brandonc, line 4:
```suggestion
BREAKING CHANGES:
* `tfe_organization`: `admin_settings` attribute was removed after being released prematurely in 0.34.0, breaking existing configurations due to requiring a token with admin privileges ([#573](https://github.com/hashicorp/terraform-provider-tfe/pull/573))
```

brandonc (commented):
What do you think about this version of the CHANGELOG? I want to follow the previous convention of reporting breaking changes.

PR #578, opened 2022-07-28
Small adjustment for flaky CI test

uturunku1:
## Description

Lately we have had a few recurrent flaky tests in CircleCi. For example these three:

=== Failed
=== FAIL: tfe TestAccTFEOutputs (2.44s)
    data_source_outputs_****.go:31: Step 1/1 error: Error running pre-apply refresh: exit status 1

=== FAIL: tfe TestAccTFEVariableSetsDataSource_full (3.18s)
    data_source_variable_set_****.go:42: Step 1/1 error: Check failed: 1 error occurred:
        	* Check 4/4 error: data.tfe_variable_set.foobar: Attribute 'variable_ids.#' expected "1", got "0"
  
=== Failed
=== RUN   TestAccTFEOrganizationRunTask_create
    resource_tfe_organization_run_task_test.go:44: Step 2/3 error: Error running apply: exit status 1

For this PR, I have focused on investigating just the first one because it has been failing consistently, much more than others, to the point where, these days, we are just merging PRs ignore this test failure.

INVESTIGATION DISCOVERIES

This test `TestAccTFEOutputs` started failing over 3 months ago in circle CI but not locally (the first failure can be found in app.circleci.com under workflow ID `2a63d148-5d92-4569-9bec-c427bd1c7e00`). 

I checked if any of the PRs/changes around that time could have directly affected `TestAccTFEOutputs` test but no changes seemed to correlated to this failure.

I also checked if updating the terraform-plugin-sdk dependency (where the main error is coming from) could have any effect (perhaps some subtle incompatibility between dependencies version could affected the performance of the test), but upgrading did not change behavior at all.

Tracking down the errors that get concatenated in this test failure can be a little disorienting because they come from multiple sources/libraries (plugin-sdk, terraform-exec, terraform and hcl)

Here is the full set of errors:

```
Step 1/1 error: Error running pre-apply refresh: exit status 1
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 19, in output "test_output_list_string":
          19: output "test_output_list_string" { value = nonsensitive(data.tfe_outputs.foobar.values.test_output_list_string) }
            ├────────────────
            │ data.tfe_outputs.foobar.values has a sensitive value
        
        This object does not have an attribute named "test_output_list_string".
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 20, in output "test_output_string":
          20: output "test_output_string" { value = nonsensitive(data.tfe_outputs.foobar.values.test_output_string) }
            ├────────────────
            │ data.tfe_outputs.foobar.values has a sensitive value
        
        This object does not have an attribute named "test_output_string".
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 21, in output "test_output_tuple_number":
          21: output "test_output_tuple_number" { value = nonsensitive(data.tfe_outputs.foobar.values.test_output_tuple_number) }
            ├────────────────
            │ data.tfe_outputs.foobar.values has a sensitive value
        
        This object does not have an attribute named "test_output_tuple_number".
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 22, in output "test_output_tuple_string":
          22: output "test_output_tuple_string" { value = nonsensitive(data.tfe_outputs.foobar.values.test_output_tuple_string) }
            ├────────────────
            │ data.tfe_outputs.foobar.values has a sensitive value
        
        This object does not have an attribute named "test_output_tuple_string".
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 23, in output "test_output_object":
          23: output "test_output_object" { value = nonsensitive(data.tfe_outputs.foobar.values.test_output_object) }
            ├────────────────
            │ data.tfe_outputs.foobar.values has a sensitive value
        
        This object does not have an attribute named "test_output_object".
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 24, in output "test_output_number":
          24: output "test_output_number" { value = nonsensitive(data.tfe_outputs.foobar.values.test_output_number) }
            ├────────────────
            │ data.tfe_outputs.foobar.values has a sensitive value
        
        This object does not have an attribute named "test_output_number".
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 25, in output "test_output_bool":
          25: output "test_output_bool" { value = nonsensitive(data.tfe_outputs.foobar.values.test_output_bool) }
            ├────────────────
            │ data.tfe_outputs.foobar.values has a sensitive value
        
        This object does not have an attribute named "test_output_bool".
```

Here is a summary of how these errors got added:

- Test `TestAccTFEOutputs` is importing plugin-sdk to call `resource.Test(t, resource.TestCase{Steps: []resource.TestStep{},  and some other test steps....})`

- plugin sdk iterates through each of the steps invoking `func testStepNewConfig(ctx context.Context, t testing.T, c TestCase, wd *plugintest.WorkingDir, step TestStep, providers *providerFactories)`
which sets a terraform config for each step and requires a "terraform refresh" before applying (otherwise data sources will not be updated): `err = runProviderCommand(ctx, t, func() error {return wd.Refresh(ctx)}, wd, providers)`

- Is during this `wd.Refresh(ctx)` check that we get the `return fmt.Errorf("Error running pre-apply refresh: %w", err)` top error.

- Now to run "refresh", plugin sdk is using this library `github.com/hashicorp/terraform-exec`

- Looking closing at this library, it's usage does not seem to be interfering, but just surfacing an error coming from terraform directly when performing this operation `func (b *Local) opRefresh(stopCtx context.Context, cancelCtx context.Context, op *backend.Operation, runningOp *backend.RunningOperation)`

- Somewhere within this operation we are loading config files. We use the `github.com/hashicorp/hcl/v2` to validate the terraform language and also check that config files do not contain variables references that have not being declared.

- Is during one of these HCL validations that the hcl library throws the error: `Unsupported attribute [some code line reference] This object does not have an attribute named [some output key]`.

HCL library is looking at, say, this reference in our config file test:

```
data "tfe_outputs" "foobar" {
  organization = "%s"
  workspace = "%s"
}
output "test_output_list_string" {
	sensitive = true
	value = data.tfe_outputs.foobar.values.test_output_list_string
}
```

and noticing that our `data "tfe_outputs" "foobar"` does not have an attribute named `test_output_list_string`, which is true.

However the PR that implemented this test + feature (https://github.com/hashicorp/terraform-provider-tfe/pull/344) states that the value for that attribute is retrieve from the State outputs not the config. 
But if the state outputs is not available, HCL as a backup would look into the config and data object for reference to the attribute, hence throwing that error.

So what could cause state outputs not being available on time before we perform this validation?

Maybe this test helper function could explain:

```
func createOutputs(t *testing.T, client *tfe.Client, rInt int, fileName string) (string, string, func()) {
	var orgCleanup func()

	org, err := client.Organizations.Create(ctx, tfe.OrganizationCreateOptions{
		Name:  tfe.String(fmt.Sprintf("tst-terraform-%d", rInt)),
		Email: tfe.String(fmt.Sprintf("%d@tfe.local", rInt)),
	})
	if err != nil {
		t.Fatal(err)
	}
	orgCleanup = func() {
		if err := client.Organizations.Delete(ctx, org.Name); err != nil {
			t.Errorf("Error destroying organization! WARNING: Dangling resources\n"+
				"may exist! The full error is shown below.\n\n"+
				"Organization: %s\nError: %s", org.Name, err)
		}
	}

	ws, err := client.Workspaces.Create(ctx, org.Name, tfe.WorkspaceCreateOptions{
		Name: tfe.String(fmt.Sprintf("tst-workspace-test-%d", rInt)),
	})
	if err != nil {
		t.Fatal(err)
	}

	state, err := ioutil.ReadFile(fileName)
	if err != nil {
		t.Fatal(err)
	}

	_, err = client.Workspaces.Lock(ctx, ws.ID, tfe.WorkspaceLockOptions{})
	if err != nil {
		t.Fatal(err)
	}
	defer func() {
		_, err := client.Workspaces.Unlock(ctx, ws.ID)
		if err != nil {
			t.Fatal(err)
		}
	}()

	sv, err := client.StateVersions.Create(ctx, ws.ID, tfe.StateVersionCreateOptions{
		MD5:    tfe.String(fmt.Sprintf("%x", md5.Sum(state))),
		Serial: tfe.Int64(0),
		State:  tfe.String(base64.StdEncoding.EncodeToString(state)),
	})
	if err != nil {
		t.Fatal(err)
	}
	return org.Name, ws.Name, orgCleanup
}
```

It looks like at the end we are uploading the new state version which should include outputs. I would think by the time we validate the outputs reference, we already have properly stored the state version and have access to read the state outputs. However, could it possible that due to some performance deterioration around our TFC APIs, now we are not giving enough time to ensure state outputs has been stored. So just in case, I added this last extra step to our test helper function to ensure we can read the state prior to parsing state outputs:

```
_, err = client.StateVersions.Read(ctx, sv.ID)
if err != nil {
	t.Fatal(err)
}
```

To be honest, I am not 100% confident API calls performance changes are the reason this test started failing three months ago but it was not failing at the time of its implementation (a year ago). The change I am making is treating a symptom not a cause.

TEST PLAN for local environment:

This test failure does not happen when running tests in dev environment; however you can force your test to reproduce this error locally. All you have to do is comment out my changes + this segment https://github.com/hashicorp/terraform-provider-tfe/blob/v0.35.0/tfe/data_source_outputs_test.go#L154-L161 + this line https://github.com/hashicorp/terraform-provider-tfe/blob/v0.35.0/tfe/data_source_outputs_test.go#L138, and then run `TF_ACC=1 TFE_HOSTNAME=app.staging.terraform.io TFE_TOKEN=your-stg-token go test -run TestAccTFEOutputs` in the `tfe` directory for this repo. You should see that same error you are seeing in CircleCI.

---
  - https://app.asana.com/0/0/1202671879461297

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -18,7 +18,7 @@ sweep:
 	go test ./tfe -v -timeout 60m -sweep=prod
 
 testacc: fmtcheck
-	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 15m
+	TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(TEST) -v $(TESTARGS) -timeout 15m
 
 vet:
 	@echo "go vet ."
--- a/tfe/data_source_outputs_test.go
+++ b/tfe/data_source_outputs_test.go
@@ -3,6 +3,7 @@ package tfe
 import (
 	"crypto/md5"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
@@ -25,8 +26,10 @@ func TestAccTFEOutputs(t *testing.T) {
 
 	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
 	fileName := "test-fixtures/state-versions/terraform.tfstate"
-	orgName, wsName, orgCleanup := createOutputs(t, client, rInt, fileName)
+	orgName, wsName, orgCleanup := createStateVersion(t, client, rInt, fileName)
 	defer orgCleanup()
+
+	waitForOutputs(t, client, orgName, wsName)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:                 func() { testAccPreCheck(t) },
@@ -68,7 +71,7 @@ func TestAccTFEOutputs_emptyOutputs(t *testing.T) {
 
 	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
 	fileName := "test-fixtures/state-versions/terraform-empty-outputs.tfstate"
-	orgName, wsName, orgCleanup := createOutputs(t, client, rInt, fileName)
+	orgName, wsName, orgCleanup := createStateVersion(t, client, rInt, fileName)
 	defer orgCleanup()
 
 	resource.Test(t, resource.TestCase{
@@ -110,7 +113,8 @@ func testCheckOutputState(name string, expectedOutputState *terraform.OutputStat
 	}
 }
 
-func createOutputs(t *testing.T, client *tfe.Client, rInt int, fileName string) (string, string, func()) {
+func createStateVersion(t *testing.T, client *tfe.Client, rInt int, fileName string) (string, string, func()) {
+	t.Helper()
 	var orgCleanup func()
 
 	org, err := client.Organizations.Create(ctx, tfe.OrganizationCreateOptions{
@@ -156,11 +160,39 @@ func createOutputs(t *testing.T, client *tfe.Client, rInt int, fileName string) 
 		Serial: tfe.Int64(0),
 		State:  tfe.String(base64.StdEncoding.EncodeToString(state)),
 	})
+
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	return org.Name, ws.Name, orgCleanup
+}
+
+func waitForOutputs(t *testing.T, client *tfe.Client, org, workspace string) {
+	t.Helper()
+	ws, err := client.Workspaces.Read(ctx, org, workspace)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Wait for outputs to be populated
+	_, err = retry(10, 3, func() (interface{}, error) {
+		svo, oerr := client.StateVersionOutputs.ReadCurrent(ctx, ws.ID)
+		if oerr != nil {
+			return nil, fmt.Errorf("could not read outputs: %w", oerr)
+		}
+
+		if len(svo.Items) == 0 {
+			return nil, errors.New("outputs are not ready")
+		}
+
+		return svo.Items, nil
+	})
+
+	if err != nil {
+		t.Fatal(err)
+	}
 }
 
 func testAccTFEOutputs_dataSource(rInt int, org, workspace string) string {
--- a/tfe/testing.go
+++ b/tfe/testing.go
@@ -3,6 +3,7 @@ package tfe
 import (
 	"os"
 	"testing"
+	"time"
 
 	tfe "github.com/hashicorp/go-tfe"
 )
@@ -87,5 +88,28 @@ func runTasksURL() string {
 func skipIfUnitTest(t *testing.T) {
 	if !isAcceptanceTest() {
 		t.Skip("Skipping test because this test is an acceptance test, and is run as a unit test. Set 'TF_ACC=1' to run.")
+	}
+}
+
+type retryableFn func() (interface{}, error)
+
+func retry(maxRetries, secondsBetween int, f retryableFn) (interface{}, error) {
+	tick := time.NewTicker(time.Duration(secondsBetween) * time.Second)
+	retries := 0
+
+	defer tick.Stop()
+
+	for {
+		<-tick.C
+		res, err := f()
+		if err == nil {
+			return res, nil
+		}
+
+		if retries >= maxRetries {
+			return nil, err
+		}
+
+		retries += 1
 	}
 }


nfagerlund (commented):
@uturunku1 Thank you for the detailed write-up on your investigation here! That was helpful.

I think I can shed some more light on why this is flaking; it's something we've also run into in the go-tfe tests AND the TFC integration tests:

- _Creating a state version is synchronous, but the **outputs** become available asynchronously._ To populate the outputs, TFC has to send the raw state data you uploaded through terraform-state-parser (TSP), and since that could take an unknown amount of time for a large state file, it dispatches that as a background job. 
- Therefore, there's a window of time where the _state version_ exists but the _outputs_ do not. 
- _Therefore,_ this helper function is relying on the result of an async operation without awaiting a completion signal for that operation. 

And an unawaited async is the classic bread-and-butter test flake. 👍🏼 

So, I think you're right that checking for the existence of the state version won't be very resilient, due to that window where the SV exists but the outputs don't. Instead, I think you want to repeatedly poll for the _outputs_ being available, with a timeout and a pause between polls. Kind of like [what we do over here in go-tfe](https://github.com/hashicorp/go-tfe/blob/ce0f146febc7a8ad6aef26d56a0d0ebc9f242991/helper_test.go#L1633). 

(Of course, the outputs polling is quite janky due to a current limitation in go-tfe — there's a simple bool attribute on the SV that says whether the async TSP job is done or not, but go-tfe doesn't transfer that into its SV struct! @laurenolivia has a ticket in our blocked column to fix that, but it got stalled by a DIFFERENT set of SV-related test flakes, which is what @brandonc 's `retry()` PR was about.)

nfagerlund:
Separately, I'm starting to think we're missing some critical documentation around state version outputs, because basically EVERYONE expects them to be processed synchronously and gets bitten by this, as evinced by having an exact copy of this same test flake in THREE separate internal projects. 🤣

uturunku1:
> @uturunku1 Thank you for the detailed write-up on your investigation here! That was helpful.
> 
> I think I can shed some more light on why this is flaking; it's something we've also run into in the go-tfe tests AND the TFC integration tests:
> 
> * _Creating a state version is synchronous, but the **outputs** become available asynchronously._ To populate the outputs, TFC has to send the raw state data you uploaded through terraform-state-parser (TSP), and since that could take an unknown amount of time for a large state file, it dispatches that as a background job.
> * Therefore, there's a window of time where the _state version_ exists but the _outputs_ do not.
> * _Therefore,_ this helper function is relying on the result of an async operation without awaiting a completion signal for that operation.
> 
> And an unawaited async is the classic bread-and-butter test flake. 👍🏼
> 
> So, I think you're right that checking for the existence of the state version won't be very resilient, due to that window where the SV exists but the outputs don't. Instead, I think you want to repeatedly poll for the _outputs_ being available, with a timeout and a pause between polls. Kind of like [what we do over here in go-tfe](https://github.com/hashicorp/go-tfe/blob/ce0f146febc7a8ad6aef26d56a0d0ebc9f242991/helper_test.go#L1633).
> 
> (Of course, the outputs polling is quite janky due to a current limitation in go-tfe — there's a simple bool attribute on the SV that says ~whether the async TSP job is done or not,~* but go-tfe doesn't transfer that into its SV struct! @laurenolivia has a ticket in our blocked column to fix that, but it got stalled by a DIFFERENT set of SV-related test flakes, which is what @brandonc 's `retry()` PR was about.)
> 
> * edit: actually, `resources-processed` is more conservative than that, now that I go digging for it. There's a _second_ async job that gets kicked off by the TSP completion callback, and `resources-processed` gets flipped at the end of THAT job. Woof!

Hey @nfagerlund ! Thank you much for filling me in all this background story. Polling for outputs, [ala go-tfe](https://github.com/hashicorp/go-tfe/blob/ce0f146febc7a8ad6aef26d56a0d0ebc9f242991/helper_test.go#L1633), sounds like the way to go. You mentioned that Lauren will be working an adding this relevant attribute `resources-processed` to the state versions struct. Since this seems like a good way of checking if the async data for outputs is available, I may wait for this attribute to get added first, before I make changes to this PR. 
I also see that Brandon merged [this retry PR](https://github.com/hashicorp/go-tfe/pull/485), so I am guessing this means that go-tfe ticket is no longer stalled?

brandonc:
@uturunku1 I went ahead and implemented the retry and merged this because the flake was pretty aggressive

PR #588, opened 2022-08-03
chore: adds local development instructions and helpers using dev_overrides provider installation

brandonc:
## Description

I've updated the instructions and added a new make rule to assist in smoke testing the provider using CLI config to  dev_override with the built binary. I based this workflow on my own experience smoke testing the provider. See CONTRIBUTING.md changes for more details and usage.

Can you help me test this?


CONTRIBUTING.md
```diff
@@ -26,10 +26,22 @@ Clone the repository, enter the directory, and build the provider:
...
 $ git clone git@github.com:hashicorp/terraform-provider-tfe
 $ cd terraform-provider-tfe
-$ make build
+$ make
 ```
 
-This will build the provider and put the binary in the `$GOPATH/bin` directory. To use the compiled binary, you have several different options (this list is not exhaustive):
+This will build the provider and put the binary in the project directory. To use the compiled binary, you have several different options (this list is not exhaustive):
+
+##### Using CLI config to provide a dev override (Using Terraform v0.14+)
```
annawinkler, line 34:
```suggestion
##### Using a CLI config to provide a dev override (Using Terraform v0.14+)
```

scripts/devoverride.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
annawinkler, line 1: Suggest to use env to find `bash` because it's not always in `/bin` (some systems have it in `/usr/local/bin` for example).

```suggestion
#!/usr/bin/env bash
```

CONTRIBUTING.md
```diff
@@ -26,10 +26,22 @@ Clone the repository, enter the directory, and build the provider:
...
+
+Use the rule `make devoverride` to generate a CLI config containing a dev override provider installation. This command will print a variable export that can be copied and pasted into a shell session while testing with terraform. To automatically export this override, use `eval $(make devoverride)`
```
annawinkler, line 36: This is super helpful! What are your thoughts on having the script set the environment variable so all the user needs to do is invoke `make devoverride` and then go along their merry way?

scripts/gofmtcheck.sh
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 # Check gofmt
-echo "==> Checking that code complies with gofmt requirements..."
```
annawinkler, line 4: Curious why this line was removed? I find it helpful to see what a script is doing when it runs.

.gitignore
```diff
@@ -5,7 +5,9 @@
 *.so
 *.dylib
 
+# Local build artifacts
```
annawinkler, line 8: 💯

CONTRIBUTING.md
```diff
@@ -26,10 +26,22 @@ Clone the repository, enter the directory, and build the provider:
...
+
+Example usage:
+
+```sh
+$ eval $(make devoverride)
+$ cd ../example-config
```
annawinkler, line 42: 🤔 What is the `example-config` directory? Is this where the terraform config that is being tested with the tfe provider is located?

annawinkler (changes requested):
Thank you for adding these updates - very helpful! I have a few usage related questions that I think would be helpful to resolve before this PR is merged. 💖

CONTRIBUTING.md
brandonc: Yep, that's what I was trying to illustrate!

annawinkler, line 42: 💅 I think it might be helpful to update the name of the directory in the example:
```suggestion
$ cd ../example-terraform-config
```

CONTRIBUTING.md
brandonc: I can think of a couple reasons not to modify the running shell: I'm trying to make the fact that we are setting TF_CLI_CONFIG_FILE more visible to the user because this overwrites the default one (which is `$HOME/.terraformrc`) and I think that's pretty aggressive thing to do silently. Also, when you're running this script from a different shell, or process (like make!) it uses a different process altogether and therefore won't modify your environment.

This follows a pattern of hooking into the running shell using eval that is done by tools like tfcdev, rbenv, nodenv, direnv, etc)

annawinkler: That makes a lot of sense - thank you for the explanation.

scripts/gofmtcheck.sh
brandonc: I removed it because I wanted to use eval with make and keep this as a dependency (and this can't be eval'd). I also noticed this script is a dependency for every make rule so it was being noisy without doing anything if you run a few rules in succession.

sebasslash (approved):
This is really great, tested & works for me! Thank you for this! 🔥

annawinkler (approved):
Looks good! 🌟 Could you please squash the commits before merging?

brandonc:
Oops! I forgot to squash. Sorry!

PR #549, opened 2022-07-12
Add Git tags support

hashimoon:
## Description

Adding support to Git Tags 

## Testing plan

Use resource like the following:

```tf
locals {
  organization = "..."
  oauth_token_id = "..."
  identifier = "user/repo
}

resource "tfe_workspace" "my-workspace" {
  name                  = "my-workspace"
  organization          = local.organization
  allow_destroy_plan    = true
  auto_apply            = true
  execution_mode        = "remote"
  speculative_enabled   = true
  file_triggers_enabled = false

  vcs_repo {
    identifier         = local.identifier
    ingress_submodules = false
    oauth_token_id = local.oauth_token_id
    tags_regex = "^\\d+.\\d+.\\d+$" # semantic versioning
  }
}
```

1. Verify the correct settings were set in TF Cloud/TFE VCS settings for the workspace that git tags semantic versioning is selected
   * Release a new tagged release in the repo specified in `local.identifier`
   * See a new run has been trigged
1. Change configuration to the following

```tf
locals {
  organization = "..." # set
  oauth_token_id = "..." # set
  identifier = "user/repo
  working_directory = "..." # set
}

resource "tfe_workspace" "my-workspace" {
  name                  = "my-workspace"
  organization          = local.organization
  allow_destroy_plan    = true
  auto_apply            = true
  execution_mode        = "remote"
  speculative_enabled   = true
  file_triggers_enabled = true
  working_directory = local.working_directory

  vcs_repo {
    identifier         = local.identifier
    ingress_submodules = false
    oauth_token_id = local.oauth_token_id
  }
}
```

1. Terraform plan/apply succeeds
1. Verify in TF Cloud/TFE in VCS settings git tags is not selected.

## External links

_Include any links here that might be helpful for people reviewing your PR. If there are none, feel free to delete this section._

- [go-tfe documentation](https://pkg.go.dev/github.com/hashicorp/go-tfe#VCSRepo)
- [API documentation PR](https://github.com/hashicorp/terraform-docs-common/pull/64)
- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/xxxx)

## Output from acceptance tests

_Please run applicable acceptance tests locally and include the output here. See TESTS.md to learn how to run acceptance tests._

_If you are an external contributor, your contribution(s) will first be reviewed before running them against the project's CI pipeline._

```
$ TESTARGS="-run TestAccTFEWorkspace" make testacc

...
```

PR #594, opened 2022-08-12
Fix issue with removing vcs_repo config from workspace with tags_regex

mjyocca:
## Description

Fix for the scenario of removing `vcs_repo` configuration from a workspace resource when previously setup the workspace to be triggered by tagged release via `tags_regex` 

## Testing plan

* Configure tfe_workspace resource with `vcs_repo` & `tags_regex`
* Apply the terraform configuration
* Remove vcs_repo configuration
* Attempt a terraform apply, should succeed and remove the vcs_repo setting from the workspace

Without this fix, last step would result in an error.

## External links

- [API documentation](https://www.terraform.io/cloud-docs/workspaces/settings/vcs#trigger-runs-when-a-git-tag-is-published)
- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/549)

```
$ TESTARGS="-run TestAccTFEWorkspace_updateRemoveVCSRepoWithTagsRegex" make testacc
...
```

---
  - https://app.asana.com/0/0/1202781099862818

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ FEATURES:
 
 BUG FIXES:
 * d/tfe_outputs: Fix referencing sensitive values in tfe_outputs ([#565](https://github.com/hashicorp/terraform-provider-tfe/pull/565))
+* r/tfe_workspace: Fix error when removing `vcs_repo` attribute block after previously setting a `tags_regex` value. ([#594](https://github.com/hashicorp/terraform-provider-tfe/pull/594))
 
 ## 0.35.0 (July 27th, 2022)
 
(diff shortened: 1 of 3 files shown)


brandonc:
This seems to work well!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ FEATURES:
 
 BUG FIXES:
 * d/tfe_outputs: Fix referencing sensitive values in tfe_outputs ([#565](https://github.com/hashicorp/terraform-provider-tfe/pull/565))
+* r/tfe_workspace: Fix error when removing `vcs_repo` attribute block after previously setting a `tags_regex` value. ([#594](https://github.com/hashicorp/terraform-provider-tfe/pull/594))
```
brandonc, line 9: This isn't a bug yet because tags_regex hasn't been released! 🙃 I would omit this.

brandonc (approved):
Works great. I think this is the right fix because the API assumes that the connection is removed when it tries to update trigger_patterns or trigger_prefixes, etc.

CHANGELOG.md
mjyocca: Haha great point 😆

PR #602, opened 2022-08-15
Add upgrade organization test helper, remove `SKIP_PAID` flag. 

sebasslash:
## Description

Adds a test helper `upgradeOrganizationSubscription()` that attempts to upgrade an organization's feature set if `TFE_HOSTNAME` is not enterprise. The helper will skip the test if it is running against an enterprise instance since all features are available in TFE unless otherwise specified. This PR also removes the `skipIfFree()` test helper for paid feature tests that were previously blocked from running in CI.

### DIFF
--- a/tfe/testing.go
+++ b/tfe/testing.go
@@ -114,14 +114,14 @@ func upgradeOrganizationSubscription(t *testing.T, client *tfe.Client, org *tfe.
 }
 
 func createBusinessOrganization(t *testing.T, client *tfe.Client) (*tfe.Organization, func()) {
-    org, orgCleanup := createOrganization(t, client, tfe.OrganizationCreateOptions{
-        Name: tfe.String("tst-"+randomString(t)),
-        Email: tfe.String(fmt.Sprintf("%s@tfe.local", randomString(t)))
-    })
+	org, orgCleanup := createOrganization(t, client, tfe.OrganizationCreateOptions{
+		Name:  tfe.String("tst-" + randomString(t)),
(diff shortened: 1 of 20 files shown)


brandonc:
Screen Shot 2022-08-15 at 3 48 52 PM 👀

tfe/testing.go
```diff
@@ -40,12 +67,76 @@ func testTfeClient(t *testing.T, options testClientOptions) *tfe.Client {
...
-	skip := os.Getenv("SKIP_PAID") == "1"
-	if skip {
-		t.Skip("Skipping test that requires a paid feature. Remove 'SKIP_PAID=1' if you want to run this test")
+func upgradeOrganizationSubscription(t *testing.T, client *tfe.Client, org *tfe.Organization) {
+	if enterpriseEnabled() {
+		t.Skip("Cannot upgrade an organization's subscription when enterprise is enabled. Set ENABLE_TFE=0 to run.")
+	}
+
+	req, err := client.NewRequest("GET", "admin/feature-sets", featureSetListOptions{
+		Q: "Business",
```
Uk1288, line 76: 🤔 what do you think about making this Q param dynamic, by allowing a `tier` param that defaults to `Business`. This way tests written for other feature sets can reuse this function as well.

sebasslash: Hmm I'm not sure there'd be a reason to run tests against a Developer (or any lower tier) org instead of a Business org. Upgrading an organization to Business serves as a catch all for any paid feature available on TFC.

brandonc (approved):

Issue #544, opened 2022-07-07
Flipping `r.tfe_workspace` from agent to default (remote) exec mode fails to plan


ausfestivus:
#### Terraform Cloud/Enterprise version

* Terraform Cloud

#### Terraform version

```plaintext
Terraform v1.2.4
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.22.0
+ provider registry.terraform.io/hashicorp/tfe v0.31.0
+ provider registry.terraform.io/hashicorp/vault v3.6.0
+ provider registry.terraform.io/microsoft/azuredevops v0.2.1
```

#### Terraform Configuration Files

The resource definition before the change. 

Note the use of an agent which we had to use temporarily to workaround an issue in  v1.2.6 of the TFC agent.

```terraform
resource "tfe_workspace" "tenantConfiguration" {
  name         = "tenantConfiguration-${var.environment}"
  organization = local.tfc_orgname[var.environment]

  description                   = null
  global_remote_state           = false
  tag_names                     = null
  working_directory             = null
  structured_run_output_enabled = false

  # temp agent config
  execution_mode = "agent"
  agent_pool_id  = data.tfe_agent_pool.temp.id
}
```

Once the issue was resolved, the above resource definition was returned to its normal configuration:

```terraform
resource "tfe_workspace" "tenantConfiguration" {
  name         = "tenantConfiguration-${var.environment}"
  organization = local.tfc_orgname[var.environment]

  description                   = null
  global_remote_state           = false
  tag_names                     = null
  working_directory             = null
  structured_run_output_enabled = false
}
```

#### Debug Output

* none provided, its a simple enough repro.

#### Expected Behavior

The TFC plan should have detected the change and completed the plan.

#### Actual Behavior

The TFC plan failed with the following output:

```plaintext
│ Error: agent_pool_id must be provided when execution_mode is 'agent'
│ 
│   with tfe_workspace.tenantConfiguration,
│   on tenants_workspace.tf line 57, in resource "tfe_workspace" "tenantConfiguration":
│   57: resource "tfe_workspace" "tenantConfiguration" {
```

#### Additional Context

To provide a little context for the order of events here:

* We hit an issue in the v1.2.6 release of the TFC agent which caused a TFC Workspace plan to fail
* Support asked us to switch the Workspace over to use `execution_mode = "agent" and pin the agent to the v1.2.5 release.
* Following this change the TFC Workspace was able to execute plans successfully.
[...]

PR #607, opened 2022-08-22
Fix ability to detect and apply agent_pool_id and execution_mode changes

Uk1288:
## Description

Workspaces can be created using the `resource "tfe_workspace"` with fields including `execution_mode` and `agent_pool_id`. After creation, if the config is updated by removing the `execution_mode` and `agent_pool_id` fields, the changes are not detected during an apply and plan fails with error: `Error: agent_pool_id must be provided when execution_mode is 'agent'`.

See issue: https://github.com/hashicorp/terraform-provider-tfe/issues/544

With this fix, changes in `execution_mode` and `agent_pool_id` fields are now detected and plans/applies are successful.

_Remember to:_

- _Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/CONTRIBUTING.md#updating-the-changelog)_

- _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/CONTRIBUTING.md#updating-the-documentation)_

## Testing plan

1.  _Create a workspace with config(Prerequisite: create an org and an agent pool, use the respective names in the config below)_ 

``` 
resource "tfe_workspace" "my-workspace" {
  name         = "my-test-workspace"
  organization = "my-org"

  description                   = "hello there"
  global_remote_state           = false
  tag_names                     = null
  working_directory             = null
  structured_run_output_enabled = false

  execution_mode = "agent"
  agent_pool_id  = "apool-my-agent-pool-id"
}
```

2.  _follow with an apply command_
3.  _change config by removing execution_mode and agent_pool_id,_ 

```
resource "tfe_workspace" "my-workspace" {
  name         = "my-test-workspace"
  organization = "my-org"

  description                   = "hello there"
  global_remote_state           = false
  tag_names                     = null
  working_directory             = null
  structured_run_output_enabled = false
}
```

4.  _follow with another apply command. Before this PR, changes are not detected and the apply fails. With these PR, changes should be detected and applied successfully_

## External links

- [API documentation](https://www.terraform.io/cloud-docs/api-docs/workspaces#workspaces-api)

```

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Unreleased
+ * r/tfe_workspace: Changes in `agent_pool_id` and `execution_mode` attributes are now detected and applied. ([#607](https://github.com/hashicorp/terraform-provider-tfe/pull/607))
 
 FEATURES:
 * r/tfe_workspace_run_task, d/tfe_workspace_run_task: Add `stage` attribute to workspace run tasks. ([#555](https://github.com/hashicorp/terraform-provider-tfe/pull/555))
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -34,7 +34,13 @@ func resourceTFEWorkspace() *schema.Resource {
 		},
 
 		CustomizeDiff: func(c context.Context, d *schema.ResourceDiff, meta interface{}) error {
-			err := validateAgentExecution(c, d)
+			// NOTE: execution mode must be set to default first before calling the validation functions
+			err := setExecutionModeDefault(c, d)
+			if err != nil {
+				return err
+			}
+
+			err = validateAgentExecution(c, d)
 			if err != nil {
 				return err
 			}
@@ -67,6 +73,7 @@ func resourceTFEWorkspace() *schema.Resource {
 			"agent_pool_id": {
 				Type:          schema.TypeString,
 				Optional:      true,
+				Default:       "",
 				ConflictsWith: []string{"operations"},
 			},
 
@@ -120,6 +127,7 @@ func resourceTFEWorkspace() *schema.Resource {
 				Type:          schema.TypeBool,
 				Optional:      true,
 				Computed:      true,
+				Deprecated:    "Use execution_mode instead.",
 				ConflictsWith: []string{"execution_mode", "agent_pool_id"},
 			},
 
@@ -254,7 +262,7 @@ func resourceTFEWorkspaceCreate(d *schema.ResourceData, meta interface{}) error 
 		options.ExecutionMode = tfe.String(v.(string))
 	}
 
-	if v, ok := d.GetOk("operations"); ok {
+	if v, ok := d.GetOkExists("operations"); ok {
 		options.Operations = tfe.Bool(v.(bool))
 	}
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## Unreleased
+ * r/tfe_workspace: Changes in `agent_pool_id` and `execution_mode` attributes is now detected and applied. ([#607](https://github.com/hashicorp/terraform-provider-tfe/pull/607))
```
annawinkler, line 2:
```suggestion
 * r/tfe_workspace: Changes in `agent_pool_id` and `execution_mode` attributes are now detected and applied. ([#607](https://github.com/hashicorp/terraform-provider-tfe/pull/607))
```

tfe/resource_tfe_workspace.go
```diff
@@ -85,7 +86,7 @@ func resourceTFEWorkspace() *schema.Resource {
 			"execution_mode": {
 				Type:          schema.TypeString,
 				Optional:      true,
-				Computed:      true,
```
annawinkler, line 88: I wonder why this was listed as `Computed`? Any thoughts?

Uk1288: This is most likely based on logic that whenever an execution_mode is not specified, it will be computed and returned. With this changes, the default value is always appended whenever the field is not specified.

Uk1288: Actually, Computed might still needed for the `execution_mode` to be updated when it changes as a result of `operation` field change. I am investigating this.

sebasslash: Looks like `operations` attribute has been marked as deprecated. Would it be easier to reason about if we removed the attribute?

Uk1288: With the new changes, I have marked it as deprecated to match the API docs.

tfe/resource_tfe_workspace_test.go
```diff
@@ -2249,6 +2297,22 @@ resource "tfe_workspace" "foobar" {
 }`, organization, organization)
 }
 
+// while testing the flow of unsetting execution mode as in TestAccTFEWorkspace_unsetExecutionMode
+// the resource "tfe_agent_pool" has been kept in both configs(testAccTFEWorkspace_executionModeAgent & testAccTFEWorkspace_executionModeNull)
+// this prevents an attempt to destroy the agent pool before dissasociating it from the workspace
+func testAccTFEWorkspace_executionModeNull(organization string) string {
+	return fmt.Sprintf(`
+resource "tfe_agent_pool" "foobar" {
```
Uk1288, line 2305: The test was failing during the agent mode unset flow because the first config had the `resource "tfe_agent_pool"` and the second config did not.The diff causes an attempt to delete the agent pool while applying the second config.
~ Approach 1 is to persist the `resource "tfe_agent_pool"` in both configs so the only diff action is unsetting the workspace execution mode.
~Approach 2 is to create agent pool external to the config and then use data source to grab the agent pool in the config. This would required using the go client interface.

I have gone with approach 1. Let me know your thoughts...

sebasslash: Much, much nicer than approach 2 :+1:

sebasslash (approved):
LGTM :fire: 

Smoke tested :heavy_check_mark:, ran acceptance test locally :heavy_check_mark:

PR #614, opened 2022-08-24
update CHANGELOG.md

laurenolivia:
## Description

Update CHANGELOG.md.

## Testing plan

N/A

## External links

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ FEATURES:
 * r/tfe_workspace_run_task, d/tfe_workspace_run_task: Add `stage` attribute to workspace run tasks. ([#555](https://github.com/hashicorp/terraform-provider-tfe/pull/555))
 * r/tfe_workspace_policy_set: Add ability to attach an existing `workspace` to an existing `policy set`. ([#589](https://github.com/hashicorp/terraform-provider-tfe/pull/589))
 
+BUG FIXES:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ FEATURES:
 * r/tfe_workspace_run_task, d/tfe_workspace_run_task: Add `stage` attribute to workspace run tasks. ([#555](https://github.com/hashicorp/terraform-provider-tfe/pull/555))
 * r/tfe_workspace_policy_set: Add ability to attach an existing `workspace` to an existing `policy set`. ([#589](https://github.com/hashicorp/terraform-provider-tfe/pull/589))
 
+BREAKING CHANGES:
```
brandonc, line 7: This isn't a breaking API change... the only changes that rise to that level would be removing/renaming an attribute, introducing a new required argument, creating new mutual exclusivity between existing arguments, etc. It's generally things that would require implementers to change their config in order to upgrade. It would be good to change this to `BUG FIXES:`

This first entry, `* tfe/resource_tfe_workspace_policy_set_test: Remove skipIfFreeOnly(t)` should be removed from the CHANGELOG because it isn't a customer-facing change that they need to be aware of.

brandonc (changes requested):

brandonc (approved):

PR #616, opened 2022-08-26
Migrate from CircleCI to Github Actions

sebasslash:
## Description

This is the final change to migrate this repository from CircleCI to Github Actions. This PR also includes the script to rebuild the tflocal instance nightly -- it can also be used to rebuild it manually.

This also updates the `tflocal-cloud` module version and adds the missing Artifactory variables. 

IMPORTANT FOR MAINTAINERS: Since this workflow requires reading Github secrets, PRs from forks will need to rebased and pushed into a local branch in order for the workflow to run successfully (approving a workflow still does not provide read access to secrets). Ideally, we can find some way to automate this workaround upon maintainer approval of a fork PR.


Uk1288:
What's the best way of smoke testing this?

brandonc (commented):
Looks good, but does it work?

.github/scripts/build_instance/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+// tokenAddr contains the target token that will be forcibly replaced every run
+const tokenAddr = "module.tflocal.var.tflocal_cloud_admin_token"
+
+var workspace string
+var organization string
+var isDestroy bool
+
+func init() {
+	flag.StringVar(&organization, "o", "hashicorp-v2", "the TFC organization that owns the specified workspace.")
```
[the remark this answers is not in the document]
sebasslash: Huh! Admittedly I'm unsure why it didn't auto-import.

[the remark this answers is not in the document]
sebasslash: And yes, looks like something went wrong with our CI instance. I've just destroyed it.

.github/workflows/instance.yml
```diff
@@ -0,0 +1,23 @@
+name: instance
```
annawinkler, line 1: 💅 Suggest a name that will be nice to see in the GH UI:

```suggestion
name: Create tflocal instance
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
+name: test
```
annawinkler, line 1: 💅 Suggest to update the name:

```suggestion
name: Run tests
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Sync dependencies
+        run: |
+          go get -v -t -d ./...
+          go mod tidy
```
annawinkler, line 28: If go mod tidy fails does it fail the build?

annawinkler (approved):
🥳 I have a few minor suggestions & one question. I'm OK to move forward with this PR though. Thank you!

.github/workflows/test.yml
brandonc: Yes. Any nonzero exit value on any step will cause the workflow to fail. In this case, the other two runs the matrix will continue because I specified "fail-fast: false" so that they can be retried individually.

brandonc: I moved the fmtcheck and vet steps to the lint workflow.

PR #627, opened 2022-09-16
Fix default in tfe_workspace documentation

gtownend:
## Description

Current documentation states that the default for file_triggers_enabled is false when the actual default is true 

## External links

[Default value in code](https://github.com/hashicorp/terraform-provider-tfe/blob/0f04f5566d26ffb5a9f4d5e0b37f0dd12b5c277a/tfe/resource_tfe_workspace.go#L110)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ FEATURES:
 
 BUG FIXES:
 * Bump `terraform-plugin-go` to `v0.6.0`, due to a crash when `tfe_outputs` had null values. ([#611](https://github.com/hashicorp/terraform-provider-tfe/pull/611))
+* r/tfe_workspace: Fix documentation of file_triggers_enabled default. 
 
 ## v0.36.0 (August 16th, 2022)
 
--- a/website/docs/r/workspace.html.markdown
+++ b/website/docs/r/workspace.html.markdown
@@ -70,7 +70,7 @@ The following arguments are supported:
   for state storage only. This value _must not_ be provided if `operations`
   is provided.
 * `file_triggers_enabled` - (Optional) Whether to filter runs based on the changed files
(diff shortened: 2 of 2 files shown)


brandonc (approved):
Thank you very much for correcting this! The default was changed in 0.32.0 but the argument docs were not changed.

philomory:
@brandonc As I understand it, in 0.32.0 the default was _supposed_ to be changed _to_ `false` (it was `true` previously), but the change was made incorrectly and so nothing was actually changed. You can see that the documentation _was_ updated in #510, it was just updated under the assumption that the intended change to the default value was actually going to take effect.

This gets even more confusing because it was stated in #510 (and the release notes for 0.32.0) that the change to the default was intended to bring the default value into line with the default value of the corresponding attribute in TFC's API, but the [API Documentation](https://www.terraform.io/cloud-docs/api-docs/workspaces) linked from the release notes clearly shows that the default _is_ `true`.

I know that changing the release notes of an already-released version after-the-fact is generally frowned upon, but you may wish to consider removing this from the release notes of version 0.32.0, because announcing a breaking change that never happens is pretty awkward as well: 
>r/tfe_workspace: Default value of the file_triggers_enabled field is changed to false. This will align the
file_triggers_enabled field default value with the default value for the same field in the
[TFC API](https://www.terraform.io/cloud-docs/api-docs/workspaces).
If the value of the file_triggers_enabled field was not explicitly set and either of the fields working_directory
(not an empty string) or trigger_prefixes was used - to keep the behavior unchanged, the file_trigger_enabled
field should now explicitly be set to true. ([#510](https://github.com/hashicorp/terraform-provider-tfe/pull/510/files))

I don't know if @matejrisek, @omarismail, or @sebasslash want to weigh in on this? It's pretty confusing, I lost quite a bit of time today trying to work out what was going on.

PR #629, opened 2022-09-20
Remove deprecated `workspace_ids` argument from `r/tfe_variable_set`

sebasslash:
## Description

`workspace_ids` has been marked as a deprecated argument since v0.33.0. This PR omits the fix #628 in favor of fully removing the "apply to a workspace" logic from this resource. The reasoning for removing this attribute as an argument is two-fold:
* Using the `workspace_ids` argument would override any existing workspaces that used this variable set. 
* The logic for creating a variable set was tightly coupled with applying it to a workspace(s) leading to some wonky behavior as noted in #628 . For ease of mind, these two operations should be separate to avoid any weird behavior. You can also read more in #537 

`workspace_ids` will remain as a computed property in the resource. The recommended method to apply a variable set to a workspace remains `r/tfe_workspace_variable_set`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
- * r/tfe_workspace: Changes in `agent_pool_id` and `execution_mode` attributes are now detected and applied. ([#607](https://github.com/hashicorp/terraform-provider-tfe/pull/607))
+
+BREAKING CHANGES:
+* `r/tfe_variable_set`: Removed deprecated argument `workspace_ids` for applying a variable set to workspace, use `r/tfe_workspace_variable_set` instead. ([#629](https://github.com/hashicorp/terraform-provider-tfe/pull/629))
 
 FEATURES:
 * r/tfe_workspace_run_task, d/tfe_workspace_run_task: Add `stage` attribute to workspace run tasks. ([#555](https://github.com/hashicorp/terraform-provider-tfe/pull/555))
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
- * r/tfe_workspace: Changes in `agent_pool_id` and `execution_mode` attributes are now detected and applied. ([#607](https://github.com/hashicorp/terraform-provider-tfe/pull/607))
+
+BREAKING CHANGES:
```
annawinkler, line 3: If this is a breaking change, is it time for a major version bump? cc @brandonc

barrettclark: The provider is still pre-1.0 so it wouldn't be a major version bump

tfe/resource_tfe_variable_set_test.go
```diff
@@ -11,67 +11,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-func TestAccTFEVariableSet_basic(t *testing.T) {
```
brandonc, line 14: Why are all these tests removed?

brandonc (changes requested):
To be honest I was hoping to leave the deprecation in for a few month months. It's only been two. Can we stash this for a future release?

tfe/resource_tfe_variable_set_test.go
sebasslash: Most of the tests here were specifically for using `workspace_ids` as an argument so I removed them. I can add another test step which uses a config that creates a `tfe_workspace_variable_set` and checks `workspace_ids`.

CHANGELOG.md
sebasslash: Yep! This also wouldn't be the first time we introduce a breaking change, see v0.32.0 or v0.35.0 :heart:

sebasslash:
I wouldn't mind holding off for now, but I'm curious as to why you'd want to maintain the deprecated status for another few months?

If we _do_ hold off we'll need to merge #628 :+1:

brandonc:
@sebasslash All deprecations need a healthy period of warning. Our official policy is no fewer than 6 months. I can provide links but not in GH comment.

PR #635, opened 2022-09-28
add datasource tfe_organization_members

Uk1288:
## Description

Add datasource `tfe_organization_members` which returns a list of active members and members with pending invite in an organization.

## Testing plan

1.  _Use config with:_
``` 
resource "tfe_organization" "bar" {
   name = "org-bar"
   email = "user@hashicorp.com"
}

data "tfe_organization_members" "foo" {
  organization = tfe_organization.bar.name
}

 ```
1.  _a list of members and members_waiting in the org should be returned_
```
{
  "id" = "my-org"
  "members" = tolist([
    {
      "organization_membership_id" = "ou-id"
      "user_id" = "user-id"
    },
  ])
  "members_waiting" = tolist([
    {
      "organization_membership_id" = "ou-id"
      "user_id" = "user-id"
    }
  ])
  "organization" = "my-org"
}

```

## External links

- [go-tfe documentation](https://pkg.go.dev/github.com/hashicorp/go-tfe#OrganizationMembership)
- [API documentation](https://www.terraform.io/cloud-docs/api-docs/organization-memberships)

```

...
```

Screen Shot 2022-10-05 at 10 35 03 AM

### DIFF
--- a/tfe/data_source_organization_members.go
+++ b/tfe/data_source_organization_members.go
@@ -58,9 +58,8 @@ func dataSourceTFEOrganizationMembersRead(d *schema.ResourceData, meta interface
 	tfeClient := meta.(*tfe.Client)
 
 	organizationName := d.Get("organization").(string)
-	options := tfe.OrganizationMembershipListOptions{Include: []tfe.OrgMembershipIncludeOpt{tfe.OrgMembershipUser}}
 
-	members, membersWaiting, err := fetchOrganizationMembers(tfeClient, organizationName, options)
+	members, membersWaiting, err := fetchOrganizationMembers(tfeClient, organizationName)
 	if err != nil {
 		return err
 	}
--- a/tfe/data_source_organization_members.go
+++ b/tfe/data_source_organization_members.go
@@ -25,7 +25,7 @@ func dataSourceTFEOrganizationMembers() *schema.Resource {
 							Computed: true,
 						},
 
-						"membership_id": {
+						"organization_membership_id": {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
@@ -43,7 +43,7 @@ func dataSourceTFEOrganizationMembers() *schema.Resource {
 							Computed: true,
 						},
 
-						"membership_id": {
+						"organization_membership_id": {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
@@ -58,7 +58,7 @@ func dataSourceTFEOrganizationMembersRead(d *schema.ResourceData, meta interface
 	tfeClient := meta.(*tfe.Client)
 
 	organizationName := d.Get("organization").(string)
-	options := &tfe.OrganizationMembershipListOptions{Include: []tfe.OrgMembershipIncludeOpt{tfe.OrgMembershipUser}}
+	options := tfe.OrganizationMembershipListOptions{Include: []tfe.OrgMembershipIncludeOpt{tfe.OrgMembershipUser}}
 
 	members, membersWaiting, err := fetchOrganizationMembers(tfeClient, organizationName, options)
 	if err != nil {
--- a/tfe/data_source_organization_members.go
+++ b/tfe/data_source_organization_members.go
@@ -1,0 +1,73 @@
+package tfe
+
+import (
+	"github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceTFEOrganizationMembers() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTFEOrganizationMembersRead,
+
+		Schema: map[string]*schema.Schema{
+			"organization": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"members": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"user_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"membership_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"members_waiting": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"user_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"membership_id": {
+							Type:     schema.TypeString,
+							Computed: true,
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: tfe/organization_members_helpers.go [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		for _, orgMembership := range organizationMembershipList.Items {
+			if orgMembership.Status == tfe.OrganizationMembershipActive {
+				member := map[string]string{"user_id": orgMembership.User.ID, "membership_id": orgMembership.ID}
+				members = append(members, member)
+			} else if orgMembership.Status == tfe.OrganizationMembershipInvited {
+				member := map[string]string{"user_id": orgMembership.User.ID, "membership_id": orgMembership.ID}
+				membersWaiting = append(membersWaiting, member)
+			} else {
```
sebasslash, line 27: I love that you sort this distinction in membership status 👍

tfe/organization_members_helpers.go
```diff
@@ -0,0 +1,41 @@
+package tfe
+
+import (
+	"fmt"
+	"log"
+
+	tfe "github.com/hashicorp/go-tfe"
+)
+
+func fetchOrganizationMembers(client *tfe.Client, orgName string, options *tfe.OrganizationMembershipListOptions) ([]map[string]string, []map[string]string, error) {
```
sebasslash, line 10: I think `options` should be defined locally here rather than as a parameter. Suppose the caller of `fetchOrganizationMembers()` mutates `options` ... could lead to some wonky behavior here. Alternatively, if you do want the caller to be able to define `options`, it should be passed by value.

tfe/testing.go
```diff
@@ -140,6 +140,22 @@ func createOrganization(t *testing.T, client *tfe.Client, options tfe.Organizati
...
+		t.Fatal(err)
+	}
+
+	return orgMembership, func() {
+		if err := client.OrganizationMemberships.Delete(ctx, orgMembership.ID); err != nil {
+			t.Errorf("Error destroying organization membership! WARNING: Dangling resources\n"+
+				"may exist! The full error is show below:\n\n"+
+				"Organization memberships:%s\nError: %s", orgMembership.ID, err)
+		}
+	}
```
sebasslash, line 156: I'm curious if we can get away with registering the cleanup function inside the helper rather than returning it:
```go
t.Cleanup(func() {
    // delete orgMembership
})
return orgMembership
```

website/docs/d/organization_members.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+# Data Source: tfe_organization_members
+
+Use this data source to get information about members of an organization.
+
+## Example Usage
+
+```hcl
+data "tfe_organization_members" "foo" {
+  organization = "organization-name"
+}
```
sebasslash, line 18: 💅 Typically in our examples we also include defining the resources our documented resource would depend on:

```hcl
resource "tfe_organization" "deathstar" {
   name = "deathstar"
   email = "sheev.palpatine@hashicorp.com"
}

data "tfe_organization_members" "stormtroopers" {
  organization = tfe_organization.deathstar.name
}
```
😄

sebasslash (changes requested):
All in all, phenomenal work 🚀 ! I've requested some minor changes, but we can discuss them if you have other thoughts! Excited for this change 😁 

Smoke tested ✅

sebasslash:
I would like this to be released in conjunction with #617 💜

website/docs/d/organization_members.html.markdown
Uk1288: good point!

tfe/organization_members_helpers.go
Uk1288: right, to make it flexible for callers of this function to pass in respective options, I will update the option to pass by value.

tfe/testing.go
Uk1288: yes, this should work as well

website/docs/d/organization_members.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - Name of the organization.
+* `members` - A list of active members of the organization.
+* `members_waiting` - A list of members with pending invite to organization.
+
+The `member` block contains:
+
+* `user_id` - The ID of the user.
+* `membership_id` - The ID of the membership.
```
brandonc, line 42: Elsewhere in the provider, we call this the `organization_membership_id` (`tfe_team_organization_member` resource) and colloquially "organization membership id" in other places (`tfe_organization_membership` resource and data source)

I think we should reuse this long name so we don't risk id confusion. WDYT?

Uk1288: this sounds reasonable to me, I will update the name to match.

tfe/data_source_organization_members.go
```diff
@@ -0,0 +1,73 @@
...
+			},
+		},
+	}
+}
+
+func dataSourceTFEOrganizationMembersRead(d *schema.ResourceData, meta interface{}) error {
+	tfeClient := meta.(*tfe.Client)
+
+	organizationName := d.Get("organization").(string)
+	options := tfe.OrganizationMembershipListOptions{Include: []tfe.OrgMembershipIncludeOpt{tfe.OrgMembershipUser}}
```
brandonc, line 61: Since we require only the user ID, there is no need to include users because the normal result only returns the ID. This should make fetching them faster since no extra user data has to be transmitted unnecessarily.

tfe/organization_members_helpers.go
```diff
@@ -0,0 +1,41 @@
+package tfe
+
+import (
+	"fmt"
+	"log"
+
+	tfe "github.com/hashicorp/go-tfe"
+)
+
+func fetchOrganizationMembers(client *tfe.Client, orgName string, options tfe.OrganizationMembershipListOptions) ([]map[string]string, []map[string]string, error) {
```
brandonc, line 10: I'm feeling like this function shouldn't accept an arbitrary ListOptions parameter. Reason being, you always want to start at page 1 and return all pages, plus the shape of the return value determines which include params you need in the first place.

brandonc (changes requested):

tfe/data_source_organization_members.go
Uk1288: Nice, I didn't realize user ID comes by default. 🎉

tfe/organization_members_helpers.go
Uk1288: I don't have a strong opinion about this one so I'll update it

brandonc (approved):

sebasslash (approved):

PR #641, opened 2022-10-03
Workspace not updated when assessments_enabled is the only change

brandonc:
## Description

We neglected to add this argument to the list of changes to probe for when updating a workspace. If assessments_enabled is the only change in a workspace, the workspace will not be updated at all.

## Testing plan

Change assessments_enabled on a tfe_workspace resource (and no other argument) and apply the change. It should now be updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+BUG FIXES:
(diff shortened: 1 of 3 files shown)


laurenolivia (approved):
🎉

PR #643, opened 2022-10-03
⬆️ Use go 1.18 when building/testing in circle ci

annawinkler:
## Description

Other 3rd party libraries are using 1.18 so let's get our build system also using 1.18.

## Testing plan

1. The circle ci tests should still run.

## External links

N/A

N/A


laurenolivia (approved):
🎉

PR #645, opened 2022-10-04
Enable suppressed linters and fix issues

laurenolivia:
## Description

Undo suppressed linters and resolve the warnings and/or errors.

_Remember to:_

- _Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/CONTRIBUTING.md#updating-the-changelog)_

- _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/CONTRIBUTING.md#updating-the-documentation)_

## Testing plan

1.  Run `golangci-lint run` locally.

## External links

- [golangci-lint docs](https://golangci-lint.run/usage/quick-start/)

Nothing should return if there are no errors.

```

...
```


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
+          cache: true
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        run: golangci-lint run
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.29
```
annawinkler, line 26: What do you think about using the latest version of the linter?

laurenolivia: I think it makes sense!

sebasslash (commented):
:wave: One thing I'd like to call out is our [existing lint workflow](https://github.com/hashicorp/terraform-provider-tfe/blob/main/.github/workflows/lint.yml). So this workflow feels like a duplicate and so we should instead update (or delete) the original one. One caveat that isn't particularly obvious is that under this current workflow we won't see what line numbers are causing the linter errors. We resolved this in the original one passing a [specific output argument](https://github.com/hashicorp/terraform-provider-tfe/blob/e32bf6a86eb452b1f1f6faf24efa497e3ff8bd53/.github/workflows/lint.yml#L24). 

More importantly, I think our attention should be on .golangci.yml and the [exclude-rules section](https://github.com/hashicorp/terraform-provider-tfe/blob/e32bf6a86eb452b1f1f6faf24efa497e3ff8bd53/.golangci.yml#L26-L162). This basically _turns off_ specific linter rules. @Uk1288 can clarify here: I think our goal was to turn back on and resolve them correct?

Uk1288:
> 👋 One thing I'd like to call out is our [existing lint workflow](https://github.com/hashicorp/terraform-provider-tfe/blob/main/.github/workflows/lint.yml). So this workflow feels like a duplicate and so we should instead update (or delete) the original one. One caveat that isn't particularly obvious is that under this current workflow we won't see what line numbers are causing the linter errors. We resolved this in the original one passing a [specific output argument](https://github.com/hashicorp/terraform-provider-tfe/blob/e32bf6a86eb452b1f1f6faf24efa497e3ff8bd53/.github/workflows/lint.yml#L24).
> 
> More importantly, I think our attention should be on .golangci.yml and the [exclude-rules section](https://github.com/hashicorp/terraform-provider-tfe/blob/e32bf6a86eb452b1f1f6faf24efa497e3ff8bd53/.golangci.yml#L26-L162). This basically _turns off_ specific linter rules. @Uk1288 can clarify here: I think our goal was to turn back on by removing them and resolve them correct?


@sebasslash yes, this is correct. For each line under the referenced section, when you remove the excluded line and rerun the linter, the errors will surface.

laurenolivia:
@sebasslash Yep, the end goal is to enable certain linters. I moved this PR over to review because it could benefit from feedback.

laurenolivia:
I'm open to a discussion about lint.yml versus golangci-lint.yml, but the sample docs reference the latter.

laurenolivia:
I synced with @Uk1288 and she clarified her exclude rules. So I am moving this back to in-progress and will mark this as `draft`.

laurenolivia:
Want to add that go-tfe uses workflows/[golangci-lint.yml](https://github.com/hashicorp/go-tfe/blob/main/.github/workflows/golangci-lint.yml)

annawinkler:
I looked up the linter, and the [official](https://github.com/marketplace/actions/run-golangci-lint) github workflow also uses what this PR suggests, so I'm good with that! It might be helpful for us to refer to the How to Use [section](https://github.com/marketplace/actions/run-golangci-lint#how-to-use) and make sure we're in alignment with what is included there.

annawinkler:
> Want to add that go-tfe uses workflows/[golangci-lint.yml](https://github.com/hashicorp/go-tfe/blob/main/.github/workflows/golangci-lint.yml)

💯 This is also a great point - having consistency across our repos reduces friction and makes things easier for the maintainers (us!).

.golangci.yml
```diff
@@ -42,50 +43,50 @@ issues:
...
       - errcheck
     path: \.go
-  - linters:
-      - gocritic
-    path: config_unix\.go
-    text: "importShadow: shadow of imported package 'user'"
+  # - linters:
+  #     - gocritic
+  #   path: config_unix\.go
+  #   text: "importShadow: shadow of imported package 'user'"
```
brandonc, line 49: Let's remove all these lines instead of commenting them.

tfe/plugin_provider_test.go
```diff
@@ -55,14 +57,18 @@ func TestPluginProvider_providerMeta(t *testing.T) {
 			"hostname":        tftypes.NewValue(tftypes.String, tc.hostname),
 			"token":           tftypes.NewValue(tftypes.String, tc.token),
 			"ssl_skip_verify": tftypes.NewValue(tftypes.Bool, tc.sslSkipVerify),
-		}))
+		}),
+		)
+		if err != nil {
+			log.Printf("error: %v", err)
```
brandonc, line 63: Is this error fatal? I assume so since it would indicate that we passed an invalid value to NewDynamicValue ?

```suggestion
			t.Fatalf("Failed to create test config: %v", err)
```

tfe/provider.go
```diff
@@ -183,7 +183,7 @@ func getClient(tfeHost, token string, insecure bool) (*tfe.Client, error) {
 	// Make sure the transport has a TLS config.
 	transport := httpClient.Transport.(*http.Transport)
 	if transport.TLSClientConfig == nil {
-		transport.TLSClientConfig = &tls.Config{}
+		transport.TLSClientConfig = &tls.Config{MinVersion: tls.VersionTLS13}
```
brandonc, line 186: Do you happen to know what TLS versions are in use in TFC/older versions of TFE? How did you decide on this min version?

brandonc (commented):
This is a good looking cleanup. Just a few questions

tfe/provider.go
laurenolivia: By not specifying a `MinVersion`, it was defaulting to a version that created a security vulnerability. This min version was the latest per the docs.

laurenolivia: I do not know what TLS versions are in use in TFC/older versions of TFE, but I can look into it.

tfe/plugin_provider_test.go
laurenolivia: According to the [docs](https://pkg.go.dev/testing#B.Fatalf) `Fatalf` is equivalent to `Logf`, analogous to `Printf`.

tfe/data_source_oauth_client.go
```diff
@@ -128,7 +129,7 @@ func dataSourceTFEOAuthClientRead(d *schema.ResourceData, meta interface{}) erro
 	case 1:
 		d.Set("oauth_token_id", oc.OAuthTokens[0].ID)
 	default:
-		return fmt.Errorf("Unexpected number of OAuth tokens: %d", len(oc.OAuthTokens)) // nolint:golint,errorlint
+		return fmt.Errorf("unexpected number of OAuth tokens: %d", len(oc.OAuthTokens)) // nolint:golint,errorlint
```
brandonc, line 132: You can remove the nolint comment now, I think?

tfe/data_source_outputs.go
```diff
@@ -18,14 +17,6 @@ type dataSourceOutputs struct {
 	tfeClient *tfe.Client
 }
 
-var stderr *os.File
-
-func init() {
-	// There is a issue that occurs when the plugin-go Serve function is used that
-	// causes os.Stderr to be overwritten. There is a fix being worked on for this.
-	stderr = os.Stderr
-}
```
brandonc, line 27: Did you verify that the fix was in place in plugin-go for this?

tfe/plugin_provider_test.go
brandonc: Right, but I guess my question is shouldn't the test run stop instead of just log if an error happens here?

tfe/provider.go
brandonc: 🛑 The issue is that we could require a version greater than what versions of TFE would be capable of. We should either stick with the default (12) and create a linter exception or verify that version 13 is compatible with TFE versions going way back

tfe/provider_test.go
```diff
@@ -286,23 +286,35 @@ func testAccPreCheck(t *testing.T) {
...
+func init() {
+	githubToken = os.Getenv("GITHUB_TOKEN")
+	githubPolicySetIdentifier = os.Getenv("GITHUB_POLICY_SET_IDENTIFIER")
+	githubPolicySetBranch = os.Getenv("GITHUB_POLICY_SET_BRANCH")
+	githubPolicySetPath = os.Getenv("GITHUB_POLICY_SET_PATH")
+	githubRegistryModuleIdentifer = os.Getenv("GITHUB_REGISTRY_MODULE_IDENTIFIER")
+	githubWorkspaceIdentifier = os.Getenv("GITHUB_WORKSPACE_IDENTIFIER")
+	githubWorkspaceBranch = os.Getenv("GITHUB_WORKSPACE_BRANCH")
+	tfeUser1 = os.Getenv("TFE_USER1")
+	tfeUser2 = os.Getenv("TFE_USER2")
```
brandonc, line 309: 💅 Not a blocking problem, but some of these strings are common enough to be shadowed by locally scoped variables in the future. I think it would be both more safe and clear to rename all these envGithubToken, envTFEUser1, etc

brandonc (commented):
Nice! The only remaining issue I have is the question about the new minimum TLS version.

tfe/plugin_provider_test.go
laurenolivia: That's my understanding of the `Fatalf` method. It exits immediately after the test failure message is generated. Unlike `Printf` or `Errorf`.

You do agree the test should fail and exit, correct?

tfe/provider.go
laurenolivia: Thanks for raising this. I did some additional reading about the two versions and `TLS12` remains in widespread use.

tfe/provider_test.go
laurenolivia: Good idea 💡

tfe/data_source_outputs.go
laurenolivia: The flag here is that `var stderr *os.File` is unused. We chatted about this briefly back in Oct and @nfagerlund offered some context around this block of code:

```
last place stderr occurred anywhere in this repo was right before we switched to using go modules 
and removed all the vendored library code
```

Am I missing any additional context?

nfagerlund: I couldn't sort out how to trace this vague comment back to a specific issue (my ace in the hole was to ask Omar, but I think he couldn't remember either). But even if the issue still existed, we're not ever _doing_ anything with that variable after setting it, so it doesn't seem likely that this would, on its own, mitigate said issue. 

[This is the only PR](https://github.com/hashicorp/terraform-plugin-go/pull/93) in terraform-plugin-go that seems relevant, but again, identifying the exact issue doesn't seem critical if we already weren't actually _doing_ anything to work around it.

brandonc (approved):

Issue #637, opened 2022-09-29
Problem with credentials_helper


chris3ware:
#### Terraform version

```plaintext
Terraform v1.3.0
on darwin_amd64
```

#### Terraform Configuration Files

```zsh
➜ ~ lsa ~/.terraform.d/plugins

Permissions Size User  Date Modified Name
.rwxr-xr-x  8.6M chris 28 Sep 12:18  terraform-credentials-keychain


➜ ~ cat ~/.terraformrc

credentials_helper "keychain" {
  args = ["--keychain=terraform"]
}
```

```terraform
provider "tfe" {
}

terraform {
  required_version = ">= 1.3.0"
  required_providers {
    github = {
      source  = "integrations/github"
      version = "~> 5.2.0"
    }

    tfe = {
      version = "~> 0.36.1"
    }
    sops = {
      source  = "carlpett/sops"
      version = "~> 0.7.1"
    }
  }

  backend "remote" {
    organization = "3ware"
    workspaces {
      name = "3ware-repoman"
    }
  }
}

resource "tfe_organization" "threeware" {
  name  = "3ware"
  email = "<EMAIL>"
}

resource "tfe_workspace" "this" {
  for_each          = { for workspace in var.workspaces : workspace.name => workspace }
  name              = each.value.name
  organization      = tfe_organization.threeware.name
  description       = each.value.description
  working_directory = try(each.value.working_directory, "terraform")
  execution_mode    = try(each.value.execution_mode, "local")
  vcs_repo {
    identifier     = "${tfe_organization.threeware.name}/${each.value.repo}"
    oauth_token_id = data.sops_file.token.data["oauth_token_id"]
  }
}
```

#### Debug Output

```plaintext
https://gist.github.com/chris3ware/2a471834152440ad2aa59e85773aea5a
```

#### Expected Behavior

TFE provider should retrieve the terraform cloud API key from masOS key chain using the credentials_helper configuration in ~/.terraformrc to refresh state of terraform cloud resources.

#### Actual Behavior

Terraform plan fails because it is unable to refresh the state of the terraform cloud resources.

#### Additional Context

I am using terraform cloud for remote state workspaces in local execution mode.

I also raised an issue on the credentials_helper repo: bendrucker/terraform-credentials-keychain#38. 

The provider works as expected when the terraform cloud api token is configured in the provider configuration, as an environment variable or as credentials in ~/.terraformrc.

sebasslash:
[...]

PR #658, opened 2022-10-13
docs: mention non-support for `credential_helper`

bendrucker:
## Description

Adds documentation that a `credentials_helper` in the user's Terraform configuration is not a supported authentication mechanism. While it's supported by the package used for fetching authn (`terraform-svchost`), only static tokens are implemented currently by this provider.

https://github.com/hashicorp/terraform-svchost/blob/master/auth/helper_program.go

Adding support for credentials helpers is non-trivial. The package handles the protocol for calling these helpers (args) but not things like the directory where helper executables are found. Given the effort involved in implementing this and the fact that it would only work in workspaces in local execution mode, documenting the limitation rather than addressing it seems like the best course.

## External links

* #637 
* https://github.com/bendrucker/terraform-credentials-keychain/issues/38#issuecomment-1277030030


sebasslash (approved):
Thank you for this :fire:

PR #660, opened 2022-10-14
Add support for fetching organization membership by username

sebasslash:
## Description

This PR adds a new attribute `username` to `d/tfe_organization_membership`, which allows for org memberships to be query-able by username. `email` is no longer a required attribute and it is now marked as optional given you can supply a name or email. While we don't enforce that one or the other be defined at the schema level, the provider will error if both are not defined or `""`. 

~This is a WIP as these changes rely on a unreleased changes to go-tfe, see [related pull request](https://github.com/hashicorp/go-tfe/pull/550).~

This PR also adds the same attribute `username` as a computed property to `r/tfe_organization_membership`.


tfe/organization_members_helpers.go
```diff
@@ -40,3 +41,61 @@ func fetchOrganizationMembers(client *tfe.Client, orgName string) ([]map[string]
...
+		// We check this just in case a user's TFE instance only has one organization member
+		if user.Email != email && user.Username != username {
+			return nil, tfe.ErrResourceNotFound
+		}
+
+		return oml.Items[0], nil
+	default:
+		for {
+			for _, member := range oml.Items {
+				if member.User.Email == email || member.User.Username == username {
```
brandonc, line 82: This works fine, but I would appreciate a little guardrail around this condition to reinforce the relationship with the parameters, in case the username is ever a blank string instead of null.

```suggestion
				if (len(email) > 0 && member.User.Email == email) || (len(username) > 0 && member.User.Username == username) {
```

brandonc (changes requested):
My only blocking feedback is that we should populate username in the state, even when reading using email. My rationale is that this data source reuses `resourceTFEOrganizationMembershipRead` and that is the behavior of email when fetching by username.

That may mean adding username to the resource schema, which creates a nice symmetry and is free because we already include user data.

tfe/organization_members_helpers.go
sebasslash: Great suggestion, will add.

brandonc (approved):
Yes

PR #667, opened 2022-10-21
Update tflocal module to v0.8, remove token/password vars

sebasslash:
## Description

PR says it all, the latest version of tflocal-cloud is v0.8. I've also removed the token/password variables to let the module seed these values.


Uk1288 (approved):
LGTM

PR #668, opened 2022-10-21
Add missing artifactory, datadog vars

sebasslash:
## Description

Required artifactory and DD variables added to module that I missed in our last PR. I should probably enable spec plans...


annawinkler (approved):
:shipit:

PR #670, opened 2022-10-24
Update test docs to mention distinction between Makefile test targets

sebasslash:
## Description

The reason for this change stems from the lack of clarity behind why we make the distinction between `make test` and `make testacc`. At the surface level, the singular difference between the two targets is whether `TF_ACC` is set or not. This PR adds a a description to our `Run the tests` section explaining in a bit more detail as to why that is, as well as attempting to provide a simple definition for what sort of test is classified as an acceptance test and which are classified as unit tests. I've also reversed the ordering to display how to run acceptance tests **first**, since they are most common type on test people will write for the provider.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -11,6 +11,7 @@ build: fmtcheck
 terraform-provider-tfe: fmtcheck
 	@go build -o terraform-provider-tfe
 
+# Run unit tests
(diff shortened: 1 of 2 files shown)


annawinkler (approved):
💯 Thank you for these updates!

PR #673, opened 2022-10-26
Introduce no-code modules in registry modules

miguelhrocha:
## Description

This PR is a follow-up of https://github.com/hashicorp/go-tfe/pull/562, which adds the `no_code` module attribute to the registry module resource. This will enable practitioners to bootstrap modules in their TFE instance set up for no-code workspace provisioning.

This change is dependent on the `go-tfe` PR mentioned before, once that merge this PR will be ready for review.

## External links

- [go-tfe no-code update](https://github.com/hashicorp/go-tfe/pull/562)

```
TESTARGS="-run TestAccTFERegistryModule_noCodeModule" envchain tfe-provider make testacc
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFERegistryModule_noCodeModule -timeout 15m
=== RUN   TestAccTFERegistryModule_noCodeModule
--- PASS: TestAccTFERegistryModule_noCodeModule (13.89s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	14.235s
```

## Screenshots
Screenshot 2022-10-26 at 13 58 51


tfe/resource_tfe_registry_module.go
```diff
@@ -76,6 +77,12 @@ func resourceTFERegistryModule() *schema.Resource {
 				ForceNew:     true,
 				RequiredWith: []string{"registry_name"},
 			},
+			"no_code": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
```
sebasslash, line 84: We don't need to specify `ForceNew: false` here

tfe/resource_tfe_registry_module.go
```diff
@@ -190,6 +198,41 @@ func resourceTFERegistryModuleCreate(d *schema.ResourceData, meta interface{}) e
...
+	})
+
+	if err != nil {
+		return fmt.Errorf("Error while waiting for module %s/%s to be ingested: %w", registryModule.Organization.Name, registryModule.Name, err)
+	}
+
+	d.SetId(registryModule.ID)
+	d.Set("no_code", registryModule.NoCode)
+
+	return nil
```
sebasslash, line 233: Typically an `Update()` will return with a call to `Read()`. While we aren't explicit about this, `Read()` should be the only method that should be used to write attributes to state (with the exception of the resource ID). 

Line 231 should be added to `Read()` instead.

sebasslash: I did realize that `Create()` writes certain attributes to state, but only because `Read()` depends on these values. Given that `Update()` will only ever be called if the `no_code` attribute is used as an argument (since all other attributes are `ForceNew: true`), we can get away with using `Read()` to write the `no_code` attribute.

sebasslash: And final question, do we have to wait for the module to be ingested if it's not being created?

tfe/resource_tfe_registry_module_test.go
```diff
@@ -230,6 +230,61 @@ func TestAccTFERegistryModule_publicRegistryModule(t *testing.T) {
 						"tfe_registry_module.foobar", "namespace", expectedRegistryModuleAttributes.Namespace),
 					resource.TestCheckResourceAttr(
 						"tfe_registry_module.foobar", "registry_name", string(expectedRegistryModuleAttributes.RegistryName)),
+					resource.TestCheckResourceAttr(
+						"tfe_registry_module.foobar", "no_code", "false"),
```
sebasslash, line 234: From what I mentioned above, this never gets set because `no_code` is only ever written to state during `Update()` and `Update()` is only called when the attribute is set as an argument in your configuration.

website/docs/r/registry_module.html.markdown
```diff
@@ -81,6 +81,7 @@ The following arguments are supported:
 * `organization` - (Optional) The name of the organization associated with the registry module. It must be set if `module_provider` is used.
 * `namespace` - (Optional) The namespace of a public registry module. It can be used if `module_provider` is set and `registry_name` is public.
 * `registry_name` - (Optional) Whether the registry module is private or public. It can be used if `module_provider` is set.
+* `no_code` - (Optional) Whether the registry module is enabled for [no-code provisioning](https://learn.hashicorp.com/tutorials/terraform/no-code-provisioning).
```
sebasslash, line 84: Should probably add: Defaults to `false`. You should also modify the examples to show this argument in use.

sebasslash (changes requested):
We're almost there :smile:  :+1: -- Some blocking feedback below :arrow_down:

tfe/resource_tfe_registry_module.go
miguelhrocha: > And final question, do we have to wait for the module to be ingested if it's not being created?

Evils of copy-pasting 😅 everything has been addressed now 👍🏼

sebasslash (approved):
I've smoked tested :heavy_check_mark: and behaves as expected. Requested changes have also been addressed :+1:

PR #661, opened 2022-10-14
[#TF-720] Add allow_force_delete_workspaces organization setting to the tfe provider

emlanctot:
## Description

This change adds the `allow_force_delete_workspaces` setting to the TFE provider.

## External links

_Include any links here that might be helpful for people reviewing your PR. If there are none, feel free to delete this section._

- [Related PR](https://github.com/hashicorp/atlas/pull/13244)

## Output from acceptance tests

_Please run applicable acceptance tests locally and include the output here. See TESTS.md to learn how to run acceptance tests._

_If you are an external contributor, your contribution(s) will first be reviewed before running them against the project's CI pipeline._

```
/opt/homebrew/Cellar/go/1.19.2/libexec/bin/go tool test2json -t /private/var/folders/2z/33zp571x44v_ndv_hk8cyy1w0000gp/T/GoLand/___1TestAccTFEOrganization_basic_in_github_com_hashicorp_terraform_provider_tfe_tfe.test -test.v -test.paniconexit0 -test.run ^\QTestAccTFEOrganization_basic\E$
=== RUN   TestAccTFEOrganization_basic
--- PASS: TestAccTFEOrganization_basic (8.08s)
PASS

Process finished with the exit code 0



/opt/homebrew/Cellar/go/1.19.2/libexec/bin/go tool test2json -t /private/var/folders/2z/33zp571x44v_ndv_hk8cyy1w0000gp/T/GoLand/___1TestAccTFEOrganization_full_in_github_com_hashicorp_terraform_provider_tfe_tfe.test -test.v -test.paniconexit0 -test.run ^\QTestAccTFEOrganization_full\E$
=== RUN   TestAccTFEOrganization_full
--- PASS: TestAccTFEOrganization_full (7.31s)
PASS

Process finished with the exit code 0

```

PR #675, opened 2022-10-26
[#TF-722] Allow safe and force delete through the tfe provider

emlanctot:
## Description

Adds the `force_delete` attribute to a workspace provider resource. 

## External links

- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/661)

```
/opt/homebrew/Cellar/go/1.19.2/libexec/bin/go tool test2json -t /private/var/folders/2z/33zp571x44v_ndv_hk8cyy1w0000gp/T/GoLand/___TestTFEWorkspace_delete_withoutCanForceDeletePermission_in_github_com_hashicorp_terraform_provider_tfe_tfe.test -test.v -test.paniconexit0 -test.run ^\QTestTFEWorkspace_delete_withoutCanForceDeletePermission\E$
=== RUN   TestTFEWorkspace_delete_withoutCanForceDeletePermission
2022/11/08 09:43:52 [DEBUG] Delete workspace ws-testing
2022/11/08 09:43:52 [DEBUG] Delete workspace ws-testing
--- PASS: TestTFEWorkspace_delete_withoutCanForceDeletePermission (0.37s)
PASS

Process finished with the exit code 0

...


```

### DIFF
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -688,14 +688,26 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error 
 
 	forceDelete := d.Get("force_delete").(bool)
 
-	if ws.Permissions.CanForceDelete == nil && !forceDelete {
-		return fmt.Errorf(
-			"Error deleting workspace %s: This workspace must be force deleted by setting force_delete=true", id)
-	}
-
-	if forceDelete {
-		err = tfeClient.Workspaces.DeleteByID(ctx, id)
+	// presence of Permissions.CanForceDelete will determine if current version of TFE supports safe deletes
+	if ws.Permissions.CanForceDelete == nil {
+		if forceDelete {
+			err = tfeClient.Workspaces.DeleteByID(ctx, id)
+		} else {
+			return fmt.Errorf(
+				"Error deleting workspace %s: This workspace must be force deleted by setting force_delete=true", id)
+		}
+	} else if *ws.Permissions.CanForceDelete {
+		if forceDelete {
+			err = tfeClient.Workspaces.DeleteByID(ctx, id)
+		} else {
+			err = tfeClient.Workspaces.SafeDeleteByID(ctx, id)
+			return errWorkspaceSafeDeleteWithPermission(id, err)
+		}
 	} else {
+		if forceDelete {
+			return fmt.Errorf(
+				"Error deleting workspace %s: missing required permissions to set force delete workspaces in the organization.", id)
+		}
 		err = tfeClient.Workspaces.SafeDeleteByID(ctx, id)
 	}
 
@@ -706,7 +718,6 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error 
 		return fmt.Errorf(
 			"Error deleting workspace %s: %w", id, err)
 	}
-
 	return nil
 }
 
@@ -793,4 +804,13 @@ func resourceTFEWorkspaceImporter(ctx context.Context, d *schema.ResourceData, m
 // Warns the user that a workspace tag containing uppercase characters will be downcased.
 func warnWorkspaceTagsCasing(ctx context.Context, tag string) {
 	log.Printf("[WARN] The tag \"%s\" contains uppercase characters that will be transformed by the API. Please update your configuration to lowercase tags in order to avoid conflicts with state.", tag)
+}
+
+func errWorkspaceSafeDeleteWithPermission(workspaceID string, err error) error {
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "conflict") {
+			return fmt.Errorf("Error deleting workspace %s: %w\nTo delete this workspace without destroying the managed resources, add force_delete = true to the resource config.", workspaceID, err)
+		}
+	}
+	return nil
 }
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -229,6 +229,11 @@ func resourceTFEWorkspace() *schema.Resource {
 					},
 				},
 			},
+			"force_delete": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
@@ -671,7 +676,29 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error 
 	id := d.Id()
 
 	log.Printf("[DEBUG] Delete workspace %s", id)
-	err := tfeClient.Workspaces.DeleteByID(ctx, id)
+
+	ws, err := tfeClient.Workspaces.ReadByID(ctx, id)
+	if err != nil {
+		if err == tfe.ErrResourceNotFound {
+			return nil
+		}
+		return fmt.Errorf(
+			"Error reading workspace %s: %w", id, err)
+	}
+
+	forceDelete := d.Get("force_delete").(bool)
+
+	if ws.Permissions.CanForceDelete == nil && !forceDelete {
+		return fmt.Errorf(
+			"Error deleting workspace %s: This workspace must be force deleted by setting force_delete=true", id)
+	}
+
+	if forceDelete {
+		err = tfeClient.Workspaces.DeleteByID(ctx, id)
+	} else {
+		err = tfeClient.Workspaces.SafeDeleteByID(ctx, id)
+	}
+
 	if err != nil {
 		if err == tfe.ErrResourceNotFound {
 			return nil
--- a/tfe/resource_tfe_workspace_test.go
+++ b/tfe/resource_tfe_workspace_test.go
@@ -1907,7 +1907,7 @@ func TestTFEWorkspace_delete_withoutCanForceDeletePermission(t *testing.T) {
 	// workspace resource delete function directly, rather than use the usual resource.
 
 	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
-	orgName := fmt.Sprintf("test-orgnaization-%d", rInt)
+	orgName := fmt.Sprintf("test-organization-%d", rInt)
 
 	client := testTfeClient(t, testClientOptions{defaultOrganization: orgName})
 	workspace, err := client.Workspaces.Create(ctx, orgName, tfe.WorkspaceCreateOptions{
@@ -1926,9 +1926,8 @@ func TestTFEWorkspace_delete_withoutCanForceDeletePermission(t *testing.T) {
 	}
 
 	err = resourceTFEWorkspaceDelete(rd, client)
-
 	if err == nil {
-		t.Fatalf("Expected an error deleting workspace with CanForceDelete=nil and force_delete=true")
+		t.Fatalf("Expected an error deleting workspace with CanForceDelete=nil and force_delete=false")
 	}
 	expectedErrSubstring := "workspace must be force deleted by setting force_delete=true"
 	if !strings.Contains(err.Error(), expectedErrSubstring) {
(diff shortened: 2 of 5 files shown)


tfe/resource_tfe_workspace.go
```diff
@@ -671,7 +676,29 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error
...
+		if err == tfe.ErrResourceNotFound {
+			return nil
+		}
+		return fmt.Errorf(
+			"Error reading workspace %s: %w", id, err)
+	}
+
+	forceDelete := d.Get("force_delete").(bool)
+
+	if ws.Permissions.CanForceDelete == nil && !forceDelete {
```
brandonc, line 691: This permission could also be false. Am I right that if the two disagree then the API will handle the error and return a detailed message that after the attempt is made?

brandonc (commented):
Not smoke tested yet but this is looking pretty good!

tfe/resource_tfe_workspace.go
emlanctot: this permission can def be false! we're relying on the atlas side to reject the force deletes because we can provide better error messaging and stay consistent with handling permissions on the atlas side as well!

emlanctot:
this PR is currently blocked until it can be rebased against #661 !

tfe/resource_tfe_workspace.go
```diff
@@ -671,7 +676,29 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error
...
+		return fmt.Errorf(
+			"Error deleting workspace %s: This workspace must be force deleted by setting force_delete=true", id)
```
brandonc, line 693: I would find this error to be very helpful in the case when there are managed resources. The API error only mentions doing a destroy run, but that's slightly confusing in the context of the provider config.

Here's what I see:
```
╷
│ Error: Error deleting workspace ws-8ZdBVKREwypBnTpi: conflict
│
│ Workspace is currently managing 1 resources. A destroy run must be applied to remove managed resources before the workspace can be safely deleted
│
│
```

And maybe what I'd like to see if CanForceDelete is true:

```
╷
│ Error: Error deleting workspace ws-8ZdBVKREwypBnTpi: conflict
│
│ Workspace is currently managing 1 resources. A destroy run must be applied to remove managed resources before the workspace can be safely deleted
| 
│ To delete this workspace without destroying the underlying resources, add force_delete = true to the tfe_resource.foo config
│
```

tfe/resource_tfe_workspace_test.go
```diff
@@ -1811,6 +1815,143 @@ func TestAccTFEWorkspace_paginatedRemoteStateConsumers(t *testing.T) {
...
+	})
+}
+
+func TestTFEWorkspace_delete_withoutCanForceDeletePermission(t *testing.T) {
+	// This test checks that workspace deletion works as expected when communicating with TFE servers which do not send
+	// the CanForceDelete workspace permission. To simulate this we use the mock workspaces client and call the
+	// workspace resource delete function directly, rather than use the usual resource.
+
+	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
+	orgName := fmt.Sprintf("test-orgnaization-%d", rInt)
```
brandonc, line 1910:
```suggestion
	orgName := fmt.Sprintf("test-organization-%d", rInt)
```

brandonc (changes requested):
I really appreciate the unit testing you added to delete because it can be difficult to set up the prerequisites for a conflict

I feel like the behavior of force_delete is unique enough to warrant mentioning in the error message because it's a meta argument. What do you think?

This is a great contribution, thanks!

tfe/resource_tfe_workspace.go
JarrettSpiker: I like the additional bit of error text @brandonc suggested

Another thought: this error right here only gets shown when talking to an old TFE and the user tries to delete a workspace without `force_delete=true`

If the user is talking to an up-to-date TFE/TFC and they try to delete a RUMful workspace without `force_delete=true`, they would just see the error returned by the API (the "Workspace is currently managing 1 resource..." bit)...I wonder if, in [this error handling block ](https://github.com/hashicorp/terraform-provider-tfe/pull/675/files#diff-e727ae216721eb3c1f24911e8b1ececae3ae9208f19b886eb627c76236471e13R702), we should detect if:
- it was a safe delete which was attempted and,
- the error is a `conflict`, and
- CanForceDelete is true

in which case we could also append Brandon's suggested error text (the "To delete this workspace without destroying the managed resources..." bit) to the outputted error

tfe/resource_tfe_workspace.go
```diff
@@ -671,15 +676,54 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error
...
+			}
+		}
+	} else {
+		if forceDelete {
+			return fmt.Errorf(
+				"Error deleting workspace %s: missing required permissions to set force_delete=true", id)
+		} else {
+			err = tfeClient.Workspaces.SafeDeleteByID(ctx, id)
+		}
 	}
```
SwiftEngineer, line 718: Disclaimer: I'm still learning Golang.

The number of nested `if` statements here is a little high, right? I think one way I think we could trim them down is to move the [error handling at the bottom](https://github.com/hashicorp/terraform-provider-tfe/pull/675/files#diff-e727ae216721eb3c1f24911e8b1ececae3ae9208f19b886eb627c76236471e13R720-R726) into a helper function. That way you could take advantage of some early returns in a few places, for example, instead of:

```go
if ws.Permissions.CanForceDelete == nil {
  // ... logic for old tfe...
} else {
  // ... logic for new tfe...
}

if err {
  // common error handling...
}
```

you could do:

```go
if ws.Permissions.CanForceDelete == nil {
  // ... logic for old tfe...
  return commonErrorHandlingHelper(err)
}

// ... logic for new tfe...
return commonErrorHandlingHelper(err)
```

Just an idea 🤙 I'm sure there are even better ways than this to trim things down you or other folks more familiar with golang than me might have

tfe/resource_tfe_workspace.go
```diff
@@ -671,15 +676,54 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error
...
+						"Error deleting workspace %s: %w\nTo delete this workspace without destroying the managed resources, add force_delete=true to the resource config", id, err)
+				}
+			}
+		}
+	} else {
+		if forceDelete {
+			return fmt.Errorf(
+				"Error deleting workspace %s: missing required permissions to set force_delete=true", id)
+		} else {
+			err = tfeClient.Workspaces.SafeDeleteByID(ctx, id)
```
SwiftEngineer, line 716: I was about to comment that I thought this error needs the same handling [as the one here](https://github.com/hashicorp/terraform-provider-tfe/pull/675/files#diff-e727ae216721eb3c1f24911e8b1ececae3ae9208f19b886eb627c76236471e13R703), but I caught myself at the last second 😂 

Holy %$#% this stuff is complex, major props for untangling it all! I couldn't find a single flaw with any of these cases, even after a triple check 🙇

SwiftEngineer (commented):
I only had one suggestion about ways of unnesting some `if` statements, the actual logic in this PR is probably the only thing I feel confident in checking and it seems like you nailed it to me 💯 !

tfe/resource_tfe_workspace.go
```diff
@@ -671,15 +676,54 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error
...
+			return fmt.Errorf(
+				"Error deleting workspace %s: This workspace must be force deleted by setting force_delete=true", id)
+		}
+	} else if *ws.Permissions.CanForceDelete {
+		if forceDelete {
+			err = tfeClient.Workspaces.DeleteByID(ctx, id)
+		} else {
+			err = tfeClient.Workspaces.SafeDeleteByID(ctx, id)
+			if err != nil {
+				if strings.Contains(err.Error(), "conflict") {
```
JarrettSpiker, line 705: If possible, I might go with a `strings.HasPrefix(err.Error(), "conflict")` here, just to be more selective in the string matching.... just so we dont end up matching against something like "The workspace i-put-conflict-in-the-name-of-my-workspace doesnt exist"

tfe/resource_tfe_workspace.go
```diff
@@ -671,15 +676,54 @@ func resourceTFEWorkspaceDelete(d *schema.ResourceData, meta interface{}) error
...
+					return fmt.Errorf(
+						"Error deleting workspace %s: %w\nTo delete this workspace without destroying the managed resources, add force_delete=true to the resource config", id, err)
+				}
+			}
+		}
+	} else {
+		if forceDelete {
+			return fmt.Errorf(
+				"Error deleting workspace %s: missing required permissions to set force_delete=true", id)
```
JarrettSpiker, line 714: 💡 this is a little nitpicky, but the user **does** have permission to set force_delete=true (because adding that to  the config doest delete the workspace)...they just dont have permission to remove a workspace which has `force_delete=true`

I think we could be more explicit here and say something like `Error deleting workspace %s: missing required permissions to set force delete workspaces in the %s organization`

tfe/resource_tfe_workspace.go
emlanctot: that makes total sense! this felt wrong, thanks for pointing me to `hasPrefix` !

tfe/resource_tfe_workspace.go
emlanctot: i feel the same way about the nesting here -- i'll give this a try !

tfe/resource_tfe_workspace.go
```diff
@@ -767,3 +806,12 @@ func resourceTFEWorkspaceImporter(ctx context.Context, d *schema.ResourceData, m
 func warnWorkspaceTagsCasing(ctx context.Context, tag string) {
 	log.Printf("[WARN] The tag \"%s\" contains uppercase characters that will be transformed by the API. Please update your configuration to lowercase tags in order to avoid conflicts with state.", tag)
 }
+
+func errWorkspaceSafeDeleteWithPermission(workspaceID string, err error) error {
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "conflict") {
+			return fmt.Errorf("Error deleting workspace %s: %w\nTo delete this workspace without destroying the managed resources, add force_delete=true to the resource config.", workspaceID, err)
```
brandonc, line 813:
```suggestion
			return fmt.Errorf("Error deleting workspace %s: %w\nTo delete this workspace without destroying the managed resources, add force_delete = true to the resource config.", workspaceID, err)
```

Might as well show our best guess at a proper `terraform fmt` property 😬

brandonc (approved):
Great! Thanks. I think the linter issue looks straightforward as well.

tfe/resource_tfe_workspace.go
emlanctot: yes ! a great idea

ktham:
This is great, if we can cut a release soon, that would be super awesome!

tfe/resource_tfe_workspace.go
```diff
@@ -767,3 +805,12 @@ func resourceTFEWorkspaceImporter(ctx context.Context, d *schema.ResourceData, m
...
 }
+
+func errWorkspaceSafeDeleteWithPermission(workspaceID string, err error) error {
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "conflict") {
+			return fmt.Errorf("Error deleting workspace %s: %w\nTo delete this workspace without destroying the managed resources, add force_delete = true to the resource config.", workspaceID, err)
+		}
+	}
+	return nil
+}
```
skeggse, line 816: the logic here seems wrong; if there's an error, and it's not a conflict, this function should still return the `err` value. i saw this behavior today when i destroyed a terraform cloud workspace, and the workspace wasn't actually deleted but the apply succeeded and the resource was removed from state.

Issue #448, opened 2022-03-22
Get resource_count information for configured tfe_workspace resources


lucymhdavies:
#### Use-cases
When reading a workspace with the `tfe_workspace` data source, it is possible to identify how many resources that workspace is currently managing.
See https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/data-sources/workspace#resource_count

However, when we create the workspace with the `tfe_workspace` resource, we do not get that information back in the attributes.

While it is already possible to write a sentinel policy that prevents deletion of workspaces:
https://github.com/hashicorp/terraform-sentinel-policies/pull/3
Not having the `resource_count` attribute available in the `tfe_workspace` resource prevents us from writing a more complex sentinel policy, along the lines of "do not allow deletion of a workspace, unless it currently does not manage any resources"

#### Proposal
As part of `dataSourceTFEWorkspaceRead` we have:
https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/data_source_workspace.go#L179

```
	d.Set("resource_count", workspace.ResourceCount)
```

We could add that to the resource schema as `Computed: true`, and include that same `d.Set` in `resourceTFEWorkspaceRead` to make that information available:
https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/resource_tfe_workspace.go#L332

Other attributes read by the data source could be added too, if they would be useful, though `resource_count` is the one I have an immediate use-case for.

If this seems like a sensible approach, I can raise a PR to implement this.

lucymhdavies:
That idea appears to work.

With this change:
```
diff --git a/tfe/resource_tfe_workspace.go b/tfe/resource_tfe_workspace.go
index d94d79d..0cf297e 100644
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -203,6 +203,11 @@ func resourceTFEWorkspace() *schema.Resource {
                                        },
                                },
                        },
+
+                       "resource_count": {
+                               Type:     schema.TypeInt,
+                               Computed: true,
+                       },
                },
        }
 }
@@ -345,6 +350,8 @@ func resourceTFEWorkspaceRead(d *schema.ResourceData, meta interface{}) error {
        d.Set("working_directory", workspace.WorkingDirectory)
        d.Set("organization", workspace.Organization.Name)

+       d.Set("resource_count", workspace.ResourceCount)
+
        var sshKeyID string
[...]

PR #682, opened 2022-11-09
Adding support for the resource_count attribute on tfe_workspace

rhughes1:
## Description

This PR exposes the `resource_count` attribute on the `tfe_workspace` resource. This is to make it easier for users writing a Sentinel policy to prevent workspaces from being deleted prior to deleting all of the resources that belong to the given workspace. This attribute is already exposed as part of the data source call.

Below is a screenshot of the documentation update on the `tfe_workspace` resource:

image

## Testing plan

1. Create a new workspace using the configuration below

```hcl
resource "tfe_organization" "test-organization" {
  name  = "my-org-name"
  email = "admin@company.com"
}

  tag_names    = ["test", "app"]
}

output "workspace_resource_count" {
  value = tfe_workspace.test.resource_count
}
```

This should yield a result like this:

```bash
$ terraform refresh
...
Outputs:

workspace_resource_count = 0
```
2. Inside the new workspace (ie `my-workspace-name`), use the following configuration (or similar) to add resources under the control of the workspace:

```hcl
resource "tfe_team" "test" {
  name         = "my-team-name"
  organization = "my-org-name"
}
```

3. Back on the original workspace, re-run a `terraform refresh` so we can now see that the `resource_count` has changed from `0` to `1`:

```
$ terraform refresh
...
Outputs:

workspace_resource_count = 1
```

## External links
- [Related Issue](https://github.com/hashicorp/terraform-provider-tfe/issues/448)

N/A at this point in time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Unreleased
 
+
 FEATURES:
+* r/tfe_workspace: Add attribute `resource_count` to `tfe_workspace` ([#682](https://github.com/hashicorp/terraform-provider-tfe/pull/682))
 
 ## v0.40.0 (December 6, 2022)
 
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -240,6 +240,10 @@ func resourceTFEWorkspace() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"resource_count": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -406,6 +410,7 @@ func resourceTFEWorkspaceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("trigger_patterns", workspace.TriggerPatterns)
 	d.Set("working_directory", workspace.WorkingDirectory)
 	d.Set("organization", workspace.Organization.Name)
+	d.Set("resource_count", workspace.ResourceCount)
 
 	// Project will be nil for versions of TFE that predate projects
 	if workspace.Project != nil {
--- a/tfe/resource_tfe_workspace_test.go
+++ b/tfe/resource_tfe_workspace_test.go
@@ -62,7 +62,6 @@ func TestAccTFEWorkspace_basic(t *testing.T) {
 						"tfe_workspace.foobar", "working_directory", ""),
 					resource.TestCheckResourceAttr(
 						"tfe_workspace.foobar", "resource_count", "0"),
-<<<<<<< HEAD
 				),
 			},
 		},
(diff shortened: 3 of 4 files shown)


brandonc (commented):
@rhughes1 Hello! And thanks for the contribution. I completely get the sentiment and know how easy it is to accidentally delete a workspace that has resources under management and I don't have a problem with adding resource_count to tfe_workspace.

I also want to point out that a policy may not be required if you use Terraform Cloud because we are actually rolling out a platform-wide safeguard with this problem in mind, known as [safe delete](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#safe-delete-a-workspace) / [force delete](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#force-delete-a-workspace) that is [making its way into the provider](https://github.com/hashicorp/terraform-provider-tfe/pull/675) as a new attribute that you must add to workspace before it can be deleted when resources are still under management.

Hopefully this could also solve your problem without additional policy code. I'll give this a smoke test but it looks pretty good.

brandonc:
@rhughes1 After more investigation, there seems to be a lot fields that are present in data.tfe_workspace that are not present in the tfe_workspace resource. I wonder if a solution like this could solve your problem while we evaluate if/how we should address the larger disparity?

```
  tag_names    = ["test", "app"]
}

data "tfe_workspace" "test" {
  depends_on = [tfe_workspace.test]
}

output "workspace_resource_count" {
  value = data.tfe_workspace.test.resource_count
}
```

rhughes1:
@brandonc So that solution you mentioned would indeed work. However, from a Terraform best practice approach I would think it's a little bit odd having a data source _and_ a resource within the same root module for the same resource (Not to mention using a `depends_on` block). I would have thought typically you should avoid this in Terraform.

I think there still might be value in exposing the `resource_count` for the Terraform workspace.

brandonc:
I would say it's not ideal to have a data source that references a resource and for what it's worth I just realized you could omit the depends_on argument by specifying the name dependency directly:

```
  tag_names    = ["test", "app"]
}

data "tfe_workspace" "test" {
  name         = tfe_workspace.test.name
}

output "workspace_resource_count" {
  value = data.tfe_workspace.test.resource_count
}
```

I think we will unify the tfe_workspace resource/data export-only arguments but might be willing to move forward in adding just resource_count for now. Unfortunately, I can't merge this PR because the [tests are failing](https://github.com/hashicorp/terraform-provider-tfe/actions/runs/3464081233) because the resource_count is not always 0. I will take another look at it this week or next unless you get to it first. Thanks!

annawinkler:
👋 Hi @rhughes1 - could you please update your fork with the latest? We'll also be looking at the test failure to see if we can suggest an update to fix the problem. ✨

rhughes1:
> 👋 Hi @rhughes1 - could you please update your fork with the latest? We'll also be looking at the test failure to see if we can suggest an update to fix the problem. ✨

@annawinkler I updated my fork with the latest changes. Let me know how else I can help move this along :)

annawinkler:
Merged https://github.com/hashicorp/terraform-provider-tfe/pull/712/

annawinkler:
Thanks @rhughes1 ! I updated the code in a separate branch to also confirm the tests were running OK too. Appreciate the contribution! 💖

PR #685, opened 2022-11-09
Nightly tfe tests workflow

sebasslash:
## Description
This PR adds the nightly workflow that builds a TFE instance, runs our set of TFE-related tests by setting `ENABLE_TFE=1`, and then tears down the instance. This is a near 1:1 copy of our existing nightly workflow in go-tfe and our newly added test workflow. 

This PR also removes the `tflocal` configuration and associated workflows, see [go-tfe 589](https://github.com/hashicorp/go-tfe/pull/589)


brandonc (approved):
giphy

.github/workflows/nightly-tfe-test.yml
```diff
@@ -0,0 +1,107 @@
...
+          GITHUB_REGISTRY_MODULE_IDENTIFIER: "hashicorp/terraform-random-module"
+          GITHUB_WORKSPACE_IDENTIFIER: "hashicorp/terraform-random-1"
+          GITHUB_WORKSPACE_BRANCH: "test"
+          GITHUB_TOKEN: "${{ secrets.TESTING_GITHUB_TOKEN }}"
+          MOD_PROVIDER: github.com/hashicorp/terraform-provider-tfe
+          MOD_TFE: github.com/hashicorp/terraform-provider-tfe/tfe
+          MOD_VERSION: github.com/hashicorp/terraform-provider-tfe/version
+        run: |
+          source $HOME/.env
+          gotestsum --format short-verbose -- $MOD_PROVIDER $MOD_TFE $MOD_VERSION -timeout=30m -run "${{ steps.test_split.outputs.run }}"
```
brandonc, line 86: I think we need a slack notification for this workflow (success or failure) to notice nightly failures. WDYT?

.github/scripts/build_instance/main.go
```diff
@@ -65,7 +76,38 @@ func createRun(ctx context.Context, client *tfe.Client) (string, error) {
...
+	// polling the status of the run every 20 seconds or so
+	// should be frequent enough. It's also long enough to ensure
+	// no ticks are dropped.
+	ticker := time.NewTicker(time.Second * 20)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Sprintf("Context canceled: %w", ctx.Err())
```
brandonc, line 95: Is this supposed to be...

```suggestion
			return fmt.Errorf("Context canceled: %w", ctx.Err())
```
?

And if so, how did this compile?

sebasslash: Sloppy on my part :disappointed: , I've relied on the static analyzer and it seems to "break" due to `goenv` (uses `$GOPATH/{version_number}/bin` and I forget to reload my shell so it can use the correct go version).

Uk1288 (approved):
🎉  thanks for putting this together @sebasslash

PR #713, opened 2022-12-07
Disable the no_code registry module test in TFE

brandonc:
This is breaking the nightly. I checked with the registry team and this is the new expected behavior.


sebasslash (approved):
:racehorse: :checkered_flag:

PR #725, opened 2022-12-20
Remove `sidebar_current:` from all docs yaml frontmatter

nfagerlund:
## Description

For reference: [the current frontmatter keys supported by the Terraform Registry's docs formatter.](https://developer.hashicorp.com/terraform/registry/providers/docs#yaml-frontmatter)

This old sidebar_current frontmatter key does nothing, and any effort spent adding it to new files or keeping it up to date is tragic. Let's yank it so newer contributors aren't tempted to mess with it!

Background story: This used to be a hack to make the sidebar navs work properly on the old Middleman-based Terraform.io website. I eventually replaced that hack with a better hack, later we stopped hosting provider docs on terraform.io at all, and then some years later we moved that site off Middleman entirely anyway.

## Testing plan

1. This is a docs-only PR.

### DIFF
--- a/website/docs/d/agent_pool.html.markdown
+++ b/website/docs/d/agent_pool.html.markdown
@@ -1,7 +1,6 @@
 ---
 layout: "tfe"
 page_title: "Terraform Enterprise: tfe_agent_pool"
-sidebar_current: "docs-datasource-tfe-agent-pool"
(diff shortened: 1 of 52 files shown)


annawinkler (approved):
💖

PR #744, opened 2022-12-23
Reword and highlight tfe_outputs note and mention how values are populated

### LINKED ISSUE #583: DOC: Could you add to the documentation a note saying that the outputs should exists before the usage of tfe_output
The current documentation for the usage of `tfe_output` is like this:

```
This data source is used to retrieve the state outputs for a given workspace.
It enables output values in one Terraform configuration to be used in another.
```

Could you add an additional note specifying that the outputs should be created prior to the usage of the `tfe_output`?
Something like:

```markdown
Please notice that the outputs should be generated prior to the usage of the tfe_output.
If the outputs are not created it will error with:

│ This object does not have an attribute named "xxxx".
```

#### Use-cases

**Workspace 1:**

```terraform
resource "google_service_account" "datasync-agent-staging-skim-staging" {
  account_id   = "test-staging"
  display_name = "test-staging"
  project      = "project1"
}
output "bar" {
  value       = google_service_account.test-staging.email
}
```
**Workspace2:**
```terraform
data "tfe_outputs" "foo" {
  organization = "my-org"
  workspace = "workspace1"
}

module "test" {
  service_account_email = data.tfe_outputs.foo.values.bar
}

```

We though that this code should run ok but instead we found this `object does not have attribute` error, and we couldn't figure out the problem. (The error message doesn't point you to the solution in a intuitive way)

I think that a simply note in the doc could help to understand the right sequence and order of execution.

brandonc:
Closes #583


website/docs/d/outputs.html.markdown
```diff
@@ -9,10 +9,8 @@ description: |-
 This data source is used to retrieve the state outputs for a given workspace.
 It enables output values in one Terraform configuration to be used in another.
 
-The `values` attribute has been defined as sensitive because the workspace
-outputs may contain sensitive values and the provider protocol does not support
-marking values as [sensitive](https://www.terraform.io/docs/language/values/outputs.html#sensitive-suppressing-values-in-cli-output) after this information is known. You can access
-the subset of non-sensitive outputs using the `nonsensitive_values` attribute.
+~> **NOTE:** The `values` attribute is preemptively marked sensitive and is only populated after a run completes on the associated workspace. To access non-sensitive values, use the `nonsensitive_values` attribute, which is a subset of outputs
```
Uk1288, line 12: I think we can merge the first and last segments of the sentence since they seem to say the same thing, that is `To access non-sensitive values...` and `...that are known to not be sensitive`. What do you think about something in the lines of, 
```
Use the `nonsensitive_values` attribute to access a subset of the outputs
that are known to be non-sensitive

```

annawinkler, line 12: I think it might be helpful to retain the link to the sensitive documentation:

```suggestion
~> **NOTE:** The `values` attribute is preemptively marked [sensitive](https://www.terraform.io/docs/language/values/outputs.html#sensitive-suppressing-values-in-cli-output) and is only populated _after_ a run completes on the associated workspace. To access non-sensitive values, use the `nonsensitive_values` attribute, which is a subset of outputs.
```

With regards to @Uk1288 's comment, I think it's clearer to have the second sentence as is, starting with `To access non-sensitive values...` because it answers the question that appeared in my mind when reading the doc, how do I access non-sensitive values?

annawinkler (approved):
💯 Thank you for the update. I have one small suggestion and then it looks good!

website/docs/d/outputs.html.markdown
brandonc: I'm going to take @Uk1288's suggestion because it also corrects from passive voice to active voice

PR #715, opened 2022-12-09
Allow tfe_organization_membership import by org name and user email

JarrettSpiker:
## Description

Allow tfe_organization_membership resource to also be imported by org-name/user-email

Presently, the resource can only be imported through its ID, which is not presented to the user in the TFE UI. This makes it very difficult to import.

## Testing plan

1.  Test that import by ID still works
1.  Test that importing by org/email works
2. Test that importing by an invalid format errors
3. Test that importing a non existent org name or email errors

## External links

- [internal JIRA](https://hashicorp.atlassian.net/browse/TF-3336)
- [resource documentation](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/organization_membership)

## Output from acceptance tests

```
❯ TF_ACC=1 go test ./...  -run TestAccTFEOrganizationMembership -tags=integration -v
?       github.com/hashicorp/terraform-provider-tfe     [no test files]
=== RUN   TestAccTFEOrganizationMembershipDataSource_basic
--- PASS: TestAccTFEOrganizationMembershipDataSource_basic (17.87s)
=== RUN   TestAccTFEOrganizationMembershipDataSource_findByName
    data_source_organization_membership_test.go:47: Please set TFE_USER1 to run this test
--- SKIP: TestAccTFEOrganizationMembershipDataSource_findByName (0.36s)
=== RUN   TestAccTFEOrganizationMembershipDataSource_missingParams
--- PASS: TestAccTFEOrganizationMembershipDataSource_missingParams (4.41s)
=== RUN   TestAccTFEOrganizationMembership_basic
--- PASS: TestAccTFEOrganizationMembership_basic (6.66s)
=== RUN   TestAccTFEOrganizationMembershipImport_ByID
--- PASS: TestAccTFEOrganizationMembershipImport_ByID (6.94s)
=== RUN   TestAccTFEOrganizationMembershipImport_ByEmail
--- PASS: TestAccTFEOrganizationMembershipImport_ByEmail (7.05s)
=== RUN   TestAccTFEOrganizationMembershipImport_invalidImportId
--- PASS: TestAccTFEOrganizationMembershipImport_invalidImportId (7.72s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/tfe 51.182s


...
```

PR #745, opened 2022-12-25
Allow team and organization_membership to be imported by user facing values

JarrettSpiker:
## Description

Allow the tfe_team resource to be imported by `<org name>/<team name>`
Allow the tfe_team_organization_member resources to be imported by `<org-name>/<team name>/<user-email>`

Presently, both these resources can only be imported through their ID, which is are presented to the user in the TFE UI. 

When importing a team, we will now check if the relevant peice of the import ID matches the format of a TFE resource ID. If not, or if no team resource exists with that ID, we can search for teams with that name in the given organization

## Testing plan

1.  Check that tfe_team can still be imported by the existing ID
2. Check that tfe_team_org_member can still be imported by the existing ID
3. Check that tfe_team can now be imported by team name
4. Check that tfe_team_org_member can now be imported by org and team name + user email
5. Check that invalid ID formats exist gracefully for both resources
6. Check that importing by a non-existent org name errors gracefully for both resources
7. Check that importing by a non-existent team name errors gracefully for both resources
8. Check that importing team membership by a user email which does not belong to the org fails gracefully

## External links

- [JIRA for team](https://hashicorp.atlassian.net/browse/TF-3356)
- [JIRA for org team memeber](https://hashicorp.atlassian.net/browse/TF-3336)

```
❯ go test ./...  -run TestAccTFETeam_ -tags=integration -v
?       github.com/hashicorp/terraform-provider-tfe     [no test files]
--- PASS: TestAccTFETeam_basic (8.00s)
=== RUN   TestAccTFETeam_full
--- PASS: TestAccTFETeam_full (7.72s)
=== RUN   TestAccTFETeam_full_update
--- PASS: TestAccTFETeam_full_update (17.36s)
=== RUN   TestAccTFETeam_import_byId
--- PASS: TestAccTFETeam_import_byId (8.52s)
=== RUN   TestAccTFETeam_import_byId_doesNotExist
--- PASS: TestAccTFETeam_import_byId_doesNotExist (8.50s)
=== RUN   TestAccTFETeam_import_byName
--- PASS: TestAccTFETeam_import_byName (8.77s)
=== RUN   TestAccTFETeam_import_missingOrg
--- PASS: TestAccTFETeam_import_missingOrg (9.43s)
=== RUN   TestAccTFETeam_import_missingTeam
--- PASS: TestAccTFETeam_import_missingTeam (8.37s)
=== RUN   TestAccTFETeam_import_teamNameWithSpaces
--- PASS: TestAccTFETeam_import_teamNameWithSpaces (8.62s)
=== RUN   TestAccTFETeam_import_teamNameWithSlashes
--- PASS: TestAccTFETeam_import_teamNameWithSlashes (8.50s)
=== RUN   TestAccTFETeam_import_teamNameWhichLooksLikeID
--- PASS: TestAccTFETeam_import_teamNameWhichLooksLikeID (8.53s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/tfe 102.647s

❯ go test ./...  -run TestAccTFETeamOrganizationMember_ -tags=integration -v
?       github.com/hashicorp/terraform-provider-tfe     [no test files]
=== RUN   TestAccTFETeamOrganizationMember_basic
--- PASS: TestAccTFETeamOrganizationMember_basic (10.43s)
=== RUN   TestAccTFETeamOrganizationMember_import_byId
--- PASS: TestAccTFETeamOrganizationMember_import_byId (9.66s)
=== RUN   TestAccTFETeamOrganizationMember_import_byTeamName
--- PASS: TestAccTFETeamOrganizationMember_import_byTeamName (10.27s)
=== RUN   TestAccTFETeamOrganizationMember_import_orgDoesNotExist
--- PASS: TestAccTFETeamOrganizationMember_import_orgDoesNotExist (9.59s)
=== RUN   TestAccTFETeamOrganizationMember_import_teamNameDoesNotExist
--- PASS: TestAccTFETeamOrganizationMember_import_teamNameDoesNotExist (9.73s)
=== RUN   TestAccTFETeamOrganizationMember_import_userEmailDoesNotExist
--- PASS: TestAccTFETeamOrganizationMember_import_userEmailDoesNotExist (9.46s)
=== RUN   TestAccTFETeamOrganizationMember_import_incorrectFormat
--- PASS: TestAccTFETeamOrganizationMember_import_incorrectFormat (10.15s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/tfe 69.605s


...
```


tfe/resource_tfe_team_organization_member.go
```diff
@@ -132,3 +133,40 @@ func unpackTeamOrganizationMemberID(id string) (teamID, organizationMembershipID
...
 	return s[0], s[1], nil
 }
+
+func resourceTFETeamOrganizationMemberImporter(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	tfeClient := meta.(*tfe.Client)
+
+	// Import formats:
+	//  - <TEAM ID>/<ORGANIZATION MEMBERSHIP ID>
+	//  - <ORGANIZATION NAME>/<TEAM NAME>/<USER EMAIL>
+	s := strings.Split(d.Id(), "/")
```
brandonc, line 143: TIL email addresses can contain forward slashes!

Perhaps it would be better to use SplitN and let it fail on the email lookup if there are more than 3 partitions

brandonc (changes requested):
Code review looks great. I haven't smoke tested yet.

tfe/resource_tfe_team_organization_member.go
JarrettSpiker: Shoot, I thought I had this working earlier with escaping the slashes, but looking again now it seems like it does not. Good catch!

> Perhaps it would be better to use SplitN and let it fail on the email lookup if there are more than 3 partitions

I like that suggestion, and am going to use it in [this similar PR](https://github.com/hashicorp/terraform-provider-tfe/pull/715/files#diff-1e092be463f14185a005a40ad7af1b58e0b9b257621d04641dd5c650620112ceR120) I have, but a potential problem is that team names can also have forward slashes (e.g. `team/with/slashes`)...so I could have an import ID like `org-name/team/with/slashes/user/name@email.com` which would partition into the the team being `team` and the email being `with/slashes/user/name@email.com`

I _think_ we would need a way to escape slashes in the emails and team names....maybe something [like this](https://codereview.stackexchange.com/a/259382)?

Or we could change the order of the ID to `<ORGANIZATION NAME>/<USER EMAIL>/<TEAM NAME>`, use SplitN, and accept that it is pretty unlikely to have both an email **and** a team name with slashes in them.

brandonc: That is trouble! I didn't realize that was true about team names, also. There's really not a good way to deal with this, and escaping doesn't feel so great because you'd need a double escape to overcome shell escaping, right?

JarrettSpiker: yah, and it would also be frustrating to have some imports (like this one) where you need to escape slashes, and others (where we can use SplitN) where you shouldnt

I wonder, then, if it is better to arrange the alternate import ID like `<ORGANIZATION NAME>/<USER EMAIL>/<TEAM NAME>` and `SplitN 3`...and just not support the rare case where you have slashes in your email...what do you think?

website/docs/r/team_organization_member.html.markdown
```diff
@@ -47,9 +47,14 @@ The following arguments are supported:
 
 ## Import
 
-A team member can be imported; use `<TEAM ID>/<ORGANIZATION MEMBERSHIP ID>`
+A team member can be imported; use `<TEAM ID>/<ORGANIZATION MEMBERSHIP ID>` or `<ORGANIZATION NAME>/<USER EMAIL>/<TEAM NAME>`
```
brandonc, line 50: I'm actually having trouble importing, but I can see that the correct member/team is being located.

```
$ terraform import tfe_team_organization_member.foo hashicorp/alisdair@hashicorp.com/test-team
tfe_team_organization_member.foo: Importing from ID "hashicorp/alisdair@hashicorp.com/test-team"...
tfe_team_organization_member.foo: Import prepared!
  Prepared tfe_team_organization_member for import
tfe_team_organization_member.foo: Refreshing state... [id=team-ssoLN3CDbW4u1HdG/ou-cWsiUYr3GeMaPJKt]
╷
│ Error: Cannot import non-existent remote object
│
│ While attempting to import an existing object to "tfe_team_organization_member.foo", the provider detected that no object exists with the given id. Only
│ pre-existing objects can be imported; check that the id is correct and that it is associated with the provider's configured region or endpoint, or use "terraform
│ apply" to create a new remote object for this resource.
╵

```

brandonc (commented):
Did I do something wrong when smoke testing? It seems that the correct test/member was found but something else went wrong:

`tfe_team_organization_member.foo: Refreshing state... [id=team-ssoLN3CDbW4u1HdG/ou-cWsiUYr3GeMaPJKt]`

This corresponds with the team/user I was trying to import, but I got the error you see below

website/docs/r/team_organization_member.html.markdown
JarrettSpiker: @brandonc is the user a member of the given team? I believe that is the error I see when trying to import a team membership which doesnt exist (i.e. alisdair is in the org, and the team exists, but isnt a member of the team).

I am battling some dev env demons right now, but will double check as soon as I get those sorted!

JarrettSpiker: I went to improve this error message by having the import read the [organization memberships from the team](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/teams#available-related-resources) it identifies based on org/team-name, but I noticed that to do that properly I will need to [update go-tfe](https://github.com/hashicorp/go-tfe/blob/main/team.go#LL203) to accept `include` options on Team Read.

@brandonc assuming I am correct and the error you are seeing is because the user is not part of the team, and just an unhelpful (pre-existing) error message instead of a new failure...what would you think about leaving that failure in with this PR? I can update go-tfe as necessary to improve it, and circle back once that is done.

brandonc (approved):
I must have been tired when I smoke tested because I forgot to add the member to the team before importing!

PR #752, opened 2023-01-04
d/workspace_ids: Require exact match when no wildcard

### LINKED ISSUE #747: tfe_workspace_ids data source returns unexpected workspaces

TFC for Business

#### Terraform version

```1.1.4```

```
data "tfe_workspace_ids" "workspace_id" {
  names        = ["workspace_name_without_asterisks"]
  organization = var.tf_organization
}
```

#### Expected Behavior

If not using prefix or suffix asterisks * with the **tfe_workspace_ids** data source, matching should be a string comparison over the current string.Contains.  

#### Actual Behavior

This causes unexpected workspaces to get returned with the data source.  If your organization has workspaces with similar names, this data source may select workspaces belonging to different environments.

#### Additional Context

The issue is found here:  https://github.com/hashicorp/terraform-provider-tfe/blob/317355beefd74ab5ef618a6b47af7d4ed224eb66/tfe/data_source_workspace_ids.go#LL59 where strings without asterisks are processed, and then are still being matched with .Contains

### PR DESCRIPTION
**Pull request by brandonc**
## Description

There was a bug in the new d/tfe_workspace_ids wildcard logic that performed a substring match even when no wildcards were present

## External links

Closes #747 

```
$ TESTARGS="-run TestAccTFEWorkspaceIDsDataSource_noMatch" make testacc
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEWorkspaceIDsDataSource_noMatch -timeout 15m
=== RUN   TestAccTFEWorkspaceIDsDataSource_noMatch
--- PASS: TestAccTFEWorkspaceIDsDataSource_noMatch (25.57s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	26.388s
```

### DIFF
--- a/tfe/data_source_workspace_ids.go
+++ b/tfe/data_source_workspace_ids.go
@@ -56,7 +56,7 @@ func includedByName(names map[string]bool, workspaceName string) bool {
 		case len(name) == 0:
 			continue
 		case !strings.HasPrefix(name, "*") && !strings.HasSuffix(name, "*"):
-			if strings.Contains(workspaceName, name) {
+			if name == workspaceName {
 				return true
 			}
 		case strings.HasPrefix(name, "*") && strings.HasSuffix(name, "*"):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 BUG FIXES:
-* d/tfe_workspace_ids: When no wildcard matchers were used in the names argument a substring match was being performed anyway
+* d/tfe_workspace_ids: When no wildcard matchers were used in the names argument a substring match was being performed anyway @brandonc ([#752](https://github.com/hashicorp/terraform-provider-tfe/pull/752))
 
 FEATURES:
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: tfe/data_source_workspace_ids.go [hunk 1]
```diff
@@ -56,7 +56,7 @@ func includedByName(names map[string]bool, workspaceName string) bool {
 		case len(name) == 0:
 			continue
 		case !strings.HasPrefix(name, "*") && !strings.HasSuffix(name, "*"):
-			if strings.Contains(workspaceName, name) {
+			if name == workspaceName {
```
laurenolivia, line 59: 👍 Nice catch. Wondering how the behavior of `strings.Contains(workspaceName, name)` creates a bug, since they both perform a string comparison?

nfagerlund: It's because `Contains` doesn't match the whole string -- `strings.Contains("partial", "art")` will return true, but `"partial" == "art"` is false.

nfagerlund (approved):
This looks good. ✅ However! You've got the wrong jira issue number in your branch name, so something unrelated ended up in the review column on our board. 😆

laurenolivia:
👍 thanks @nfagerlund.

laurenolivia (approved):
✅

Issue #761, opened 2023-01-10
problem destroying created users if they are pending


ninjadude333:
#### Terraform Cloud/Enterprise version
```
Terraform Cloud Ent 1.3.7
```

#### Terraform version
```
Terraform v1.3.5
on windows_amd64
```

#### Terraform Configuration Files

```
resource "tfe_team" "Maturity_Networking" {
  name         = "Maturity-Networking"
  organization = "my-org"
}

resource "tfe_team_access" "Networking" {
  access       = "write"
  team_id      = tfe_team.Maturity_Networking.id
  workspace_id = tfe_workspace.xyz.id
}

resource "tfe_organization_membership" "Network-member" {
  organization = "my-org"
  email        = "moshe@terasky.com"
}

resource "tfe_team_organization_members" "Network-membership" {
  team_id = tfe_team.Maturity_Networking.id
  organization_membership_ids = [
    tfe_organization_membership.Network-member.id
  ]
}
```

#### Debug Output

```
Error: Error removing organization membership [ou-cBPYrENVPk5kdCrg ou-qLK7jAGpXqwat9J2] to team team-7YqSiLKNGc8iBZQy: bad request
│
│ ou-cBPYrENVPk5kdCrg is not a member
│ bad request
```

#### Expected Behavior
the destory run should have completed successfully and delete the pending users. (or at least finish the run without breaking)

#### Actual Behavior
destroy run is failing with the above error

#### Additional Context
if the pending user was already pending, then the apply run will also fail with the error:
```
│ Error: Error creating membership masha@terasky.com for organization TeraSky: invalid attribute
│
│ User is already an organization member
│
│   with tfe_organization_membership.App2-members["masha@terasky.com"],
│   on teams.tf line 72, in resource "tfe_organization_membership" "App2-members":
│   72: resource "tfe_organization_membership" "App2-members" {
```

Uk1288:
Hello @ninjadude333, I was able to successfully run terraform apply and destroy on the terraform configuration above. The debug output above tries to remove 2 organization memberships but the config depicts adding only 1 membership. Can you confirm that, 
(1) this is the complete config producing this error 
(2) the two org memberships exist before running terraform destroy

[...]

PR #767, opened 2023-01-13
update website doc to depict the use of sets in members resource

Uk1288:
## Description

Related Issue: [761](https://github.com/hashicorp/terraform-provider-tfe/issues/761)

Terraform creates an instance for each member of a given set, when `for_each` is used to configure a resource block. 

If `for_each` is used with resources that do bulk management like the `tfe_team_organization_members` and `tfe_team_members`, it results in the creation of multiple instances managing the same set of resources. See the example config below:

```
locals {
  all_users = toset([
    "user1@hashicorp.com",
    "user2@hashicorp.com",
  ])
}

resource "tfe_team" "test" {
  name         = "my-team-name"
  organization = "my-org-name"
}

resource "tfe_organization_membership" "all_membership" {
  organization = "my-org-name"
  for_each     = local.all_users
  email        = each.key
}

resource "tfe_team_organization_members" "all-members" {
  team_id  = tfe_team.test.id
  for_each = local.all_users
  organization_membership_ids = [
    tfe_organization_membership.all_membership[each.key].id
  ]
}

```

The above config results in multiple instances of `tfe_team_organization_members` managing a set of memberships. Running an Apply will succeed. During the destroy phase, each instance of the `tfe_team_organization_members` resource attempts to delete the same organization members. The first delete request succeeds but the subsequent requests fail. If there are `n` number of `local.all_users`, there will be `n` DELETE API request made with the same membership id set (the current delete implementation attempts to delete all existing memberships).

What is needed here, is for Terraform to create just `1 instance` of `tfe_team_organization_members` which will manage all the organization members in the set. The `for` expression will accomplish this, see the updated tfe_team_organization_members config below:

```
resource "tfe_team_organization_members" "all-members" {
  team_id = tfe_team.test.id
  organization_membership_ids = [for member in local.all_users : 
     tfe_organization_membership.all_membership[member].id]
}

```

After investigating this issue, the conclusion is that it is beneficial to add this use case example to the doc hence this PR.

## Testing plan

1.  _Apply and compare the listed config above_

## External links

- [API documentation - for-each meta argument](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each)
- [API documentation - for expression](https://developer.hashicorp.com/terraform/language/expressions/for)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ BUG FIXES:
 
 ENHANCEMENTS:
 * Update API doc links from terraform.io to developer.hashicorp domain by @uk1288 [#764](https://github.com/hashicorp/terraform-provider-tfe/pull/764)
+* Update website docs to depict the use of set with `tfe_team_organization_members` and `tfe_team_members` by @uk1288 [#767](https://github.com/hashicorp/terraform-provider-tfe/pull/767)
 
 FEATURES:
 * r/tfe_team: Teams can now be imported using `<ORGANIZATION NAME>/<TEAM NAME>` ([#745](https://github.com/hashicorp/terraform-provider-tfe/pull/745))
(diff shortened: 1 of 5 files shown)


laurenolivia (approved):
👍 Really clear description! Smoke tested this and it works as expected.

PR #773, opened 2023-01-20
feat: Data source for workspace tags in an organization

rhughes1:
## Description

This is a new data source is meant to retrieve all the workspace tags for a given organization. With this new data source, we can get the ID, name, and instance count for the organization.

## Testing plan

1. Create a Terraform Cloud/Enterprise organization
2. Create a new workspace with tags (For example, `foo` and `bar`)
3. Create another workspace with similar tags and different tags (For example, `foo` and `test`)
4. Write the necessary Terraform configuration to pull down the tags for a given organization.

```
data "tfe_organization_workspace_tags" "example" {
  organization = "my-org-name"
}
```
5. In the `tags` attribute, we should see something like

```
[
  {
    id = "tag-12345q"
    name = "foo"
    instance_count = "2"
  },
  {
    id = "tag-qwert12"
    name = "bar"
    instance_count = "1"
  },
  {
    id = "tag-qwe333rt12"
    name = "test"
    instance_count = "1"
  },
]
```

Doc Preview:

image

## External links
- [go-tfe documentation](https://pkg.go.dev/github.com/hashicorp/go-tfe#OrganizationTags)
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/organization-tags#list-tags)


uturunku1:
@rhughes1 Hi Ryan! Thank you for your valuable contribution. I have try out your changes, and so far they look good. One of my colleagues was suggesting that you rename one of the tag attributes. What do you think of changing `instance_count` to `workspace_count`? Maybe the word "instance" could be a little confusing in this context?

rhughes1:
I kept it to `instance_count` just to keep it aligned with the API. I can easily make the change if that's preferred?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 FEATURES:
+* **New Data Source**: d/tfe_organization_workspace_tags is a new datasource to allow reading workspace tags within an organization, by @rhughes1 ([#773](https://github.com/hashicorp/terraform-provider-tfe/pull/773))
```
brandonc, line 4: Thanks for contributing this! I only see one adjustment I'd like you to make to be more in keeping with other resource conventions. Can you rename this to `tfe_organization_tags`?

Reason being, it isn't scoped to a particular workspace and better reflects usage of the actual endpoint, [Organization Tags](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/organization-tags)

Eventually, we'll want to add arguments that reflect the other capabilities of the endpoint like tag name filtering, but that doesn't diminish the usefulness of the data source. Thanks again!

brandonc (changes requested):
I smoke tested this and it worked as expected. Let me know what you think! Thanks again

CHANGELOG.md
rhughes1: Of course! The reason I called it `organization_workspace_tags` was _in case_ the new projects feature would eventually get the ability to tag we wouldn't have to rename this data source. I called it `organization_workspace_tags` so it was very explicit to end users.

tfe/provider.go
```diff
@@ -106,25 +106,25 @@ func Provider() *schema.Provider {
...
-			"tfe_policy_set":              dataSourceTFEPolicySet(),
-			"tfe_organization_members":    dataSourceTFEOrganizationMembers(),
+			"tfe_organizations":               dataSourceTFEOrganizations(),
+			"tfe_organization":                dataSourceTFEOrganization(),
+			"tfe_agent_pool":                  dataSourceTFEAgentPool(),
+			"tfe_ip_ranges":                   dataSourceTFEIPRanges(),
+			"tfe_oauth_client":                dataSourceTFEOAuthClient(),
+			"tfe_organization_membership":     dataSourceTFEOrganizationMembership(),
+			"tfe_organization_run_task":       dataSourceTFEOrganizationRunTask(),
+			"tfe_organization_workspace_tags": dataSourceTFEOrganizationTags(),
```
brandonc, line 116:
```suggestion
			"tfe_organization_tags": dataSourceTFEOrganizationTags(),
```

One more and I think we are good

tfe/data_source_organization_tags_test.go
```diff
@@ -0,0 +1,60 @@
...
+		},
+	})
+}
+
+func testAccTFEOrganizationTagsDataSourceConfig(rInt int) string {
+	return fmt.Sprintf(`
+	resource "tfe_organization" "foobar" {
+		name  = "tst-terraform-%d"
+		email = "admin@company.com"
+	  }
```
nfagerlund, line 37: The indentation on this terraform config string is kinda messed up, probably because your editor was trying to "help" by adding Go-appropriate tab literals. Could I get you to de-indent the top-level of this config so it's flush against the left, and then use two spaces for indenting block bodies?

This is lower priority than my other comments.

tfe/data_source_organization_tags.go
```diff
@@ -0,0 +1,81 @@
...
+							Computed: true,
+						},
+
+						"id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"workspace_count": {
+							Type:     schema.TypeString,
```
nfagerlund, line 36: I think this should be a `schema.TypeInt`, to simplify using it as a number from within Terraform configs.

tfe/data_source_organization_tags.go
```diff
@@ -0,0 +1,81 @@
...
+
+	options := tfe.OrganizationTagsListOptions{}
+	for {
+		organizationTagsList, err := tfeClient.Client.OrganizationTags.List(ctx, organizationName, &options)
+		if err != nil {
+			return fmt.Errorf("Error retrieving organization tags: %w", err)
+		}
+
+		for _, orgTag := range organizationTagsList.Items {
+			tag := map[string]string{"id": orgTag.ID, "name": orgTag.Name, "workspace_count": strconv.Itoa(orgTag.InstanceCount)}
```
nfagerlund, line 64: I'm guessing this was why you opted to make the instance count a TypeString instead of a TypeInt, right? But it's fine to build the tag as a `map[string]interface{}` and use different value types for different keys — the schema helper package is designed to help recover from that kind of type erasure. 

For an example of an existing resource that does exactly this, take a look at the `resourceTFETeamAccessRead()` function: `workspace_locking` and `run_tasks` are `TypeBool` in the schema, and the rest are `TypeString`.

https://github.com/hashicorp/terraform-provider-tfe/blob/f805460f41b8a77266b66eb03fe34d6be61c595d/tfe/resource_tfe_team_access.go#L238-L246

(As a bonus, note that `dataSourceTFETeamAccessRead()` re-uses this same function to build a list of object-typed values, which is about what you're doing.)

brandonc (changes requested):
Hey, sorry for not getting feedback the last couple of days. We had some issues with our test infrastructure due to some internal tooling upgrades.

When I run your new test, I get the following error. Can you replicate this?

```
$ TESTARGS="-run TestAccTFEOrganizationTagsDataSource_basic" make testacc
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEOrganizationTagsDataSource_basic -timeout 15m
=== RUN   TestAccTFEOrganizationTagsDataSource_basic
    data_source_organization_tags_test.go:16: Step 1/1 error: Check failed: 1 error occurred:
        	* Check 2/2 error: data.tfe_organization_tags.foobar: list or set attribute 'tags' must be checked by element count key (tags.#) or element value keys (e.g. tags.0). Set element value checks should use TestCheckTypeSet functions instead.

--- FAIL: TestAccTFEOrganizationTagsDataSource_basic (26.97s)
FAIL
FAIL	github.com/hashicorp/terraform-provider-tfe/tfe	27.304s
FAIL
make: *** [testacc] Error 1
```

rhughes1:
@brandonc I modified the tests so that it's specifically looking at the attributes being set within the `tags` attribute for more clarity. Do those changes make sense?

brandonc (approved):
Nice! Thanks!

PR #774, opened 2023-01-24
multiple workspaces management example

Uk1288:
## Description

This PR show an example of using the tfe provider to manage multiple workspaces with associated workspace variables. 

 _[Origin PR](https://github.com/hashicorp/terraform-provider-tfe/pull/129)_

## Testing plan

1.  Run the workspaces example config in the PR

## External links

- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/129)


laurenolivia (approved):
awesome ✨

examples/workspaces_variables/managed/main.tf
```diff
@@ -0,0 +1,77 @@
...
+# Variables
+#
+
+variable "sens_tf_var" {
+  default = ""
+}
+
+variable "sens_env_var" {
+  default = ""
+}
```
sebasslash, line 14: Are these supposed to be marked sensitive? Going of the `sens_` prefix.

examples/workspaces_variables/managed/main.tf
```diff
@@ -0,0 +1,77 @@
...
+
+#
+# Output all of the variables for demonstration purposes
+#
+
+output "sens_tf_var" {
+  value = var.sens_tf_var
+}
```
sebasslash, line 46: Same here

examples/workspaces_variables/managed/main.tf
```diff
@@ -0,0 +1,77 @@
...
+  default = "a default string"
+}
+
+# Value is initially supplied via some.auto.tfvars
+variable "a_number" {
+  type = number
+}
+
+variable "a_list" {
+  type = list
```
sebasslash, line 27: Should be `list(<type>)`, same goes for `variable "a_map"`

examples/workspaces_variables/manager/main.tf
```diff
@@ -0,0 +1,156 @@
...
+}
+
+variable "vars_mapped_by_workspace_name" {
+    description = <<-EOT
+    This is the map of workspaces and variables. A workspace is created for each
+    top level key and then variables are set on the workspace.
+    EOT
+}
+
+variable "additional_vars" {
```
sebasslash, line 29: These variables should specify their type. Given this is an example for those starting to manage TFC workspaces we should eliminate as many assumptions as we can.

sebasslash (changes requested):
[Non-blocking feedback] There are tons of different ways to write a configuration -- personally I find this example to be a tad bit convoluted surrounding how variables are populated for the managed workspaces, especially if the assumption here is the customer is starting out on managing cloud workspaces. I'm not particularly fond of having one giant map being the source of truth to provision variables for each workspace -- but again, personal bias. Also it's possible to get away with using one github repository for both manager/managed configurations and simply changing the working directory. 

[Blocking feedback] Bias aside, there are some really small blocking items down below :arrow_down: 

That said, this is great work! :rocket:

sebasslash:
Mentioning @brenthc for visibility

sebasslash (approved):

PR #776, opened 2023-01-27
Add copywrite headers and GH Action

barrettclark:
### :hammer_and_wrench: Description

* HashiCorp public repos will begin being scanned for source code copyright headers starting Feb.1, 2023. We want to make sure this repo is in adherence to the automated checks.
* Add Github action which will run a check on the repo during pull requests. This is simply a validation and will not automatically correct any source.
* Bring repo up-to-date using the `copywrite` CLI.

### :link: External Links

* [GH Action](https://github.com/marketplace/actions/setup-copywrite)
* `copywrite` [CLI](https://github.com/hashicorp/copywrite#getting-started)


sebasslash (approved):

barrettclark:
I consolidated 4 workflows that all run on `push` into a single CI workflow, like we have in other repos.

.github/workflows/mkdocs.yml
```diff
@@ -1,20 +0,0 @@
-name: publish-docs
-on:
-  push:
-    branches:
-      - main
```
brandonc, line 5: ☝️

brandonc (changes requested):
I like your enthusiasm but mkdocs should not be a CI action. This should only happen on main branch. Can you restore this file?

.github/workflows/mkdocs.yml
barrettclark: Yup. I redid and force pushed the second commit where I did the consolidation.

.copywrite.hcl
```diff
@@ -0,0 +1,14 @@
+schema_version = 1
+
+project {
+  license        = "MPL-2.0"
+  copyright_year = 2018
```
brandonc, line 5: Not a big deal, but do you think we should bump this up? I wonder where 2018 came from 🤔

brandonc (approved):

.copywrite.hcl
barrettclark: The copywrite script suggested it. I assumed it was when the repo was created.

Issue #338, opened 2021-08-09
resource/tfe_workspace: Provide a way to work around manually configured GitHub App VCS configs


gdavison:
Currently, the resource `tfe_workspace` does not allow VCS authentication with the GitHub App method, as described at https://www.terraform.io/docs/cloud/vcs/github-app.html#feature-limitations.

When using the GitHub App method, it would be useful to allow the workspace to be created using Terraform and then for the VCS configuration to be applied manually. Unfortunately, this causes perpetual diffs to be reported by Terraform. If `ignore_changes` is used, any changes to other parameters on the workspace causes an `internal server error` response.

#### Use-cases

Configuring a workspace with GitHub App instead of OAuth

#### Attempted Solutions

1. Create `tfe_workspace` without `vcs_repo`, then manually configure VCS. This works for creation, but leads to a perpetual difference reported by Terraform. Using `ignore_changes` removes the difference, but any updates cause an `internal server error` response.
1. Manually create Workspace and import. This still requires `ignore_changes` and updates cause an `internal server error` response.

#### Proposal

Several possibilities:

1. In the AWS provider, we have a `DiffSuppressFunc` which we use on some optional blocks to allow the Terraform configuration to not specify the block. Setting it to `Computed` in addition to `Optional` might work as well.

```go
func suppressMissingOptionalConfigurationBlock(k, old, new string, d *schema.ResourceData) bool {
	return old == "1" && new == "0"
}
```

1. In Terraform state, when GitHub App is used, the `oauth_token_id` attribute is empty. Diff detection could ignore changes when the attribute is empty and not send the `vcs_repo` in the update operation

pksunkara:
Are there any plans to implement this? This is the most annoying thing about TFE provider IMO, unable to create workspaces.

brandonc:
I agree this could be improved. This could qualify as more of a bug.

nfagerlund:
Aside from the provider behavior though, "internal server error" probably should not be appearing when you do this; exposing a 500 to users is a bug, right?

ezzatron:
I believe this should be resolved if #808 gets merged 🤞

lucymhdavies:
https://github.com/hashicorp/terraform-provider-tfe/pull/808 was indeed merged. Is this still an issue?

pksunkara:
Nope, this can be closed.

PR #777, opened 2023-01-27
Mark `vcs_repo` as computed, allowing for manual configuration. 

sebasslash:
## Description

This change stems from #338 , where users were unable to manually configure a VCS connection on a TFC workspace that was provisioned with Terraform. By marking the attribute as computed, it can read the value from the API and not result in perpetual planned changes if the attribute is not defined in the configuration. 

### Acceptance Test
```
export GITHUB_TOKEN=<your github token>
export GITHUB_WORKSPACE_IDENTIFIER=<your-username>/<repo-name>
TESTARGS="-run TestAccTFEWorkspace_vcsRepoDrift" make testacc
```
### Smoke Test
1. Create a workspace config and apply
```hcl
  name = "foo"
  organization = "hashicorp"
}
```
2. Visit the UI and add a VCS connection to the workspace you just provisioned
3. Reapply the above config and there should be no planned changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ BUG FIXES:
 ENHANCEMENTS:
 * Update API doc links from terraform.io to developer.hashicorp domain by @uk1288 [#764](https://github.com/hashicorp/terraform-provider-tfe/pull/764)
 * Update website docs to depict the use of set with `tfe_team_organization_members` and `tfe_team_members` by @uk1288 [#767](https://github.com/hashicorp/terraform-provider-tfe/pull/767)
-* d/tfe_workspace: Add `execution_mode` field to workspace datasoure @Uk1288 ([#772](https://github.com/hashicorp/terraform-provider-tfe/pull/772))
+* d/tfe_workspace: Add `execution_mode` field to workspace datasoure by @uk1288 ([#772](https://github.com/hashicorp/terraform-provider-tfe/pull/772))
+* r/tfe_workspace: `vcs_repo` is now marked as computed, allowing for manual VCS configuration by @sebasslash ([#777](https://github.com/hashicorp/terraform-provider-tfe/pull/777))
 
 FEATURES:
 * **New Provider Config**: `organization` defines a default organization for all resources, making all resource-specific organization arguments optional.
--- a/tfe/resource_tfe_workspace.go
+++ b/tfe/resource_tfe_workspace.go
@@ -204,6 +204,7 @@ func resourceTFEWorkspace() *schema.Resource {
 			"vcs_repo": {
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true,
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ BUG FIXES:
 ENHANCEMENTS:
 * Update API doc links from terraform.io to developer.hashicorp domain by @uk1288 [#764](https://github.com/hashicorp/terraform-provider-tfe/pull/764)
 * Update website docs to depict the use of set with `tfe_team_organization_members` and `tfe_team_members` by @uk1288 [#767](https://github.com/hashicorp/terraform-provider-tfe/pull/767)
-* d/tfe_workspace: Add `execution_mode` field to workspace datasoure @Uk1288 ([#772](https://github.com/hashicorp/terraform-provider-tfe/pull/772))
+* d/tfe_workspace: Add `execution_mode` field to workspace datasoure by @uk1288 ([#772](https://github.com/hashicorp/terraform-provider-tfe/pull/772))
+* r/tfe_workspace: `vcs_repo` is now marked as computed, allowing for manual VCS configuration by @sebasslash ([#777](https://github.com/hashicorp/terraform-provider-tfe/pull/777))
```
nfagerlund, line 10: Mmm, this description is a bit too "inside baseball" for normal users, though it'd make sense to veteran provider developers. How bout,

```suggestion
* r/tfe_workspace: If `vcs_repo` isn't specified in the config, allow manually connecting a VCS repo without reporting it as a diff by @sebasslash ([#777](https://github.com/hashicorp/terraform-provider-tfe/pull/777))
```

nfagerlund (commented):
All right, I ran the code with an example config and it works as described when I do out-of-band manual changes. 

I've got one big reservation, and I'd like to hear your thonks about it before approving: 

This change makes it impossible for Terraform to remove an existing VCS configuration. Since `vcs_repo` is a nested block rather than an object-typed attribute, the only way to represent absence is by omitting it, and omitting it now means "ignore." 

I liked the cleanliness of the original reporter's idea of using `ignore_changes` to accomplish this, and it's a shame it blew up. Did we ever learn why that didn't work as expected? Is there anything we can do to make _that_ approach work, preserving the option of enforcing the absence of a VCS connection?

sebasslash:
This is an excellent find and definitely gives me pause. I'm unsure as to why `ignore_changes` does not work as expected but I'm guessing that it might have something to do with the "replace on update" logic around VCS connections. Will spend sometime investigating more.

sebasslash:
TIL, SDKv2 attributes marked as computed reuse previous state and ignore any removals from your configuration. The plugin framework solves this exact issue which would eliminate the need for us to hack together a way to determine if we should reuse previous state or changes from the configuration. 

I'm going to close this and move this bugfix to our v1.0 roadmap as we will be migrating to the plugin framework.

PR #781, opened 2023-01-31
Changelog: Restore the "Unreleased" header

nfagerlund:
## Description

This just adds the "Unreleased" header back to the changelog, as a landing pad for upcoming PRs.


annawinkler (approved):
I like this so then it's clear which section to update now.

brandonc:
Before I saw that this was a PR I committed the same change directly to main.

nfagerlund:
I'll do that next time

PR #785, opened 2023-02-03
Disable copyright checking temporarily

annawinkler:
## Description

The GH action to check copywrite is failing right now with an API rate limit exceeded.

Since that's blocking our builds, I'm disabling the check until the issue is resolved.

## External links

- [Issue](https://github.com/hashicorp/copywrite/issues/28)


nfagerlund (approved):

brandonc (approved):

brandonc (changes requested):
You'll need to comment out the command itself as well

PR #786, opened 2023-02-03
Add tfe_workspace_run resource

Uk1288:
### Add tfe_workspace_run resource as per [discussed issue](https://github.com/hashicorp/terraform-provider-tfe/issues/742)

Resource format:
```
resource "tfe_workspace_run" "root_A" {
  workspace_id    = "ws_root_A_ID"

  apply {
    wait_for_run = true
    manual_confirm = true
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }

  destroy { 
    wait_for_run = true
    manual_confirm = true
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }
}
```

## Testing plan

1.  _Sample test config (Be sure that the org and workspaces exist)_
```
resource "tfe_workspace_run" "root_A" {
  workspace_id    = "ws_root_A_ID"

  apply {
    manual_confirm = true
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }

  destroy { 
    manual_confirm = true
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }
}

resource "tfe_workspace_run" "B_depends_on_A" {
  workspace_id    = "ws_B_depends_on_A_ID"
  depends_on   = [tfe_workspace_run.root_A]

  apply {
    manual_confirm = true
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }

  destroy {
    manual_confirm = true
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }
}

resource "tfe_workspace_run" "D_depends_on_B" {
  workspace_id    = "ws_D_depends_on_B_ID"
  depends_on   = [tfe_workspace_run.B_depends_on_A]

  apply {
    manual_confirm = false
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }

  destroy {
    manual_confirm = false
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }
}

resource "tfe_workspace_run" "C_depends_on_A" {
  workspace_id    = "ws_C_depends_on_A_ID"
  depends_on   = [tfe_workspace_run.root_A]

  apply {
    manual_confirm = false
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }

  destroy {
    manual_confirm = false
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }
}

resource "tfe_workspace_run" "E_depends_on_C_and_D" {
  workspace_id    = "ws_E_depends_on_C_and_D_ID"
  depends_on   = [tfe_workspace_run.D_depends_on_B, tfe_workspace_run.C_depends_on_A]

  apply {
    manual_confirm = false
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }

  destroy {
    manual_confirm = false
    retry = true
    retry_attempts = 2
    retry_backoff_min = 2
    retry_backoff_max = 15
  }
}
```
2.  _the `apply` runs should apply in order from root to leave nodes_
3.  _the `destroy` runs should apply from leave nodes to root node_
4. A second test case would be testing with `wait_for_run = false`. In this case when a run is created, it proceeds to a stage where no other human action is needed then it is considered complete without awaiting for run final state. If `manual_confirm` is `true`, then run will run wait up until approval is given.

## External links

- [Issue](https://github.com/hashicorp/terraform-provider-tfe/issues/742)

```
$ TESTARGS="-run TestAccTFEWorkspaceRun_" make testacc

...
```

Screen Shot 2023-05-02 at 6 52 37 AM

Screen Shot 2023-04-26 at 3 01 42 PM

Screen Shot 2023-04-26 at 3 02 41 PM

### DIFF
--- a/tfe/resource_tfe_workspace_run.go
+++ b/tfe/resource_tfe_workspace_run.go
@@ -67,10 +67,11 @@ func resourceTFEWorkspaceRun() *schema.Resource {
 				ForceNew: true,
 			},
 			"apply": {
-				Type:     schema.TypeList,
-				Elem:     resourceTFEWorkspaceRunSchema(),
-				Required: true,
-				MaxItems: 1,
+				Type:         schema.TypeList,
+				Elem:         resourceTFEWorkspaceRunSchema(),
+				Optional:     true,
+				AtLeastOneOf: []string{"apply", "destroy"},
+				MaxItems:     1,
 			},
 			"destroy": {
 				Type:     schema.TypeList,
--- a/tfe/resource_tfe_workspace_run.go
+++ b/tfe/resource_tfe_workspace_run.go
@@ -63,7 +63,7 @@ func resourceTFEWorkspaceRun() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"organization": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 			},
 			"workspace": {
--- a/tfe/resource_tfe_workspace_run.go
+++ b/tfe/resource_tfe_workspace_run.go
@@ -100,6 +100,11 @@ func resourceTFEWorkspaceRun() *schema.Resource {
 							Optional: true,
 							Default:  30,
 						},
+						"wait_for_run": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+						},
 					},
 				},
 				Optional:     true,
@@ -134,6 +139,11 @@ func resourceTFEWorkspaceRun() *schema.Resource {
 							Type:     schema.TypeInt,
 							Optional: true,
 							Default:  30,
+						},
+						"wait_for_run": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
 						},
 					},
 				},
--- a/tfe/resource_tfe_workspace_run.go
+++ b/tfe/resource_tfe_workspace_run.go
@@ -103,6 +103,16 @@ func resourceTFEWorkspaceRunUpdate(d *schema.ResourceData, meta interface{}) err
 }
 
 func resourceTFEWorkspaceRunRead(d *schema.ResourceData, meta interface{}) error {
+	// First check whether this is a destroy-only run
+	_, ok := d.GetOk("apply")
+	if !ok {
+		// If there's no apply, then there won't be anything to "read" until we
+		// do a destroy run. Return now and leave the ID alone, so that we keep
+		// the resource in the state and get a destroy run when the time comes.
+		log.Printf("[DEBUG] Run %s (random ID) has no apply; nothing to read for refresh", d.Id())
+		return nil
+	}
+
 	config := meta.(ConfiguredClient)
 
 	log.Printf("[DEBUG] Read run for: %s", d.Id())
@@ -110,6 +120,10 @@ func resourceTFEWorkspaceRunRead(d *schema.ResourceData, meta interface{}) error
 	_, err := config.Client.Runs.Read(ctx, runID)
 	if err != nil {
 		if errors.Is(err, tfe.ErrResourceNotFound) {
+			// It would be very strange for this to happen, since runs can't
+			// normally be deleted independently. But this *probably* means we
+			// never performed the initial apply, so we'll remove the missing
+			// run from the state to force an apply to happen.
 			log.Printf("[DEBUG] Run %s does not exist", d.Id())
 			d.SetId("")
 			return nil
--- a/tfe/resource_tfe_workspace_run.go
+++ b/tfe/resource_tfe_workspace_run.go
@@ -14,16 +14,17 @@ var (
 	backoffMax = 3000.0
 )
 
-var planPendingStatus = map[tfe.RunStatus]bool{
+var planPendingStatuses = map[tfe.RunStatus]bool{
 	tfe.RunPending:        true,
 	tfe.RunPlanQueued:     true,
 	tfe.RunPlanning:       true,
 	tfe.RunCostEstimating: true,
 	tfe.RunPolicyChecking: true,
 	tfe.RunQueuing:        true,
+	tfe.RunFetching:       true,
 }
 
-var planTerminalStatus = map[tfe.RunStatus]bool{
+var planTerminalStatuses = map[tfe.RunStatus]bool{
 	tfe.RunPlanned:            true,
 	tfe.RunPlannedAndFinished: true,
 	tfe.RunErrored:            true,
@@ -33,19 +34,20 @@ var planTerminalStatus = map[tfe.RunStatus]bool{
 	tfe.RunPolicyOverride:     true,
 }
 
-var applyPendingStatus = map[tfe.RunStatus]bool{
+var applyPendingStatuses = map[tfe.RunStatus]bool{
 	tfe.RunConfirmed:   true,
 	tfe.RunApplyQueued: true,
 	tfe.RunApplying:    true,
 	tfe.RunQueuing:     true,
+	tfe.RunFetching:    true,
 }
 
-var applyDoneStatus = map[tfe.RunStatus]bool{
+var applyDoneStatuses = map[tfe.RunStatus]bool{
 	tfe.RunApplied: true,
 	tfe.RunErrored: true,
 }
 
-var confirmationDoneStatus = map[tfe.RunStatus]bool{
+var confirmationDoneStatuses = map[tfe.RunStatus]bool{
 	tfe.RunConfirmed:   true,
 	tfe.RunApplyQueued: true,
 	tfe.RunApplying:    true,
@@ -74,34 +76,29 @@ func resourceTFEWorkspaceRun() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"manual_confirm": {
-							Type:        schema.TypeBool,
-							Optional:    true,
-							Default:     false,
-							Description: "",
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
 						},
 						"retry": {
-							Type:        schema.TypeBool,
-							Optional:    true,
-							Default:     true,
-							Description: "",
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
 						},
 						"retry_attempts": {
-							Type:        schema.TypeInt,
-							Optional:    true,
-							Default:     3,
-							Description: "",
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  3,
 						},
 						"retry_backoff_min": {
-							Type:        schema.TypeInt,
-							Optional:    true,
-							Default:     1,
-							Description: "",
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  1,
 						},
 						"retry_backoff_max": {
-							Type:        schema.TypeInt,
-							Optional:    true,
-							Default:     30,
-							Description: "",
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  30,
 						},
 					},
 				},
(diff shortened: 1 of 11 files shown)


website/docs/r/workspace_run.html.markdown
```diff
@@ -0,0 +1,192 @@
...
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_workspace_run"
+description: |-
+  Manages run create and destroy lifecycles in a workspace.
+---
+
+# tfe_workspace_run
+
+Provides a resource to manage create and destroy lifecycles in a workspace.
+The `tfe_workspace_run` expects to own exactly one apply during a creation and one destroy during a destruction. This implies that the `tfe_workspace_run` resource will queue an apply even though previous successful applies exist in the workspace prior to adding this resource.
```
lucymhdavies, line 11: The wording here is a bit unclear.

Maybe something like:
> expects to own exactly one apply during a creation and/or one destroy during a destruction.

I.e. with both `apply` and `destroy` blocks being optional (but you have to specify at least 1), clarify the behaviour if you only specify one of them and not the other.

lucymhdavies:
Forgive me if this isn't ready to be reviewed yet.
While of course I'm not on the team responsible for the provider... this is looking like a good implementation of the solution discussed in the linked issue, and I look forward to this being included.
As an active user of the multispace provider, this version with a single resource, and optional apply/destroy blocks was my favourite of the ones suggested :)

And I like the way you're handling retries with a recursive call to the same function; that's pretty neat. + exponential backoffs in retries are nice; I hadn't even considered that.

I think the only thing missing based on the discussions is the option to kick off a run, and then consider that good enough. i.e. do not wait for the run to finish before considering the resource created.
Shouldn't be the default behaviour, but it's a very handy optional behaviour I use all the time from the multispace provider (or rather... from my fork of it).



But yeah, short version: looking good, and really looking forward to this!

Uk1288:
> Forgive me if this isn't ready to be reviewed yet. While of course I'm not on the team responsible for the provider... this is looking like a good implementation of the solution discussed in the linked issue, and I look forward to this being included. As an active user of the multispace provider, this version with a single resource, and optional apply/destroy blocks was my favourite of the ones suggested :)
> 
> And I like the way you're handling retries with a recursive call to the same function; that's pretty neat. + exponential backoffs in retries are nice; I hadn't even considered that.
> 
> I think the only thing missing based on the discussions is the option to kick off a run, and then consider that good enough. i.e. do not wait for the run to finish before considering the resource created. Shouldn't be the default behaviour, but it's a very handy optional behaviour I use all the time from the multispace provider (or rather... from my fork of it).
> 
> But yeah, short version: looking good, and really looking forward to this!

@lucymhdavies thanks for the review! The PR is ready to be reviewed. Good point regarding the start and forget run option. I will get on that.

Uk1288:
**Should errors raised by stats APIs terminate the process or log warnings?**

Currently, while a run is in progress, we continuosly poll stats(e.g run status, run position in workspace or organization queue) to update the user on the state of the run. 
Errors arising from gathering these informative stats terminates the process on the terminal even though the run continues to proceed.

I am wondering what others think about these errors being logged as warnings as opposed to raising a terminating error, which in some cases might be a false error since the run continues to proceed.

tfe/resource_tfe_workspace_run.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+			"workspace": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"apply": {
+				Type: schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
nfagerlund, line 77: 🤔 My instincts are telling me that having these two manually duplicated inner block schemas is a maintenance hazard. Do you reckon it's worth de-duplicating, with a lil helper function that allocates and returns a `map[string]*schema.Schema`?

nfagerlund (changes requested):
That was a fun and helpful test session we did! I think I captured all the stuff that came up... hopefully these are all pretty straightforward!

tfe/resource_tfe_workspace_run.go
```diff
@@ -0,0 +1,192 @@
...
+		Read:          resourceTFEWorkspaceRunRead,
+		Update:        resourceTFEWorkspaceRunUpdate,
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"organization": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"workspace": {
```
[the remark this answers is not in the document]
Uk1288: I replaced the `workspace` with `workspace_id`, but then I quickly realized we still need the `organization` name for polling the capacity and org queue while the run is executing. Using the `...Client.Workspaces.ReadByID` does not return the org name. So if we are going with `workspace_id` we still the 2 args, i.e both `workspace_id` and `organization` name. I have made the update to depict this situation

tfe/resource_tfe_workspace_run.go
Uk1288: Sure, that makes sense to me

tfe/resource_tfe_workspace_run.go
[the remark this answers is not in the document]
Uk1288: Ignore the comment above, the workspace object includes the org name in its relation.

Uk1288:
**Followup BUG FIXED** 
Previously, when using the `tfe_workspace_run` with only the `destroy` block set as in the code block below, the SDK returns an error: `Provider produced inconsistent result after apply`. 

```
resource "tfe_workspace_run" "root_A" {
  workspace_id    = "ws_root_A_ID"

  destroy { 
    manual_confirm = false
  }
}
```

This happens because without specifying the `apply` block no resource is created as such `d.SetId()` is never called with a non-empty string - i.e the runID in this case. 

Screen Shot 2023-04-26 at 12 36 27 PM

To preserve the apply, destroy rule of creating & destroying resources, the apply block must be mandatory, to ensure that after a successful apply, a non-empty ID is set and the destroy finds a created resource.

lucymhdavies:
It should be possible to allow omiting the Apply block (i.e. using `tfe_workspace_run` to _only_ trigger a destroy) without seeing that error. My fork of the multispace provider does that (using `no-op` as the resource ID).

https://github.com/lucymhdavies/terraform-provider-multispace/blob/lmhd-main/internal/provider/resource_run.go#L110-L121

This way, a dummy resource _is_ created in the state, but that resource doesn’t refer to anything in the real world.
(But destroying the resource triggers a destroy of the workspace we’re managing)

Whether an ID of `no-op` is ideal for this… probably not. Maybe want to do something similar to whatever the `null_resource` does here.
(not sure off-hand what `null_resource` does for ID)

Either way… it should be possible to get it working :)

lucymhdavies:
Looks like null_resource literally just picks a random number.
https://github.com/hashicorp/terraform-provider-null/blob/main/internal/provider/resource.go#L72

So that might work 🤔

Uk1288:
> Looks like null_resource literally just picks a random number. https://github.com/hashicorp/terraform-provider-null/blob/main/internal/provider/resource.go#L72
> 
> So that might work 🤔

This looks promising, I will give it a shot! Thanks for making the suggestion.

tfe/resource_tfe_workspace_run.go
```diff
@@ -0,0 +1,157 @@
...
+	return createWorkspaceRun(d, meta, isDestroyRun, currentRetryAttempts)
+}
+
+func resourceTFEWorkspaceRunUpdate(d *schema.ResourceData, meta interface{}) error {
+	// update is a noop since this resource only creates a run during a destroy or an initial apply phase
+	return nil
+}
+
+func resourceTFEWorkspaceRunRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(ConfiguredClient)
```
nfagerlund, line 106: Here's where the problem was: if a run has no apply (and hence a random ID), the Read fails with a 404. If that causes us to `SetID("")`, then the resource gets deleted from the state, and during a Terraform destroy, that means we decide the resource is already gone and hence there's nothing to do during the destruction! It's absent during the destroy plan, and you don't get a workspace destroy run. 

I think the solution is to just bail early if we know there shouldn't be any run to read. That leaves the ID (and hence resource) in place, so we get our proper destroy on schedule. This patch worked for me. 

```suggestion
	// First check whether this is a destroy-only run
	_, ok := d.GetOk("apply")
	if !ok {
		// If there's no apply, then there won't be anything to "read" until we
		// do a destroy run. Return now and leave the ID alone, so that we keep
		// the resource in the state and get a destroy run when the time comes.
		log.Printf("[DEBUG] Run %s (random ID) has no apply; nothing to read for refresh", d.Id())
		return nil
	}

	config := meta.(ConfiguredClient)
```

tfe/resource_tfe_workspace_run.go
```diff
@@ -0,0 +1,157 @@
...
+
+	log.Printf("[DEBUG] Read run for: %s", d.Id())
+	runID := d.Id()
+	_, err := config.Client.Runs.Read(ctx, runID)
+	if err != nil {
+		if errors.Is(err, tfe.ErrResourceNotFound) {
+			log.Printf("[DEBUG] Run %s does not exist", d.Id())
```
nfagerlund, line 113: The whole idea of a run as a resource is juuuust askew enough from Terraform's normal resource-centric view of the world that I'm not 100% positive that this standard "remove from state if 404 on read" logic makes sense. As long as we've got the early out I suggested above, I think it's probably fine, but we might as well make our logic crystal clear in the comments, since the correct behavior (or even how this situation could come about) is not at all obvious to a newcomer to this code.

```suggestion
			// It would be very strange for this to happen, since runs can't
			// normally be deleted independently. But this *probably* means we
			// never performed the initial apply, so we'll remove the missing
			// run from the state to force an apply to happen.
			log.Printf("[DEBUG] Run %s does not exist", d.Id())
```

nfagerlund (changes requested):
Okay!! I gave this a workout, and found two problems: 

### Destroy run gets skipped when there's no apply

There's a straightforward answer to this one, though! See suggestions below. I tried that suggested patch and it works as intended. 

### Rare race condition for workspace safe delete

A destroy run actually creates a new state version — a basically empty one. And that new state version has to get processed to count resources, before the workspace will accept a safe-delete request; if you try while it's still cranking, you'll get 409 conflict with `Latest workspace state is being processed to discover resources, please try again later`. 

So, if you're destroying both a run and a workspace (i.e. doing the normal), there's a race if Terraform tries to delete the workspace too quickly after finishing the workspace destroy run. 🥲😂

You can easily get around that by setting `force_delete = true` on the workspace! But, I feel like part of the appeal of managing the destroy yourself is that you should be able to leave safe_delete turned on. 😭

What do you think we should do about that? My current thoughts are: 

- For now, maybe say something about it in the resource docs, to suggest force_deleting if you're doing this? 
- But ACTUALLY, the tfe_workspace resource SHOULD be handling a "hold on a sec" 409 very differently from a "hell no" 409! That resource can totally be destroyed safely, it just needs to cool its jets and wait in a timeout loop for a few minutes (eventually giving up if it takes like 10m, probably). That's way outside the scope of this PR, but I'd like your take on it anyway. 

### Anyway,

If you're satisfied with the missing-destroy fix, merge that and I'll ✅, and we can iterate going forward as people get a chance to play with this a bit more. The "still processing" 409 is a general-purpose problem with workspace that predates this new resource (although it makes it more likely to be exposed), so I won't block on that. 

I'd also like to expand the docs a bit to give some guidance on what we think this resource is useful for, to set peoples expectations, but we're pretty busy with a BUNCH of stuff right now, so it's probably better to just merge while it's hot and come back to that as we get some breathing room.

nfagerlund (approved):
All right, as discussed in slack: this resource is probably good to ship at this point, and we've got another issue (#876) to chase down the race condition on workspace safe-delete. 

@Uk1288 thanks again for your patience shepherding this all the way through!

lucymhdavies (approved):
Excited for this new resource. Great work getting this in place :)

lucymhdavies:
And we're live! Great work everyone :)

https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/workspace_run

PR #792, opened 2023-02-06
Clarify <organization> usage in vcs_repo config blocks

### LINKED ISSUE #789: Clearer documentation regarding tfe_workspace vcs_repo identifier arguement.
#### Use-cases

The documentation for the tfe_workspace here https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/workspace#vcs_repo says the vcs_repo identifier argument should be in the format of ```<organization>/<repository> ``` however that it isn't entirely clear if you mean the organization within TFC or in your VCS. It turned out to be the later. I recommend you make it clear that it's NOT the organization in TFC.

#### Proposal

I propose that you explicitly state that the ```<organization>```  is not referring to the organization name in TFC but in your VCS.

brandonc:
## Description

We should be try to differentiate between VCS organization from TFC organization more distinctly.

Closes #789


annawinkler (approved):
✨

PR #794, opened 2023-02-07
Add comment about the test setup and env vars

annawinkler:
## Description

Add a comment about where the github environment variables are set when the tests are run.


nfagerlund (approved):
I'm not actually sure we need this here, given that these vars are documented in the docs/test.md page and we found them pretty quickly as soon as we started wondering where they were. But also, no harm in it.

PR #832, opened 2023-03-24
Added some additional guidance around source_url and source_name

lucymhdavies:
Given that you can't update these in-place, any changes to these parameters will result in the workspace being destroyed then recreated.

Probably most people do not want to do that, so an ignore_changes lifecycle meta-argument is a good idea.

### DIFF
--- a/website/docs/r/workspace.html.markdown
+++ b/website/docs/r/workspace.html.markdown
@@ -98,7 +98,7 @@ The following arguments are supported:
    Requires `source_name` to also be set.
    **Note:** The API does not (currently) allow this to be updated after a
    workspace has been created, so modifying this value will result in the
-   workspace being replaced.
+   workspace being replaced. To disable this, use an [ignore changee](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) lifecycle meta-argument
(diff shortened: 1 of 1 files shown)


website/docs/r/workspace.html.markdown
```diff
@@ -98,7 +98,7 @@ The following arguments are supported:
    Requires `source_name` to also be set.
    **Note:** The API does not (currently) allow this to be updated after a
    workspace has been created, so modifying this value will result in the
-   workspace being replaced.
+   workspace being replaced. To disable this, use an [ignore changee](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) lifecycle meta-argument
```
brandonc, line 101:
```suggestion
   workspace being replaced. To disable this, use an [ignore changes](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) lifecycle meta-argument
```

brandonc: Thanks for adding this callout!

lucymhdavies: Good catch!

nfagerlund (approved):

PR #844, opened 2023-04-10
Add Optional Expiry to Auth Tokens

juliannatetreault:
## Description
Allow an optional `expired_at` to be set on authentication API tokens via the provider.

## Testing plan

- Ensure that organization and team API tokens that omit the optional `expired_at` field still work as expected.
- Ensure that an `expired_at` can be set on organization and team API tokens and that once set, the expiration date displays in the UI (`/app/org-name/authentication-tokens`) with the expiration date that was provided.
- Ensure that an invalid `expired_at` cannot be set on organization and team API tokens and that if attempted to be set, an error is returned.

## External links

- [Related go-tfe PR](https://github.com/hashicorp/go-tfe/pull/672)
- [API documentation](https://github.com/hashicorp/terraform-docs-common/pull/321)

#### `resource_tfe_organization_token_test.go` results:

```
=== RUN   TestAccTFEOrganizationToken_basic
--- PASS: TestAccTFEOrganizationToken_basic (8.99s)
PASS

Process finished with the exit code 0
...
=== RUN   TestAccTFEOrganizationToken_existsWithoutExpiry
--- PASS: TestAccTFEOrganizationToken_existsWithoutExpiry (18.02s)
PASS

Process finished with the exit code 0
...

=== RUN   TestAccTFEOrganizationToken_existsWithExpiry
--- PASS: TestAccTFEOrganizationToken_existsWithExpiry (9.48s)
PASS

Process finished with the exit code 0
```

#### `resource_tfe_team_token_test.go` results:

```
=== RUN   TestAccTFETeamToken_basic
--- PASS: TestAccTFETeamToken_basic (10.22s)
PASS

Process finished with the exit code 0
...
=== RUN   TestAccTFETeamToken_existsWithoutExpiry
--- PASS: TestAccTFETeamToken_existsWithoutExpiry (10.95s)
PASS

Process finished with the exit code 0
...
=== RUN   TestAccTFETeamToken_existsWithExpiry
--- PASS: TestAccTFETeamToken_existsWithExpiry (10.51s)
PASS

Process finished with the exit code 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ FEATURES:
 * `r/tfe_team`: Add attribute `manage_membership` to `organization_access` on `tfe_team` by @JarrettSpiker ([#801](https://github.com/hashicorp/terraform-provider-tfe/pull/801))
 * **New Resource**: `r/tfe_workspace_run` manages create and destroy lifecycles in a workspace, by @uk1288 ([#786](https://github.com/hashicorp/terraform-provider-tfe/pull/786))
 * `r/tfe_variable`: Add a `readable_value` attribute, which will provide an un-redacted representation of the variable's value in plan outputs if the variable is not sensitive, and which may be referenced by downstream resources by @JarrettSpiker ([#801](https://github.com/hashicorp/terraform-provider-tfe/pull/867))
+* `r/tfe_organization_token`: Add optional `expired_at` field to organization tokens by, @juliannatetreault ([#844](https://github.com/hashicorp/terraform-provider-tfe/pull/844))
+* `r/tfe_team_token`: Add optional `expired_at` field to team tokens by, @juliannatetreault ([#844](https://github.com/hashicorp/terraform-provider-tfe/pull/844))
 
 ENHANCEMENTS:
 
@@ -14,10 +16,6 @@ BUG FIXES:
 * `r/tfe_variable`: Don't silently erase or override the `value` of a sensitive variable on changes to other attributes when `ignore_changes = [value]` is set, by @nfagerlund ([#873](https://github.com/hashicorp/terraform-provider-tfe/pull/873), fixing issue [#839](https://github.com/hashicorp/terraform-provider-tfe/issues/839))
 
 ## v0.44.1 (April 21, 2023)
-
-FEATURES:
-* `r/tfe_organization_token`: Add optional `expired_at` field to organization tokens by, @juliannatetreault ([#844](https://github.com/hashicorp/terraform-provider-tfe/pull/844))
-* `r/tfe_team_token`: Add optional `expired_at` field to team tokens by, @juliannatetreault ([#844](https://github.com/hashicorp/terraform-provider-tfe/pull/844))
 
 BUG FIXES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ BUG FIXES:
 
 ## v0.44.1 (April 21, 2023)
 
+FEATURES:
+* `r/tfe_organization_token`: Add optional `expired_at` field to organization tokens by, @juliannatetreault ([#844](https://github.com/hashicorp/terraform-provider-tfe/pull/844))
+* `r/tfe_team_token`: Add optional `expired_at` field to team tokens by, @juliannatetreault ([#844](https://github.com/hashicorp/terraform-provider-tfe/pull/844))
+
 BUG FIXES:
 
 * Fixed a documentation bug in the new `r/tfe_no_code_module` resource, incorrectly labelling the attribute `registry_module` as `module`
--- a/tfe/resource_tfe_organization_token.go
+++ b/tfe/resource_tfe_organization_token.go
@@ -78,13 +78,11 @@ func resourceTFEOrganizationTokenCreate(d *schema.ResourceData, meta interface{}
 	options := tfe.OrganizationTokenCreateOptions{}
 
 	// Check whether the optional expiry was provided.
-	_, expiredAtProvided := d.GetOk("expired_at")
+	expiredAt, expiredAtProvided := d.GetOk("expired_at")
 
 	// If an expiry was provided, parse it and update the options struct.
 	if expiredAtProvided {
-		expiredAt := d.Get("expired_at").(string)
-
-		expiry, err := time.Parse(time.RFC3339, expiredAt)
+		expiry, err := time.Parse(time.RFC3339, expiredAt.(string))
 
 		options.ExpiredAt = &expiry
 
--- a/tfe/resource_tfe_organization_token.go
+++ b/tfe/resource_tfe_organization_token.go
@@ -73,6 +73,7 @@ func resourceTFEOrganizationTokenCreate(d *schema.ResourceData, meta interface{}
 		}
 		log.Printf("[DEBUG] Regenerating existing token for organization: %s", organization)
 	}
+
 	if err != nil {
 		expiredAt := d.Get("expired_at").(string)
 		return fmt.Errorf("%s must be a valid date or time", expiredAt)
--- a/tfe/resource_tfe_organization_token.go
+++ b/tfe/resource_tfe_organization_token.go
@@ -7,10 +7,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log"
-
 	tfe "github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"log"
+	"time"
 )
 
 func resourceTFEOrganizationToken() *schema.Resource {
@@ -45,7 +45,7 @@ func resourceTFEOrganizationToken() *schema.Resource {
 			"expired_at": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Computed: true,
+				ForceNew: true,
 			},
 		},
 	}
@@ -74,12 +74,26 @@ func resourceTFEOrganizationTokenCreate(d *schema.ResourceData, meta interface{}
 		log.Printf("[DEBUG] Regenerating existing token for organization: %s", organization)
 	}
 
-	if err != nil {
+	// Get the token create options.
+	options := tfe.OrganizationTokenCreateOptions{}
+
+	// Check whether the optional expiry was provided.
+	_, expiredAtProvided := d.GetOk("expired_at")
+
+	// If an expiry was provided, parse it and update the options struct.
+	if expiredAtProvided {
 		expiredAt := d.Get("expired_at").(string)
-		return fmt.Errorf("%s must be a valid date or time", expiredAt)
+
+		expiry, err := time.Parse(time.RFC3339, expiredAt)
+
+		options.ExpiredAt = &expiry
+
+		if err != nil {
+			return fmt.Errorf("%s must be a valid date or time, provided in iso8601 format", expiredAt)
+		}
 	}
 
-	token, err := config.Client.OrganizationTokens.Create(ctx, organization)
+	token, err := config.Client.OrganizationTokens.CreateWithOptions(ctx, organization, options)
 	if err != nil {
 		return fmt.Errorf(
 			"error creating new token for organization %s: %w", organization, err)
--- a/tfe/resource_tfe_organization_token.go
+++ b/tfe/resource_tfe_organization_token.go
@@ -41,6 +41,12 @@ func resourceTFEOrganizationToken() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+
+			"expired_at": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -66,6 +72,10 @@ func resourceTFEOrganizationTokenCreate(d *schema.ResourceData, meta interface{}
 			return fmt.Errorf("a token already exists for organization: %s", organization)
 		}
 		log.Printf("[DEBUG] Regenerating existing token for organization: %s", organization)
+	}
+	if err != nil {
+		expiredAt := d.Get("expired_at").(string)
+		return fmt.Errorf("%s must be a valid date or time", expiredAt)
 	}
 
 	token, err := config.Client.OrganizationTokens.Create(ctx, organization)
--- a/tfe/resource_tfe_organization_token_test.go
+++ b/tfe/resource_tfe_organization_token_test.go
@@ -102,7 +102,6 @@ func TestAccTFEOrganizationToken_existsWithForce(t *testing.T) {
 func TestAccTFEOrganizationToken_existsWithoutExpiry(t *testing.T) {
 	token := &tfe.OrganizationToken{}
 	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
-	expiredAt := "null"
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -115,7 +114,7 @@ func TestAccTFEOrganizationToken_existsWithoutExpiry(t *testing.T) {
 					testAccCheckTFEOrganizationTokenExists(
 						"tfe_organization_token.foobar", token),
 					resource.TestCheckResourceAttr(
-						"tfe_organization_token.expiry", "expired_at", expiredAt),
+						"tfe_organization_token.expiry", "expired_at", ""),
 				),
 			},
 		},
@@ -287,7 +286,6 @@ resource "tfe_organization" "foobar" {
 
 resource "tfe_organization_token" "foobar" {
   organization = tfe_organization.foobar.id
-  expired_at = null
 }
 
 resource "tfe_organization_token" "error" {
--- a/tfe/resource_tfe_organization_token_test.go
+++ b/tfe/resource_tfe_organization_token_test.go
@@ -125,7 +125,7 @@ func TestAccTFEOrganizationToken_existsWithoutExpiry(t *testing.T) {
 func TestAccTFEOrganizationToken_existsWithExpiry(t *testing.T) {
 	token := &tfe.OrganizationToken{}
 	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
-	expiredAt := "2051-04-11T23:15:59+00:00"
+	expiredAt := "2051-04-11T23:15:59Z"
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -308,7 +308,7 @@ resource "tfe_organization_token" "foobar" {
 
 resource "tfe_organization_token" "expiry" {
   organization  = tfe_organization.foobar.id
-  expired_at 	= "2051-04-11T23:15:59+00:00"
+  expired_at 	= "2051-04-11T23:15:59Z"
 }`, rInt)
 }
 
--- a/tfe/resource_tfe_organization_token_test.go
+++ b/tfe/resource_tfe_organization_token_test.go
@@ -110,7 +110,7 @@ func TestAccTFEOrganizationToken_existsWithoutExpiry(t *testing.T) {
 		CheckDestroy: testAccCheckTFEOrganizationTokenDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccTFEOrganizationToken_basic(rInt),
+				Config: testAccTFEOrganizationToken_existsWithoutExpiry(rInt),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckTFEOrganizationTokenExists(
 						"tfe_organization_token.foobar", token),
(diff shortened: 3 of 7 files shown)


tfe/resource_tfe_organization_token.go
```diff
@@ -68,6 +74,11 @@ func resourceTFEOrganizationTokenCreate(d *schema.ResourceData, meta interface{}
 		log.Printf("[DEBUG] Regenerating existing token for organization: %s", organization)
 	}
 
+	if err != nil {
```
JarrettSpiker, line 77: ❓ how do we know here that the `err` is that the expiry is invalid? As opposed to something like the user not being authorized to create new tokens

tfe/resource_tfe_organization_token.go
```diff
@@ -41,6 +41,12 @@ func resourceTFEOrganizationToken() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+
+			"expired_at": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
```
JarrettSpiker, line 48: ❓ why is this computed? My understanding of computed is that it is for properties which cant be known before the object is created.

https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#computed

website/docs/r/organization_token.html.markdown
```diff
@@ -28,6 +28,7 @@ The following arguments are supported:
 * `force_regenerate` - (Optional) If set to `true`, a new token will be
   generated even if a token already exists. This will invalidate the existing
   token!
+* `expired_at` - (Optional) A date and time in which the organization token will expire.
```
JarrettSpiker, line 31: I think that this would be a good place to include information on what format we expect this value to be passed in, maybe with examples. For example, we should specify that this value should be UTC https://hashicorp.atlassian.net/browse/TF-5300

I think it would also be worth mentioning that if this is not set that the token will never expire

tfe/resource_tfe_organization_token_test.go
```diff
@@ -99,6 +99,67 @@ func TestAccTFEOrganizationToken_existsWithForce(t *testing.T) {
...
+				),
+			},
+
+			{
+				Config: testAccTFEOrganizationToken_existsWithExpiry(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckTFEOrganizationTokenExists(
+						"tfe_organization_token.expiry", token),
+					resource.TestCheckResourceAttr(
+						"tfe_organization_token.expiry", "organization", orgName),
```
JarrettSpiker, line 156: I think it would make sense here to check the the expiry value from the configuration has been set on the created `resource`

tfe/resource_tfe_organization_token_test.go
```diff
@@ -215,3 +276,36 @@ resource "tfe_organization_token" "regenerated" {
...
+resource "tfe_organization" "foobar" {
+  name  = "tst-terraform-%d"
+  email = "admin@company.com"
+}
+
+resource "tfe_organization_token" "foobar" {
+  organization = tfe_organization.foobar.id
+}
+
+resource "tfe_organization_token" "error" {
```
JarrettSpiker, line 291: ❓ why would this configuration result in an error? It looks like we have set [ExpiredAt](https://github.com/hashicorp/terraform-provider-tfe/pull/844/files#diff-3ba6ff1c80cd647a2c785e0c5f5fb72a8aa985181fb2689b41686131d98bfb0fR45) as an optional attibute (which I think is correct!). So I would expect that a configuration like this would create a token which just doesnt expire

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## v0.44.1 (April 21, 2023)
 
+FEATURES:
+* `r/tfe_organization_token`: Add optional `expired_at` field to organization tokens by, @juliannatetreault ([#844](https://github.com/hashicorp/terraform-provider-tfe/pull/844))
```
JarrettSpiker, line 4: Not urgent (and probably not worth doing until we are actually ready to release this), but a heads up that we will need to change where in the changelog this is, up to the "Unreleased" section

tfe/resource_tfe_organization_token_test.go
```diff
@@ -99,6 +99,52 @@ func TestAccTFEOrganizationToken_existsWithForce(t *testing.T) {
...
+	rInt := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
+	orgName := fmt.Sprintf("tst-terraform-%d", rInt)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckTFEOrganizationTokenDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccTFEOrganizationToken_basic(rInt),
```
JarrettSpiker, line 113: ❓ was this meant to sue the new `testAccTFEOrganizationToken_existsWithoutExpiry`?

JarrettSpiker:
If you would like the make the compile errors go away, and allow tests to run or allow other developers to clone and test this, you can run + commit

```
go get github.com/hashicorp/go-tfe@10f40dcbde314dbc81c1c4e35a0aa89823ca0ff0
```

This will tell go to use your draft version of go-tfe from that commit.

It will need to be updated to an actual released go-tfe version before this can be merged....but it will need to be updated regardless when the actual version of go-tfe is released, so 🤷 no extra work needed by pointing to an unreleased version of go-tfe in the mean time

tfe/resource_tfe_organization_token_test.go
juliannatetreault: Yup! 🤦‍♀️

website/docs/r/team_token.html.markdown
```diff
@@ -32,6 +32,20 @@ The following arguments are supported:
...
   token!
+* `expired_at` - (Optional) A date and time in which the team token will expire. The expiration date must be passed in
+iso8601 format. If no expiration date is supplied, the expiration date will default to null and never expire.
+
+## Example Usage
+
+Basic usage:
+
+```hcl
+resource "tfe_team" "test" {
```
JarrettSpiker, line 43: I think that this is a copy-paste error, this looks like the configuration for a team, not a team token

tfe/resource_tfe_organization_token.go
```diff
@@ -68,7 +74,26 @@ func resourceTFEOrganizationTokenCreate(d *schema.ResourceData, meta interface{}
 		log.Printf("[DEBUG] Regenerating existing token for organization: %s", organization)
 	}
 
-	token, err := config.Client.OrganizationTokens.Create(ctx, organization)
+	// Get the token create options.
+	options := tfe.OrganizationTokenCreateOptions{}
+
+	// Check whether the optional expiry was provided.
+	_, expiredAtProvided := d.GetOk("expired_at")
```
JarrettSpiker, line 81: Small suggestion: I believe that the return value which is getting ignored here (the `_`) is actually going to be the expiredAt value if it exists. So we could change this to

```
expiredAt, expiredAtProvided := d.GetOk("expired_at")
```
and no longer need [line 85](https://github.com/hashicorp/terraform-provider-tfe/pull/844/files#diff-3ba6ff1c80cd647a2c785e0c5f5fb72a8aa985181fb2689b41686131d98bfb0fR85)

tfe/resource_tfe_organization_token.go
```diff
@@ -68,7 +74,26 @@ func resourceTFEOrganizationTokenCreate(d *schema.ResourceData, meta interface{}
...
+
+	// If an expiry was provided, parse it and update the options struct.
+	if expiredAtProvided {
+		expiredAt := d.Get("expired_at").(string)
+
+		expiry, err := time.Parse(time.RFC3339, expiredAt)
+
+		options.ExpiredAt = &expiry
+
+		if err != nil {
```
JarrettSpiker, line 91: Could we please add a test to check that we are failing correctly when the string provided as the expiry is not in an acceptable format (for both team and org tokens)?

tfe/resource_tfe_organization_token_test.go
```diff
@@ -99,6 +99,52 @@ func TestAccTFEOrganizationToken_existsWithForce(t *testing.T) {
...
+	orgName := fmt.Sprintf("tst-terraform-%d", rInt)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckTFEOrganizationTokenDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccTFEOrganizationToken_existsWithoutExpiry(rInt),
+				Check: resource.ComposeTestCheckFunc(
```
JarrettSpiker, line 114: Since we have a test here for checking that tokens can be created without an expiry, is it possible to add a check that the created token has a null expiration date? This would help guard against regressions where we dont return properly [on this line](https://github.com/hashicorp/terraform-provider-tfe/pull/844/files#diff-3ba6ff1c80cd647a2c785e0c5f5fb72a8aa985181fb2689b41686131d98bfb0fR92), for example.

The same would also be nice on the team tokens, if possible.

tfe/resource_tfe_organization_token_test.go
```diff
@@ -215,3 +261,36 @@ resource "tfe_organization_token" "regenerated" {
   force_regenerate = true
 }`, rInt)
 }
+
+func testAccTFEOrganizationToken_existsWithoutExpiry(rInt int) string {
```
JarrettSpiker, line 265: Not a big deal, but I believe we could just use `testAccTFEOrganizationToken_basic` in the place where we use this function.

Same for `testAccTFETeamToken_existsWithoutExpiry` and `testAccTFETeamToken_basic`

JarrettSpiker (changes requested):
One easy fix in the docs, and a few suggestions on cleaning up the code a bit.

If the requested tests are a bigger lift than I am thinking, let me know!

website/docs/r/team_token.html.markdown
juliannatetreault: Great catch! I've updated this.

tfe/resource_tfe_organization_token.go
juliannatetreault: Added! ✅

tfe/resource_tfe_organization_token_test.go
juliannatetreault: Done! ✅

tfe/resource_tfe_organization_token.go
juliannatetreault: Great suggestion! I've updated this for both the org and team token.

tfe/resource_tfe_organization_token_test.go
juliannatetreault: I decided to repurpose these tests to check that a created token can have a `null` expiry, per your suggestion [here](https://github.com/hashicorp/terraform-provider-tfe/pull/844#discussion_r1202900712)!

website/docs/r/team_token.html.markdown
```diff
@@ -32,6 +32,20 @@ The following arguments are supported:
...
+iso8601 format. If no expiration date is supplied, the expiration date will default to null and never expire.
+
+## Example Usage
+
+Basic usage:
+
+```hcl
+resource "tfe_team_token" "test" {
+  name         = "my-team-name"
+  organization = "my-org-name"
```
JarrettSpiker, line 45:
```suggestion
  team_id = "team-id"
```

JarrettSpiker (commented):
looking good to me, just one more small error in the docs

JarrettSpiker (approved):
This LGTM!

brandonc (commented):
I wonder if we should suggest such a dramatically long expiration period in the docs... and more generally I wonder if there is more guidance we can give for generating this format in particular.. Example: You can generate this string using the `date` tool in unix systems.. unfortunately this tool has different flags in linux and darwin and other systems:

darwin:
`date -Iseconds -v"+90d"`

linux:
`date -Iseconds -d"+90 days"`

Or you can use a terraform function that doesn't make it easy to work with long durations:

```
$ terraform console
> timeadd(timestamp(), "2160h")
"2023-08-28T23:04:47Z"
```

To open up a slightly broader discussion... I don't believe that this format is perfectly suitable for this use case because it must be static to be useful and doesn't lend itself to human readability. What I mean by static is that it would not make sense to invoke a terraform function like timestamp() to define this value because it would force a re-create on the resource every time it was evaluated. Nor is there an easy set of terraform functions to format a more human readable date... We expect users to statically define an expiration date in iso8601 format, which is hard to read and hard to write. So what if we chose a more human readable format that was also easy to produce? Like RFC 5322 (email format)

```
$ date -v"+90d" -R
Mon, 28 Aug 2023 17:10:48 -0600
```

This can be reliably parsed with go, looks nice in config, and we could even document the terraform format string for the expected format (it's `EEE, DD MMM YYYY hh:mm:ss ZZZZ`)

```
$ terraform console
> formatdate("EEE, DD MMM YYYY hh:mm:ss ZZZZ", timeadd(timestamp(), "2160h"))
```

Tell me what you think.
/cc @JarrettSpiker

JarrettSpiker:
@brandonc: @juliannatetreault  and I discussed this, and definitely agree that we can improve the documentation with some more realistic examples, and provide the user with some commands they can use to generate an appropriate timestamp. 

> To open up a slightly broader discussion... I don't believe that this format is perfectly suitable for this use case because it must be static to be useful and doesn't lend itself to human readability. 

I don't know that I totally agree here though. Maybe it is personal preference but I don't find `2023-08-28T23:04:47Z` harder to read than `Mon, 28 Aug 2023 17:10:48 -0600`... in both the date is apparent, and I am going to need to plug the time into some sort of time zone conversion tool if it is relevant.

And I think I would actually prefer to write the `2023-08-28T23:04:47Z` format, since I find it intuitively less ambiguous (e.g. no needing to guess what the correct format for the month name is).

I would be open to changing the format if there are strong feelings about it. The current format being alright could be just personal preference though...and with all things being equal I lean towards using the format that the API will be using.

On a related note, just googling around I found the [time provider](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/static)...do you think that that would be something worth recommending as a way to generate this timestamp? Something like:

```
resource "time_static" "example" {}

resource "tfe_organization_token" "token" {
    organization = data.tfe_organization.org.name
    expired_at = timeadd(time_static.example.rfc3339, "2160h")
}
```

brandonc:
@JarrettSpiker @juliannatetreault I am super into recommending the time provider in the documentation! I wouldn't mind keeping iso8601 if it makes the examples this simple! Plus, it has the added benefit of being the same standard format that timestamp(), etc. uses.

```
resource "time_rotating" "example" {
  rotation_days = 90
}

resource "tfe_organization_token" "token" {
    organization = data.tfe_organization.org.name
    expired_at = time_rotating.example.id
}
```

website/docs/r/organization_token.html.markdown
```diff
@@ -38,8 +38,33 @@ Basic usage:
...
+Generating the `expired_at` string using the date tool in unix systems (linux):
+```
+date -Iseconds -d"+30 days"
+```
+
+Generating the `expired_at` string using the `timeadd` Terraform function:
+```
+$ terraform console
+> timeadd(timestamp(), "720h")
+"2023-07-21T02:02:23Z"
```
brandonc, line 67: With the time_rotating example, I don't think all this detail is necessary anymore. Also, can you add the `expired_at` token below in the "Attributes Reference" with specific detail/example of the format? (both resources)

brandonc (changes requested):
👇

website/docs/r/organization_token.html.markdown
juliannatetreault: Good call out, will do!

website/docs/r/organization_token.html.markdown
```diff
@@ -28,11 +28,34 @@ The following arguments are supported:
...
   token!
+* `expired_at` - (Optional) A date and time in which the organization token will expire. The expiration date must be passed in
+  iso8601 format. If no expiration date is supplied, the expiration date will default to null and never expire.
 
 ## Attributes Reference
 
 * `id` - The ID of the token.
 * `token` - The generated token.
+* `expired_at` - The token's expiration date. The expiration date must be passed in
+iso8601 format. If no expiration date is supplied, the expiration date will default to null and never expire.
```
brandonc, line 39: I'm so sorry I gave you the wrong feedback. I was looking at a single commit and not the entire change. I didn't see this was already added to the section above and I got them mixed up.

```suggestion
```

brandonc, line 39:
```suggestion
* `expired_at` - The token's expiration date. The expiration date must be a date/time string in 
RFC3339 format (For example, "2024-12-31T23:59:59Z"). If no expiration date is supplied, the expiration date will default to null and never expire.
```

website/docs/r/team_token.html.markdown
```diff
@@ -32,11 +32,34 @@ The following arguments are supported:
...
   token!
+* `expired_at` - (Optional) A date and time in which the team token will expire. The expiration date must be passed in
+iso8601 format. If no expiration date is supplied, the expiration date will default to null and never expire.
 
 ## Attributes Reference
 
 * `id` - The ID of the token.
 * `token` - The generated token.
+* `expired_at` - The token's expiration date. The expiration date must be passed in
+iso8601 format. If no expiration date is supplied, the expiration date will default to null and never expire.
```
brandonc, line 43:
```suggestion
```

website/docs/r/team_token.html.markdown
```diff
@@ -32,11 +32,34 @@ The following arguments are supported:
 * `force_regenerate` - (Optional) If set to `true`, a new token will be
   generated even if a token already exists. This will invalidate the existing
   token!
+* `expired_at` - (Optional) A date and time in which the team token will expire. The expiration date must be passed in
+iso8601 format. If no expiration date is supplied, the expiration date will default to null and never expire.
```
brandonc, line 36:
```suggestion
* `expired_at` - The token's expiration date. The expiration date must be a date/time string in 
RFC3339 format (For example, "2024-12-31T23:59:59Z"). If no expiration date is supplied, the expiration date will default to null and never expire.
```

brandonc (changes requested):
I made a mistake!

website/docs/r/organization_token.html.markdown
```diff
@@ -28,6 +28,28 @@ The following arguments are supported:
 * `force_regenerate` - (Optional) If set to `true`, a new token will be
   generated even if a token already exists. This will invalidate the existing
   token!
+* `expired_at` - (Optional) The token's expiration date. The expiration date must be a date/time string in RFC3339
```
juliannatetreault, line 31: I've updated the description for `expired_at` per your suggestions, but opted to leave the "optional" here since the expiry is optional and seems consistent with the other optional arguments, but will happily remove it if preferred! Additionally, I moved the example usage back to where it initially was (beneath the arguments docs!) to keep it more inline with the expiry for context. 

Let me know what you think, @brandonc! 😄

website/docs/r/organization_token.html.markdown
```diff
@@ -28,6 +28,28 @@ The following arguments are supported:
...
+
+## Example Usage
+
+When a token has an expiry:
+
+```hcl
+resource "tfe_organization_token" "test" {
+  organization = "my-org-name"
+}
+
```
brandonc, line 43: I think this is an extra resource defined in the example. If you are trying to show examples with and without an expired_at, that would be OK, but they should be separated by different code blocks maybe

```suggestion
```

website/docs/r/organization_token.html.markdown
```diff
@@ -28,6 +28,28 @@ The following arguments are supported:
...
+resource "time_rotating" "example" {
+  rotation_days = 30
+}
+
+resource "tfe_organization_token" "test" {
+  organization = data.tfe_organization.org.name
+  expired_at = time_rotating.example.id
```
brandonc, line 50: OK so I tested this out and the docs for the provider seem to be wrong. `id` returns the base timstamp and `rotation_rfc3339` returns the offset timestamp. The examples should look like this:

```suggestion
  expired_at = time_rotating.example.rotation_rfc3339
```

brandonc (changes requested):
Very close 🎉

website/docs/r/organization_token.html.markdown
brandonc: Fix here https://github.com/hashicorp/terraform-provider-time/pull/137

juliannatetreault: Good to know, thank you! I've updated the examples

website/docs/r/organization_token.html.markdown
juliannatetreault: Whoops! That was an extra resource 🙈

brandonc (approved):
WTG

PR #846, opened 2023-04-11
Upgrades: terraform-plugin-go v0.15.0 and terraform-plugin-mux v0.9.0

nfagerlund:
## Description

A lil dependency bump. We'll want the newest versions of these for integrating the new framework, and since there's no particular reason to wait on them, I'll put them in their own little PR. 

These are the latest versions of these dependencies as of this writing. The main notable change from our perspective is the package reorganization that happened in mux v0.5.0, requiring some updates to calls in main.go and provider_test.go.

## Testing plan

1.  Normal operation of the provider ought to be unaffected in any way, as shown by the acceptance tests etc. 
2. If anyone has time to test this quickfast on something antique like 0.12, it'd probably be a good idea, but there oughtn't be anything that would affect compatibility. 

## External links

- [mux changelog](https://github.com/hashicorp/terraform-plugin-mux/blob/main/CHANGELOG.md)
- [plugin-go changelog](https://github.com/hashicorp/terraform-plugin-go/blob/main/CHANGELOG.md)


brandonc (approved):

nfagerlund:
There we go, that did the trick.

PR #836, opened 2023-03-29
[BETA] CRUD support for no-code registry modules

dsa0x:
## Description

This adds support for managing a no-code registry modules. It introduces a new resource called tfe_nocode_module which can be attached to a tfe_registry_module resource, to make it a no-code module. The user can also define variable options for the no-code module, adding a finite number of values allowed for the variable.

## Testing plan
Note: Because the organization needs to have permissions to the featureflag, and the organization will be created during the test, you'll have to test this against a local atlas instance where the flag is defaulted to true for everyone.
- Create a User in TFC 
- Create an Organization (e.g `my-org`) and ensure they have access to the new no-code API (you can do that in launchdarkly)
- Generate an API token for your user in TFC
- Point the dev environment to your user and TFC instance. e.g
```go
// tfe/resource_tfe_no_code_module_test.go
func init() {
	os.Setenv("TF_ACC", "1")
	os.Setenv("ENABLE_BETA", "1")
	os.Setenv("TFE_ADDRESS", "https://tfcdev-foobar.eu.ngrok.io")
	os.Setenv("TFE_HOSTNAME", "tfcdev-foobar.eu.ngrok.io")
	os.Setenv("TFE_TOKEN", "your-token")
}
```

- Manually test with various configurations similar to:
    - Basic no-code module
        ```hcl      
        resource "tfe_organization" "foobar" {
        name  = "my-org"
        email = "admin@company.com"
        }
        
        resource "tfe_registry_module" "foobar" {
	        organization    = tfe_organization.foobar.id
	        module_provider = "my_provider"
	        name            = "test_module"
        }
        
        resource "tfe_nocode_module" "foobar" {
	        organization = tfe_organization.foobar.id
	        registry_module = tfe_registry_module.foobar.id
        }
        ```

   - With variable options
      ```hcl
      resource "tfe_organization" "foobar" {
      name  = "my-org"
      email = "admin@company.com"
      }
      
      resource "tfe_registry_module" "foobar" {
	      organization    = tfe_organization.foobar.id
	      module_provider = "my_provider"
	      name            = "test_module"
      }
      
      resource "tfe_nocode_module" "foobar" {
	      organization = tfe_organization.foobar.id
	      registry_module = tfe_registry_module.foobar.id
      
	      variable_options {
		      name    = "ami"
		      type    = "string"
		      options = [ "ami-0", "ami-1", "ami-2" ]
	      }
      
	      variable_options {
		      name    = "region"
		      type    = "string"
[...]

PR #859, opened 2023-04-20
r/tfe_no_code_module

lucymhdavies:
fix docs to update module argument to registry_module, as mentioned at bottom of https://github.com/hashicorp/terraform-provider-tfe/pull/836


lucymhdavies:
Ah. Looks like a dupe of https://github.com/hashicorp/terraform-provider-tfe/pull/858

Issue #726, opened 2022-12-20
Alternative way to provide non-sensitive values to tfe_variable and tfe_policy_set_parameter


brandonc:
#### Use-cases

Currently, tfe_variable and tfe_policy_set_parameter resources accept a sensitive value argument that cannot be directly evaluated within the plan diff. This is a limitation of the protocol, which doesn't allow a provider to specify when a particular _value_ is sensitive and instead the provider must mark the entire schema as sensitive, just in case it is.

#### Proposal

An alternative schema that can be used when a value is known to not be sensitive.

PR #867, opened 2023-04-25
Add a readable_value attribute to tfe_variables

JarrettSpiker:
## Description

Add a readable_value attribute to the tfe_variable resource. This aims to solve the usability problem where variable values, and all values which depend on them, are redacted from plan outputs regardless of if the variable is sensitive.

This new computed property would only be set if the variable is non-sensitive, and would not be redacted from outputs. Any other resource attribute depending on it (instead of on `value` directly) would also not be redacted from plan outputs.

## Testing plan

1. Changing the value of a non-sensitive variable should update the readable_value
1. Changing the value of a sensitive variable should leave the readable_value blank
2. Changing a variable from sensitive to non-sensitive should set the readable_value
1. Changing a variable from non-sensitive to sensitive should clear the readable_value

## External links

- [JIRA](https://hashicorp.atlassian.net/browse/TF-5533)

```
❯ TF_ACC=1 go test github.com/hashicorp/terraform-provider-tfe/tfe  -run TestAccTFEVariable_ -tags=integration -v
--- PASS: TestAccTFEVariable_basic (8.11s)
--- PASS: TestAccTFEVariable_basic_variable_set (9.54s)
--- PASS: TestAccTFEVariable_update (13.13s)
--- PASS: TestAccTFEVariable_update_key_sensitive (13.61s)
=== RUN   TestAccTFEVariable_readable_value
--- PASS: TestAccTFEVariable_readable_value (15.37s)
=== RUN   TestAccTFEVariable_readable_value_becomes_sensitive
--- PASS: TestAccTFEVariable_readable_value_becomes_sensitive (21.37s)
=== RUN   TestAccTFEVariable_varset_readable_value
--- PASS: TestAccTFEVariable_varset_readable_value (18.53s)
=== RUN   TestAccTFEVariable_varset_readable_value_becomes_sensitive
--- PASS: TestAccTFEVariable_varset_readable_value_becomes_sensitive (27.12s)
--- PASS: TestAccTFEVariable_import (8.83s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/tfe 135.814s

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ BREAKING CHANGES:
 FEATURES:
 * `r/tfe_team`: Add attribute `manage_membership` to `organization_access` on `tfe_team` by @JarrettSpiker ([#801](https://github.com/hashicorp/terraform-provider-tfe/pull/801))
 * **New Resource**: `r/tfe_workspace_run` manages create and destroy lifecycles in a workspace, by @uk1288 ([#786](https://github.com/hashicorp/terraform-provider-tfe/pull/786))
+* `r/tfe_variable`: Add a `readable_value` attribute, which will provide an un-redacted representation of the variable's value in plan outputs if the variable is not sensitive, and which may be referenced by downstream resources by @JarrettSpiker ([#801](https://github.com/hashicorp/terraform-provider-tfe/pull/867))
 
 ENHANCEMENTS:
 
--- a/tfe/resource_tfe_variable.go
+++ b/tfe/resource_tfe_variable.go
@@ -234,7 +234,7 @@ func (r *resourceTFEVariable) Schema(ctx context.Context, req resource.SchemaReq
 			"readable_value": schema.StringAttribute{
 				Optional: true,
 				Computed: true,
-				Description: "A non-sensitive read-only copy of the variable value, which can be viewed or references " +
+				Description: "A non-sensitive read-only copy of the variable value, which can be viewed or referenced " +
 					"in plan outputs without being redacted. Will only be present if the variable is not sensitive",
 				PlanModifiers: []planmodifier.String{
 					&updateReadableValuePlanModifier{},
--- a/tfe/resource_tfe_variable.go
+++ b/tfe/resource_tfe_variable.go
@@ -40,6 +40,7 @@ type modelTFEVariable struct {
 	ID            types.String `tfsdk:"id"`
 	Key           types.String `tfsdk:"key"`
 	Value         types.String `tfsdk:"value"`
+	ReadableValue types.String `tfsdk:"readable_value"`
 	Category      types.String `tfsdk:"category"`
 	Description   types.String `tfsdk:"description"`
 	HCL           types.Bool   `tfsdk:"hcl"`
@@ -67,6 +68,9 @@ func modelFromTFEVariable(v tfe.Variable, lastValue types.String) modelTFEVariab
 	// instead, because the API never lets us read it again.
 	if v.Sensitive {
 		m.Value = lastValue
+		m.ReadableValue = types.StringNull()
+	} else {
+		m.ReadableValue = m.Value
 	}
 	return m
 }
@@ -91,6 +95,9 @@ func modelFromTFEVariableSetVariable(v tfe.VariableSetVariable, lastValue types.
 	// instead, because the API never lets us read it again.
 	if v.Sensitive {
 		m.Value = lastValue
+		m.ReadableValue = types.StringNull()
+	} else {
+		m.ReadableValue = m.Value
 	}
 	return m
 }
@@ -222,6 +229,15 @@ func (r *resourceTFEVariable) Schema(ctx context.Context, req resource.SchemaReq
 						variableSetIDRegexp,
 						"must be a valid variable set ID (varset-<RANDOM STRING>)",
 					),
+				},
+			},
+			"readable_value": schema.StringAttribute{
+				Optional: true,
+				Computed: true,
+				Description: "A non-sensitive read-only copy of the variable value, which can be viewed or references " +
+					"in plan outputs without being redacted. Will only be present if the variable is not sensitive",
+				PlanModifiers: []planmodifier.String{
+					&updateReadableValuePlanModifier{},
 				},
 			},
 		},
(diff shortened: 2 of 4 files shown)


rhughes1:
@JarrettSpiker Looks like this partially solves https://github.com/hashicorp/terraform-provider-tfe/issues/726. Guess the same idea needs to be applied to the `policy_set_parameter` resource as well.

CHANGELOG.md
brandonc, line 1: Unfortunately, we've chosen this moment to [migrate tfe_variable to the plugin framework](https://github.com/hashicorp/terraform-provider-tfe/pull/873) so this is going to require a pretty epic rebase soon. :disappointed:

nfagerlund: Yeah sorry 😞 

...but at least the acceptance tests will only need single-line updates!

tfe/resource_tfe_variable.go
```diff
@@ -224,6 +231,18 @@ func (r *resourceTFEVariable) Schema(ctx context.Context, req resource.SchemaReq
 					),
 				},
 			},
+			"readable_value": schema.StringAttribute{
+				Optional: true,
+				Computed: true,
+				Description: "A non-sensitive read-only copy of the variable value, which can be viewed or references " +
```
nfagerlund, line 237:
```suggestion
				Description: "A non-sensitive read-only copy of the variable value, which can be viewed or referenced " +
```

tfe/resource_tfe_variable.go
```diff
@@ -224,6 +231,18 @@ func (r *resourceTFEVariable) Schema(ctx context.Context, req resource.SchemaReq
 					),
 				},
 			},
+			"readable_value": schema.StringAttribute{
+				Optional: true,
```
nfagerlund, line 235:
```suggestion
```

(...etc. etc., also removing the Validators and the whole custom validator implementation below)

website/docs/r/variable.html.markdown
```diff
@@ -91,6 +91,46 @@ in the state.
 ## Attributes Reference
 
 * `id` - The ID of the variable.
+* `readable_value` - Only present if the variable is non-sensitive. A copy of the value which will not be marked as sensitive in plan outputs.
+
+### Using readable_value
+
+While the `value` field may be referenced in other resources, for safety it is always treated as sensitive. This means that it will always be redacted from plan outputs, and any other resource attributes which depend on it will also be redacted.
+
+The `readable_value` attribute is not sensitive, and will not be redacted. This allows other resources to reference it, while keeping their plan outputs readable.
```
nfagerlund, line 100:
```suggestion
The `readable_value` attribute is not sensitive, and will not be redacted; instead, it will be null if the variable is sensitive. This allows other resources to reference it, while keeping their plan outputs readable.
```

Well, idk, that one counts as "take it or leave it," since you say it enough times elsewhere in the docs. 😅 I'm just role-playing as a paranoid who stops reading early and freaks out a bit.

nfagerlund (changes requested):
Hey, thanks for this! (AND for accommodating our awkwardly-timed rewrite!) I gave it a test-drive and it works just great. 👍🏼

However, you worked a bit harder than you needed to! 😅 In fact, you don't need that custom validator at all: if you OMIT BOTH of `Required` and `Optional`, the attribute becomes read-only, which I think is similar to what sdkv2 did. I did a rebuild and test-run after commenting out your validator field in the attribute schema and deleting `Optional: true`; it results in an error like this: 

```
╷
│ Error: Invalid Configuration for Read-Only Attribute
│
│   with tfe_variable.explodo,
│   on main.tf line 46, in resource "tfe_variable" "explodo":
│   46:   readable_value = "bunk!!!"
│
│ Cannot set value for this attribute as the provider has marked it as read-only. Remove the configuration line setting the value.
│
│ Refer to the provider documentation or contact the provider developers for additional information about configurable and read-only attributes that are supported.
╵
```

...which I think is exactly what you want!

tfe/resource_tfe_variable.go
JarrettSpiker: Wow, that is exactly what I was looking for! Thanks Nick!

tfe/resource_tfe_variable.go
```diff
@@ -707,11 +722,46 @@ func (r *resourceTFEVariable) ImportState(ctx context.Context, req resource.Impo
 	resp.Diagnostics.Append(diags...)
 }
 
+type updateReadableValuePlanModifier struct{}
+
+func (u *updateReadableValuePlanModifier) Description(ctx2 context.Context) string {
```
nfagerlund, line 727:
```suggestion
func (u *updateReadableValuePlanModifier) Description(ctx context.Context) string {
```

(And similar comment for the other uses of this below.)

I think it's probably better to use the conventional `ctx` name for these arguments... unless I'm missing something? 

We've still got a "default context" top-level `ctx` var hanging around in the package, and I notice that you're actually referencing _that_ object a few lines south of here when you reference `ctx` in the body of `PlanModifyString`. I'm guessing that's on purpose, but my understanding is: because these new framework functions can always rely on receiving an appropriate context from their caller, we don't need to mess around with having a top-scope "default context" at all — so it's actually good and proper for the function arguments to shadow the "default context" var that the sdkv2 provider relies on and just use whatever they got passed.

tfe/resource_tfe_variable_test.go
```diff
@@ -189,6 +191,193 @@ func TestAccTFEVariable_update_key_sensitive(t *testing.T) {
...
+
+	// Test that we throw an error if the user attempts to explicitly set the readable_value in the configuration
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV5ProviderFactories: testAccMuxedProviders,
+		CheckDestroy:             testAccCheckTFEVariableDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccTFEVariable_readablevalue_setExplicitly(rInt, variableValue1),
+				ExpectError: regexp.MustCompile(`Invalid Configuration for Read-Only Attribute`),
```
nfagerlund, line 375: Given that read-only attributes are a built-in and fairly fundamental feature of the framework, I think we can get away without having this acceptance test! It would be prudent to have if we were using a custom validator, but now that that's out I think we can just remove it. 👍🏼

nfagerlund (changes requested):
I caught two more minor things in my final once-over, but after that I think we're good here. LMK if I misunderstood anything with that `ctx2` thing.

tfe/resource_tfe_variable.go
JarrettSpiker: You are totally right! This was an artifact of GoLand automatically completing the interface implementation for me, and I hadnt noticed. I will update it all to be `ctx`

nfagerlund (approved):
Ok, this is in great shape! Thank you again for this.

That test failure is a known flake, and coincidentally this instance of it finally tells me what I need to know in order to (probably) fix it... but that's for another PR. I'll re-run 'til it passes and then merge this.

brandonc:
@maxforasteiro after some investigation, we found the cause to be a change to the variables endpoint in Terraform Cloud that was mistaking your HCL variable value for JSON. This helped us uncover some design problems so we rolled back that change in Terraform Cloud. Can you try your apply again and let me know if you see the same error?

maxforasteiro:
hey @brandonc, it worked yesterday right before I made the comment here when I changed the version of tfe from `0.45.0` to `0.44.1` hence why I believed they were connected. I can try to change the version back to `0.45.0` to see if it still persists.

PR #873, opened 2023-05-04
Integrate terraform-plugin-framework, and rewrite the tfe_variable resource

### LINKED ISSUE #839: sensitive tfe_variable values are wiped upon description change
Terraform Enterprise v202209-2

#### Terraform version

```plaintext
Terraform v1.4.2
+ provider registry.terraform.io/hashicorp/tfe v0.43.0
```

##### Parent Workspace

```terraform
// TFE_TOKEN, TFE_HOSTNAME set on workspace
provider "tfe" {}

data "tfe_workspace" "poc-child" {
  name         = "poc-child"
  organization = "Sandbox"
}

resource "tfe_variable" "aws_access_key_id" {
  key          = "AWS_ACCESS_KEY_ID"
  value        = ""
  sensitive    = false
  category     = "env"
  workspace_id = data.tfe_workspace.poc-child.id

  lifecycle {
    ignore_changes = [value]
  }
}

resource "tfe_variable" "aws_secret_access_key" {
  key          = "AWS_SECRET_ACCESS_KEY"
  value        = ""
  sensitive    = true
  category     = "env"
  workspace_id = data.tfe_workspace.poc-child.id
  //description  = "description change"

  lifecycle {
    ignore_changes = [value]
  }
}

resource "tfe_variable" "aws_region" {
  key          = "AWS_REGION"
  value        = "us-east-1"
  sensitive    = false
  category     = "env"
  workspace_id = data.tfe_workspace.poc-child.id
}
```

##### Child Workspace

```terraform
resource "aws_s3_bucket" "b" {
  bucket = "my-tf-test-bucket"

  tags = {
    Name        = "My bucket"
    Environment = "Dev"
  }
}
```

#### Debug Output

```plaintext
...
```

#### Expected Behavior

1. Create parent & child workspace
2. `Parent` - Apply workspace as coded above (should succeed)
a. Populate the values for `AWS_ACCESS_KEY_ID` & `AWS_SECRET_KEY` in the TFE UI
3. `Child` - Apply workspace as coded above (should succeed)
4. `Parent` - Remove the comment for `tfe_variable.aws_secret_access_key` to add the `description`. Run `terraform apply` & confirm (should succeed)
5. `Child` - Modify the `aws_s3_bucket.b` resource's `bucket` attribute with an extra character. Run `terraform plan`.

The plan run in step 5 **should** present the `bucket` change.

#### Actual Behavior

The `parent` run adding the `description` attribute on the `tfe_variable.aws_secret_access_key` resource only shows the description change, given the `ignore_changes` includes the `value` attribute. However, the `child` workspace fails due to a provider authentication error below.

The change to the `description` on the environment variable `AWS_SECRET_ACCESS_KEY` has inadvertently and silently wiped the value that was previously added manually.

```terraform 
Terraform v1.4.2
on linux_amd64
Configuring remote state backend...
Initializing Terraform configuration...

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Invalid provider configuration
│ 
│ Provider "registry.terraform.io/hashicorp/aws" requires explicit
│ configuration. Add a provider block to the root module and configure the
│ provider's required arguments as described in the provider documentation.
│ 
╵
╷
│ Error: configuring Terraform AWS Provider: no valid credential sources for Terraform AWS Provider found.
│ 
│ Please see https://registry.terraform.io/providers/hashicorp/aws
│ for more information about providing credentials.
│ 
│ AWS Error: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, failed to get API token, operation error ec2imds: getToken, http response error StatusCode: 400, request to EC2 IMDS failed
│ 
│ 
│   with provider["registry.terraform.io/hashicorp/aws"],
│   on <empty> line 0:
│   (source code not available)
│ 
╵
```

#### Additional Context

- The issue seems to only apply to `sensitive` variables, as non-sensitive variables seem unchanged in this scenario.
- Both `env` and `terraform` variable types are affected.
- I cannot replicate the issue when making the same `description` change via the TFE UI or API. Only through the `tfe` provider.

nfagerlund:
## Description

Fixes #839.

Wow, lots of sturm und drang for one little bug fix, huh? But anyway, this starts us on the path to adopting the shiny new plugin framework _(which, by the way, is actually incredibly excellent to work with compared to the v2 sdk)_ and validates our theory that it grants the power to resolve that perennially obnoxious "can't tell absent from zero" family of defects.

TODO: 

- [x] Minor docs tweak
- [x] Changelog

## Testing plan

Sorry, this is outrageously convoluted, which is why no one apparently noticed the bug until the reporter on #839.

1.  Make an arbitrary terraform config that relies on a sensitive variable being set to a particular value. (Probably easiest is to just switch between two output values based on an equality check against a known value.)
2. Make a terraform config that uses this provider to create a TFC workspace (which will use the config from step 1), and a sensitive variable in that workspace. The var should have the value set to an **empty string,** and should have `ignore_changes = [value]`. Like this:

    ```hcl
    resource "tfe_variable" "affected" {
      key          = "my_var"
      value        = ""
      category     = "terraform"
      workspace_id = tfe_workspace.var-sensitive-nuke-repro.id
      description  = "not edited yet"
      sensitive    = true
      lifecycle {
        ignore_changes = [value]
      }
    }
    ```
3. Apply the config from step 2.
4. **Manually, in the TFC UI,** change the value of the sensitive variable in your new workspace to the Particular Value that your config from 1 relies on.
5. Do an apply in the new workspace and verify the value is set as expected. 
6. Edit your config from step 2 to change **the `description`** of the variable, leaving `value` unchanged. Do an apply, and note that Terraform only reports a change to description as being necessary. 
7. Do an apply in the new workspace again.
    - In released versions of the provider, this now breaks because the value was **forcibly, silently reset** to an empty string, despite a description edit being the only planned change.
    - With this PR, it works fine.

**Additionally:** Slam on the `tfe_variable` resource as hard as you can and try to squeeze any kind of unexpected behavior out of it. It ought to behave. 

```
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEVariable_ -timeout 15m
--- PASS: TestAccTFEVariable_basic (37.70s)
--- PASS: TestAccTFEVariable_basic_variable_set (32.52s)
--- PASS: TestAccTFEVariable_update (55.84s)
--- PASS: TestAccTFEVariable_update_key_sensitive (55.95s)
--- PASS: TestAccTFEVariable_import (37.09s)
=== RUN   TestAccTFEVariable_rewrite
--- PASS: TestAccTFEVariable_rewrite (56.65s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/tfe	275.978s
```


nfagerlund:
Unless someone finds a notable defect in resource behavior, I think the main open questions for review are:

- How hard do we want to work for testing the state migrator? I actually DID get some good examples and suggestions for how to test those (mostly [this stuff in the `random` provider](https://github.com/hashicorp/terraform-provider-random/blob/main/internal/provider/resource_password_test.go#L1122)), so I'm willing to take a swing at it. On the one hand, probably this code path will never be executed in anger ever again, but on the other hand, half the point of this PR is to lay down a track in the dirt for the next person to follow, so maybe we want a fully tested migrator around just for that reason. 🤷 
- Can you spot anything in the resource implementation that needs better explanation or should be organized in a more elegant way?
- Do we want to change the physical layout of the files at this time, to prepare for growth in the number of framework resources and set them apart from the sdkv2 resources?

uturunku1:
I am excited to test these changes today! 
I am looking at the instructions for Testing Plan, and I wanted to ask for some clarifications. As far as I understand the testing plan is suggesting two work with two separate terraform configs: config from step 1 and config from step 2. 
Questions:
1. Step 3 is "Apply the config from step 2", so at what time do we "Apply the config from step 1"?
2. Step 2 says "create a TFC workspace which will use the config from step 1". Could you elaborate how is `resource "tfe_workspace" "var-sensitive-nuke-repro"` using the first config?

tfe/provider_next.go
```diff
@@ -0,0 +1,118 @@
+package tfe
```
brandonc, line 1:
```suggestion
// Copyright (c) HashiCorp, Inc.
// SPDX-License-Identifier: MPL-2.0

package tfe
```

brandonc (approved):
I couldn't break it.

tfe/provider_next.go
```diff
@@ -0,0 +1,118 @@
...
+	res.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"hostname": schema.StringAttribute{
+				Description: descriptions["hostname"],
+				Optional:    true,
+			},
+			"token": schema.StringAttribute{
+				Optional:    true,
+				Description: descriptions["token"],
+				// TODO: should be sensitive, but that's a breaking change.
```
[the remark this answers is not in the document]
brandonc: I wonder if this is just an awkward side effect of using the same schema types as resource schema 🤔

tfe/provider_next.go
nfagerlund: BAH THANK YOU. Got one file, missed the other.

tfe/provider_next.go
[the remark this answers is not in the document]
nfagerlund: I think so! I just noticed that the muxed provider explodes if every internal provider doesn't have the exact same schema, so we'd have to change them all. And doing so might not actually be a "breaking" change per se, I haven't really tried yet. 🤔

nfagerlund:
@uturunku1 

> As far as I understand the testing plan is suggesting to work with two separate terraform configs: config from step 1 and config from step 2. 

Yeah, that's right! 

>     1. Step 3 is "Apply the config from step 2", so at what time do we "Apply the config from step 1"?

In step 5, and in step 7. 

>     2. Step 2 says "create a TFC workspace which will use the config from step 1". Could you elaborate how is `resource "tfe_workspace" "var-sensitive-nuke-repro"` using the first config?

You'd have a `vcs_repo` block that grabs your config 1 from Github or whatever. Alternately, you could use CLI mode and just upload the config from a local workdir. 

Sorry, I know it's confusing! **Ultimately all we actually want to test is this:** If Terraform didn't say it was gonna change the value of the sensitive variable, the value had better not be changed after apply. But because the API never allows you to inspect the value again, you have to do SOMETHING tricky and annoying to work around that! You could skip the whole "config 1" roundabout entirely by spying on TFC internals in your local dev — either dumping the entire request to logs whenever a variable gets mutated, or cracking into the rails console and grabbing and decrypting the variable value out of the DB. But I wanted to make sure I was also writing down a validation that external testers could manage. 😓

uturunku1:
> @uturunku1
> 
> > As far as I understand the testing plan is suggesting to work with two separate terraform configs: config from step 1 and config from step 2.
> 
> Yeah, that's right!
> 
> > 1. Step 3 is "Apply the config from step 2", so at what time do we "Apply the config from step 1"?
> 
> In step 5, and in step 7.
> 
> > 2. Step 2 says "create a TFC workspace which will use the config from step 1". Could you elaborate how is `resource "tfe_workspace" "var-sensitive-nuke-repro"` using the first config?
> 
> You'd have a `vcs_repo` block that grabs your config 1 from Github or whatever. Alternately, you could use CLI mode and just upload the config from a local workdir.
> 
> Sorry, I know it's confusing! **Ultimately all we actually want to test is this:** If Terraform didn't say it was gonna change the value of the sensitive variable, the value had better not be changed after apply. But because the API never allows you to inspect the value again, you have to do SOMETHING tricky and annoying to work around that! You could skip the whole "config 1" roundabout entirely by spying on TFC internals in your local dev — either dumping the entire request to logs whenever a variable gets mutated, or cracking into the rails console and grabbing and decrypting the variable value out of the DB. But I wanted to make sure I was also writing down a validation that external testers could manage. 😓
> 
> In other words, Config 1 only exists to indirectly inspect the sensitive value before and after apply. You can use whatever you want to do that inspection!

Thanks for clarifying, that makes sense! I'll just my local dev to inspect the sensitive values. I'll try this on Monday morning and report back.

uturunku1 (commented):
This looks great. I appreciate the individual commits, it helped me digesting all the changes included in this PR. To be honest, I didn't read the Migration docs because I felt that between the Confluence page you wrote and your commits, I had plenty information.

As far as your questions posted on this comment https://github.com/hashicorp/terraform-provider-tfe/pull/873#issuecomment-1535573131 ...

1. I don't think that we need to test state migrator that hard because  we'll continue adding changes that will put it to test
2. No, I feel like there is a good amount of comments.
3.  Hmm, I am not sure about this one. It's a meaningful change. Part of me wants to say yes. Maybe this is something we can discuss in the team meeting and come out with a decision together?

tfe/resource_tfe_variable_migrate_test.go
```diff
@@ -1,45 +0,0 @@
-// Copyright (c) HashiCorp, Inc.
-// SPDX-License-Identifier: MPL-2.0
-
-package tfe
```
Uk1288, line 4: TO-DO: remove file name `resource_tfe_variable_migrate_test` from `.golangci.yml`

tfe/resource_tfe_variable.go
```diff
@@ -10,428 +10,709 @@ import (
...
+				Computed: true,
+				Default:  booldefault.StaticBool(false),
+				PlanModifiers: []planmodifier.Bool{
+					boolplanmodifier.RequiresReplaceIf(
+						func(ctx context.Context, req planmodifier.BoolRequest, resp *boolplanmodifier.RequiresReplaceIfFuncResponse) {
+							if req.StateValue.ValueBool() && !req.ConfigValue.ValueBool() {
+								resp.RequiresReplace = true
+							}
+						},
+						"Force replacement if sensitive argument changed from true to false.",
```
Uk1288, line 191: The forced replacement is only when `sensitive` changes from `true to false`, what about when there is a reverse change, `false -> true`?

tfe/resource_tfe_variable.go
```diff
@@ -10,428 +10,709 @@ import (
...
-
-	config := meta.(ConfiguredClient)
+// isWorkspaceVariable is a helper function for switching between tfe_variable's
+// two separate CRUD implementations.
+func isWorkspaceVariable(ctx context.Context, data AttrGettable) bool {
+	var variableSetID types.String
+	// We're ignoring the diagnostics returned by GetAttribute, because we'll
+	// be destructuring the entire schema value shortly in the real
+	// implementations; any notable problems will be reported at that point.
+	data.GetAttribute(ctx, path.Root("variable_set_id"), &variableSetID)
```
Uk1288, line 247: I think it might be clearer if `isWorkspaceVariable` checks that workspaceID is present as opposed to the check for the absence of variableSetID. Is there a reason this route was chosen? WDYT?

tfe/resource_tfe_variable.go
```diff
@@ -10,428 +10,709 @@ import (
...
+		Description: plan.Description.ValueStringPointer(),
+		HCL:         plan.HCL.ValueBoolPointer(),
+		Sensitive:   plan.Sensitive.ValueBoolPointer(),
+	}
+	// Specifically, we ONLY want to set Value if our planned value would be a
+	// CHANGE from the prior state. This is so we don't accidentally reset the
+	// value of a sensitive variable on unrelated changes when `ignore_changes =
+	// [value]` is set. (Basically: since we can't observe the real-world
+	// condition of a sensitive variable, we don't KNOW whether setting it to
+	// our last-known value is a safely idempotent operation or not. This is why
```
Uk1288, line 452:
```suggestion
	// our last-known value is a safe idempotent operation or not. This is why
```

tfe/resource_tfe_variable.go
nfagerlund: My understanding is: _that's_ fine, and the API will let you do it without a blip. It's just that the API won't let you turn sensitive _off,_ because it's not willing to reveal that prior value to you again once it's been contaminated with the sensitive flag.

tfe/resource_tfe_variable.go
nfagerlund: I think that was just a combination of my tendency to avoid a negated `if` when it's convenient, plus some influence from the logic in the old implementation. You're probably right, might as well switch it to a `if !workspace_id.IsNull()`.

tfe/resource_tfe_variable_migrate_test.go
nfagerlund: !! Good catch, thank you!

Uk1288 (approved):
LGTM

nfagerlund:
Recording for posterity: we discussed the prospect of reorganizing the provider at this time, and decided not to. The principle we're probably going to follow as we migrate resources is: if we nest it as a separate domain in our API docs, we should nest it in the provider's internal organization as well. That means basically

- admin API
- registry API

and nothing really else.

PR #880, opened 2023-05-11
Docs: Agents are now also available in Terraform Enterprise

nfagerlund:
## Description

So we don't need these "Cloud only" notes anymore.

Reference: https://developer.hashicorp.com/terraform/enterprise/admin/agents-on-tfe

## Testing plan

This is a docs-only PR.

### DIFF
--- a/website/docs/d/agent_pool.html.markdown
+++ b/website/docs/d/agent_pool.html.markdown
@@ -9,10 +9,6 @@ description: |-
 
 Use this data source to get information about an agent pool.
 
-~> **NOTE:** This data source requires using the provider with Terraform Cloud and a Terraform Cloud 
-for Business account. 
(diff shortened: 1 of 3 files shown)


website/docs/d/agent_pool.html.markdown
```diff
@@ -33,4 +29,4 @@ The following arguments are supported:
 
 In addition to all arguments above, the following attributes are exported:
 
-* `id` - The agent pool ID.
\ No newline at end of file
+* `id` - The agent pool ID.
```
annawinkler, line 32: 🤔 Is this a non-change change?

annawinkler (approved):
:shipit:

website/docs/d/agent_pool.html.markdown
nfagerlund: Yeah, that's my editor making sure the file ends with a newline. 😅 Ah well.

PR #881, opened 2023-05-12
Retry on workspace delete if still processing

### LINKED ISSUE #876: tfe_workspace resource should wait and retry safe delete on "still processing" errors

Cloud, prod.

#### Terraform version

```plaintext
1.4.4
```

(See examples in #786)

#### Expected Behavior

The new (in PR) `tfe_workspace_run` resource lets you manage a workspace's final destroy run! If you set `wait_for_run = true` and the run implicitly depends on the workspace itself, then Terraform will wait for the full destruction of the workspace's resources before deleting it. Nice! 

However, there's a race condition if the workspace isn't set to `force_delete = true`, because:

- The destroy run creates a new state version.
- TFC processes the state version asynchronously to count the resources in it.
- If the current state version doesn't have its resources processed yet, TFC responds to safe-delete requests with a 409, with the detail text `"Latest workspace state is being processed to discover resources, please try again later"`.

So, if you win the race and it tries to delete the workspace before async processing is done, your apply fails. 

#### Actual Behavior

A response of "just wait a sec" is actually very different from a response of "definitely not!" The workspace resource should distinguish between those two errors, and should sleep in a retry loop a couple of times when it gets the "still processing" response. 

Currently, the detail text is the only way to distinguish the two. We might also want to consider adding a more stable distinguishing characteristic to the jsonapi error objects on the TFC side.

#### Additional Context

We debated instead making the tfe_workspace_run resource's destroy runs wait for the workspace's resource count to drop to 0, as a more emphatic destroy run completion signal. But ultimately, the real correctness issue is in the workspace resource, so we should fix it there.

brandonc:
## Description

* `r/tfe_workspace`: Retry during workspace safe delete if resources are still being processed.

Fixes #876 

## External links

- [Related PR](https://github.com/hashicorp/go-tfe/pull/703)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ FEATURES:
 
 ENHANCEMENTS:
 
+* `r/tfe_workspace`: Retry workspace safe delete if resources are still being processed to determine safety. ([#881](https://github.com/hashicorp/terraform-provider-tfe/pull/881))
+
 BUG FIXES:
 
 * `r/tfe_variable`: Don't silently erase or override the `value` of a sensitive variable on changes to other attributes when `ignore_changes = [value]` is set, by @nfagerlund ([#873](https://github.com/hashicorp/terraform-provider-tfe/pull/873), fixing issue [#839](https://github.com/hashicorp/terraform-provider-tfe/issues/839))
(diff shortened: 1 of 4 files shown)


annawinkler (approved):
:shipit:

PR #897, opened 2023-05-23
Bump all GitHub Actions and pin; avoid deprecated Node 12

tbehling:
Commit is the result of running HashiCorp's internal pinning tool, for all GitHub Actions workflows in the repo.

https://github.blog/changelog/2023-05-04-github-actions-all-actions-will-run-on-node16-instead-of-node12/

Note, this unsolicited PR is a courtesy and has not been tested before submission.  The PR should be evaluated and merged by its responsible team.

### DIFF
--- a/.github/actions/lint-provider-tfe/action.yml
+++ b/.github/actions/lint-provider-tfe/action.yml
@@ -7,7 +7,7 @@ runs:
   using: composite
   steps:
     - name: Setup Go Environment
-      uses: actions/setup-go@v3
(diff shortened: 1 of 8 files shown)


annawinkler:
Thank you @tbehling ! We have this [PR](https://github.com/hashicorp/terraform-provider-tfe/pull/869) that was opened prior to yours - wondering which one we'd like to keep? I'd also like to see if we could add `terraform-cloud-action` so we can resolve this issue:

```
uses: hashicorp-forge/terraform-cloud-action/outputs@v1 # TSCCR: no entry for repository "hashicorp-forge/terraform-cloud-action"
```

What are your thoughts?

PR #904, opened 2023-05-26
Removed managed policy note

glennsarti:
## Description

Managed policies are not deprecated in TFC, therefore this commit removes this note.

## Testing plan

N/A


glennsarti:
I suspect it _was_ a direction at some point in time. But now with the new PnP packaging, non-versioned policies are a 1st class policy citizen

PR #909, opened 2023-05-26
Set admin role tokens from the env

annawinkler:
## Description

* Set the TFC Admin API role tokens from the environment
* Bump the version of `hashicorp-forge/terraform-cloud-action/` to the version that includes a fix to outputting nested sensitive values.

## Testing plan

The tokens should be set when CI is run (e.g., https://github.com/hashicorp/terraform-provider-tfe/actions/runs/5192171587/jobs/9360961243?pr=909#step:4:14)


annawinkler:
Verified that the tokens are in the output:

Screenshot 2023-06-06 at 1 20 24 PM

Issue #898, opened 2023-05-23
tfe_workspace_run resource's wait behavior is unreliable


nfagerlund:
#### Terraform Cloud/Enterprise version

main

#### Terraform version

latest, doesn't matter.

#### Terraform Configuration Files

From the `TestAccTFEWorkspaceRun_withApplyOnlyBlock` test:

```terraform
resource "tfe_workspace_run" "ws_run_parent" {
    workspace_id    = "<SOME_ID>"

    apply {
        manual_confirm = false
    }
}

resource "tfe_workspace_run" "ws_run_child" {
    workspace_id    = "<SOME ID>"
    depends_on   = [tfe_workspace_run.ws_run_parent]

    apply {
        manual_confirm = false
    }

    destroy {
        manual_confirm = false
    }
}
```

#### Test failure output

```plaintext
=== Failed
=== FAIL: tfe TestAccTFEWorkspaceRun_withApplyOnlyBlock (33.13s)
    resource_tfe_workspace_run_test.go:33: Step 1/1 error: Error running apply: exit status 1
        
        Error: run errored while applying run run-UbHRYDycYKSJT8tU (waited til status planned, currently status cost_estimating): transition not allowed
        
          with tfe_workspace_run.ws_run_parent,
          on terraform_plugin_test.tf line 3, in resource "tfe_workspace_run" "ws_run_parent":
           3: 	resource "tfe_workspace_run" "ws_run_parent" {
        

DONE 98 tests, 36 skipped, 1 failure in 328.060s
Error: Process completed with exit code 1.
```

#### Expected Behavior

The `TestAccTFEWorkspaceRun_withApplyOnlyBlock` test for the tfe_workspace_run resource should be reliably passing.

#### Actual Behavior

Instead, it fails randomly! 

After adding some extra logging, it's looking like this indicates a legit logic error in the resource, which we didn't catch during review and testing. It only happens rarely!

- In TFC's run state machine, certain states like `:planned` or `:post_plan_completed` can act as EITHER:
    - A semi-final resting state awaiting confirmation.
    - An incredibly brief transitory state en route to one of the other things that happens between a plan or an apply.
- In the resource, we're only treating those states as the former. 
- If we happen to poll during the exact instant that a run passes through the transitory form of a maybe-resting state, we try to confirm and apply when it's not allowed. 

So, we need to await a more reliable signal that the run is ready to rumble. Possibly `r.Actions.IsConfirmable`?

[...]

PR #921, opened 2023-06-06
Incorporate missing post plan terminal statuses to ensure plan is completely processed before attempting to confirm run

Uk1288:
## Description
The changes in this PR checks potential terminal post plan statuses to ensure that a plan has been completely processed before proceeding to confirm the associated run. The terminal status changes depending on the post plan lifecycle that a run has. For example a plan can terminate at `planned` status or `cost_estimated` or `policy_checked`.

See issue for more details: https://github.com/hashicorp/terraform-provider-tfe/issues/898

[This PR depends on go-tfe PR#712](https://github.com/hashicorp/go-tfe/pull/712)

## Testing plan

1.  _Workspace_run resource should create run as usual, the only way to produce the issue consistently is to force a run to remain in a status, e.g `policy_checked`, for a longer period of time such that workspace_run receives that status while polling the run. I achieved this by adding delays to each run state_

## External links

- [Blocked by go-tfe PR](https://github.com/hashicorp/go-tfe/pull/712)

```
$ TESTARGS="-run TestAccTFEWorkspaceRun" make testacc

...
```

### DIFF
--- a/tfe/resource_tfe_workspace_run.go
+++ b/tfe/resource_tfe_workspace_run.go
@@ -18,31 +18,27 @@ var (
 )
 
 var planPendingStatuses = map[tfe.RunStatus]bool{
-	tfe.RunPending:        true,
-	tfe.RunPlanQueued:     true,
-	tfe.RunPlanning:       true,
-	tfe.RunCostEstimating: true,
-	tfe.RunPolicyChecking: true,
-	tfe.RunQueuing:        true,
-	tfe.RunFetching:       true,
-}
-
-var planTerminalStatuses = map[tfe.RunStatus]bool{
-	tfe.RunPlanned:            true,
-	tfe.RunPlannedAndFinished: true,
-	tfe.RunErrored:            true,
-	tfe.RunCostEstimated:      true,
-	tfe.RunPolicyChecked:      true,
-	tfe.RunPolicySoftFailed:   true,
-	tfe.RunPolicyOverride:     true,
+	tfe.RunPending:           true,
+	tfe.RunPlanQueued:        true,
+	tfe.RunPlanning:          true,
+	tfe.RunCostEstimating:    true,
+	tfe.RunPolicyChecking:    true,
+	tfe.RunQueuing:           true,
+	tfe.RunFetching:          true,
+	tfe.RunPostPlanRunning:   true,
+	tfe.RunPostPlanCompleted: true,
+	tfe.RunPrePlanRunning:    true,
+	tfe.RunPrePlanCompleted:  true,
 }
 
 var applyPendingStatuses = map[tfe.RunStatus]bool{
-	tfe.RunConfirmed:   true,
-	tfe.RunApplyQueued: true,
-	tfe.RunApplying:    true,
-	tfe.RunQueuing:     true,
-	tfe.RunFetching:    true,
+	tfe.RunConfirmed:       true,
+	tfe.RunApplyQueued:     true,
+	tfe.RunApplying:        true,
+	tfe.RunQueuing:         true,
+	tfe.RunFetching:        true,
+	tfe.RunQueuingApply:    true,
+	tfe.RunPreApplyRunning: true,
 }
 
 var applyDoneStatuses = map[tfe.RunStatus]bool{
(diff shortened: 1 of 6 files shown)


tfe/resource_tfe_workspace_run.go
```diff
@@ -18,42 +18,61 @@ var (
...
+	tfe.RunPending:           true,
+	tfe.RunPlanQueued:        true,
+	tfe.RunPlanning:          true,
+	tfe.RunCostEstimating:    true,
+	tfe.RunPolicyChecking:    true,
+	tfe.RunQueuing:           true,
+	tfe.RunFetching:          true,
+	tfe.RunPostPlanRunning:   true,
+	tfe.RunPostPlanCompleted: true,
+	tfe.RunPrePlanRunning:    true,
```
brandonc, line 30: With the exception of `RunPostPlanCompleted`, I don't understand how some of these can be terminal statuses, when they appear to be transition states... Can you elaborate?

tfe/workspace_run_helpers.go
brandonc, line 1: In general, I find your diagnosis and fix to be convincing... but I'm also finding the helpers to be difficult to read and reason about and feel like we should take a moment to simplify them. I identified these specific things that could make this module simpler:

1. syntactically, I think there are two problems with awaitRun:
    a) nested switch statements-- I find it difficult to mentally retain a stack context once we reach that second nesting.
    b) this function takes a lot of arguments, which is an indicator that it could be doing too much or the input data is not related enough
2. mixed responsibilities: there are `interface{}` and provider-specific args in many helpers. Example: meta interface{} and schema.ResourceData args could be replaced with strong types like ConfiguredClient, retry/retry backoff integers, workspace ID, etc -- The idea would be that the helpers aren't interpreting provider schema or interfaces so that narrows the scope of their input.
3. the main entry point, createWorkspaceRun is looong so it may be best to try and break it into conceptual stages. I don't have a strong feeling about how to improve it, yet, but just moving code to subroutines that reflect the actual run stages might be a first step.

Does any of this resonate with you? What are your thoughts?

Uk1288: yes, this makes sense to me. I broke down the methods into smaller ones to represent action steps. The only one I have not updated is the double switch statement because the method is now smaller and straight forward. If we still want to get rid of that, I can replace it with an `if..else` statement. WDYT?

Uk1288:
- [ ] TO-DO: upgrade go-tfe after changes are good to go.

nfagerlund (changes requested):
This is getting there! I think you've split out the duties in a fairly legible way among these functions; some of the argument lists are still a bit unpleasant, but I think that's destiny, kind of, when the task is this annoying.

But I have some major concerns about mutating that top-scope `planPendingStatuses` map, and I don't think we should do it. I think what I'd suggest for a refactor goes something like this:

### The root issue:

- `isPlanComplete()` needs to inherently know about the list of plan terminal statuses.
- The call to `awaitRun()` needs to pass in the list of plan pending statuses.
- The actual lists of plan pending/terminal statuses depend on the run, _plus_ whether `hasPostPlanTaskStage` (which is inherent in the run but has to be discovered by making additional requests and then remembered for later, and we don't want those tightly-scoped helper funcs hitting the client.)

### What to do:

- Make a func that takes (run, hasPostPlanTaskStage) and returns a fresh planTerminalStatuses map.
- Make a func that takes (run, hasPostPlanTaskStage) and returns a fresh planPendingStatuses map.
- Make a func that takes planTerminalStatuses _and returns an anonymous function that can be used as an `isPlanComplete()` func._
- Then completely remove `hasPostPlanTaskStage` from the signatures of awaitRun and all the completion checker helper funcs, because `isPlanComplete` doesn't need it anymore -- it'll be a closure that already knows the things it currently needs `hasPostPlanTaskStage` in order to derive.

tfe/workspace_run_helpers.go
```diff
@@ -359,3 +407,94 @@ func backoff(min, max float64, iter int) time.Duration {
...
+		tfe.RunPlannedAndFinished: true,
+		tfe.RunPolicySoftFailed:   true,
+		tfe.RunPolicyOverride:     true,
+	}
+
+	if hasPolicyCheck {
+		planTerminalStatuses[tfe.RunPolicyChecked] = true
+
+		planPendingStatuses[tfe.RunCostEstimated] = true
+		planPendingStatuses[tfe.RunPlanned] = true
```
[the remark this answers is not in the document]
Uk1288: Right, mutating the global planPendingStatuses map raised questions for me as well, am glad that this comment validated my thought.

nfagerlund:
So these revisions look pretty great. Once the required extra status is merged into go-tfe and the provider can build properly, hit me up if I'm around and I'll try and do a smoke test; I think I still have a config sitting around from the last time we jammed on this.

Issue #968, opened 2023-07-18
panic: interface conversion: interface {} is nil, not string


liamstevens:
#### Terraform Cloud/Enterprise version
v0.46.0

Terraform Cloud

#### Terraform version

```plaintext
Terraform v1.4.6
on darwin_arm64
+ provider registry.terraform.io/hashicorp/tfe v0.46.0
```

#### Terraform Configuration Files

```terraform
resource "tfe_workspace" "workspaces" {
  for_each           = var.workspaces
  name               = each.value.name
  organization       = <SNIP>
  allow_destroy_plan = false
  auto_apply         = try(each.value.auto_apply, false)
  execution_mode     = "local"
  working_directory  = each.value.working_directory
  trigger_patterns   = each.value.trigger_patterns
  vcs_repo {
    github_app_installation_id = <SNIP>
    identifier                 = <SNIP>
    ingress_submodules         = try(each.value.include_submodules, false)
  }
  speculative_enabled = true
  global_remote_state = true
}
```

#### Debug Output

```plaintext
...
```

#### Expected Behavior

`terraform apply` completes as per the plan
#### Actual Behavior

`terraform apply` fails with a provider panic.

PR #969, opened 2023-07-18
Add validation in trigger patterns update

liamstevens:
## Description

This is a fix for #968 to add error handling during update operations on `trigger_pattern` attributes.

## Testing plan
Much like #519, the error and fix can be replicated as follows:

1. The minimal configuration is as below:

```terraform
provider "tfe" {}

  name = "panic-crash"
  organization = "star-destroyers"
  trigger_patterns = [""]
}
```
2. Run `terraform apply` - resources will be created successfully.
3. Update the `trigger_patterns` config and run another `terraform apply`. It will fail due to the Update method not performing error checking.
4. Pass your overrides:
```TF_CLI_CONFIG_FILE=/Users/jedimaster/path/to/tfrcfile terraform apply```
5. Success!


sebasslash (changes requested):
Thanks so much for contributing this fix 🔥 I see #519 inspired you a bit 😆  (I must have missed this panic, whoops). Just a small nitpick on the changelog wording. Once addressed, I will rebase these changes into a new PR in order for them to run against CI.

PR #980, opened 2023-07-25
Fix workspace interface conv panic

brandonc:
Run tests for #969


sebasslash (approved):
🚀. Looks like we are having issues with our CI instance, we'll re-run the workflow once fixed. As far as the changes are concerned, good to merge. Let's squash the commit history here too.

PR #1028, opened 2023-08-25
Migrate provider code to internal/ package

bflad:
## Description

Reference: https://developer.hashicorp.com/terraform/plugin/best-practices/interacting-with-providers

This change moves the provider code from the exported `tfe` package to the `internal/provider` package, which is a recommended practice since the compatibility and versioning of a provider is generally only considered for the interface with Terraform and practitioner configurations, not with other potential Go modules. The recommended way for other code to interact with a provider is via Terraform CLI's machine-readable interfaces, e.g. JSON outputs, and worst case gRPC.

The internalized code structure has been used across many official provider codebases already for a long time, including the utility providers, `hashicorp/aws`, and `hashicorp/azurerm`.

(Happy to adjust this change, such as keeping the provider code in `internal/tfe` package, if that is more desirable.)

_Remember to:_

- [ ] _Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)_
- [ ] _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)_

## Testing plan

1.  _Describe how to replicate_
1.  _the conditions_
1.  _under which your code performs its purpose,_
1.  _including example Terraform configs where necessary._

## External links

_Include any links here that might be helpful for people reviewing your PR. If there are none, feel free to delete this section._

- [go-tfe documentation](https://pkg.go.dev/github.com/hashicorp/go-tfe?tab=doc#xxxx)
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/xxxx)
- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/xxxx)

## Output from acceptance tests

_Please run applicable acceptance tests locally and include the output here. See [testing.md](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/testing.md) to learn how to run acceptance tests._

_If you are an external contributor, your contribution(s) will first be reviewed before running them against the project's CI pipeline._

```
$ TESTARGS="-run TestAccTFEWorkspace" make testacc

...
```

PR #1029, opened 2023-08-25
Migrate provider code to internal/ package

brandonc:
Non-fork branch of #1028


brandonc:
✅ I'm approving this PR as if it were the referenced fork but admin-merging it.

PR #1034, opened 2023-08-31
add client caching by config, extract client and logging packages

brandonc:
## Description

Adds tfe.Client caching and extracts a client and logging package from the provider package. It doesn't make sense for each provider (sdk v2 + framework) to initialize its own client because it has to go through ping and service discovery requests. This may also slightly speed up testing since each test begins by initializing a new client.

## Testing plan

Use `TF_LOG_PROVIDER=trace` to see that only one disco request is made when using the provider (terraform does another one). Previously, there were two (per provider alias!)


brandonc:
Fingers crossed but this seems to have smoothed out some test times

internal/client/config.go
```diff
@@ -0,0 +1,253 @@
...
+	Services   *disco.Disco
+	HTTPClient *http.Client
+	TFEHost    svchost.Hostname
+	Token      string
+	Insecure   bool
+}
+
+// Key returns a string that is comparable to other ClientConfiguration values
+func (c ClientConfiguration) Key() string {
+	return fmt.Sprintf("%x %s/%v", sha256.New().Sum([]byte(c.Token)), c.TFEHost, c.Insecure)
```
nfagerlund, line 55: Oh, ha, okay — I don't know if there's any way you could manage to configure an insecure and a secure client for the same hostname, but I can definitely see why we'd want them in separate slots if someone managed it.

nfagerlund (approved):
Okay, this is good — the layout of the code makes sense to me (i.e. having this in a new sibling package), and I gave it a try with the extra provider logging and saw exactly what you described — 0.48.0 does discovery three times (once for each muxed provider (sdkv2, framework, and low-level)), but this branch does it once.

internal/client/config.go
brandonc: You can add it to the provider config as `ssl_skip_verify` So you _could_ be aliasing the provider with an insecure for some odd reason.

PR #1116, opened 2023-10-25
Fix flaky TFESAMLSettings resource tests, add comments

nfagerlund:
## Description

These tests flake violently, because they're competing for control of global singleton state. The flakes show a distinctive pattern: they only flake when at least two of the resource test funcs end up on separate test runner partitions in CI, and re-running a single one failed partition alone always succeeds (because there's no concurrent competition anymore).

The fix is to make sure all of these tests ALWAYS run in serial on a single test runner partition. Smooshing into a single outer func and calling t.Run without t.Parallel seems like it should fit the bill.

changelog N/A, docs N/A (aside from code comments) 

## Testing plan

1.  It doesn't flake in CI. 

N/A (problem only occurs in parallelized CI)

### DIFF
--- a/internal/provider/data_source_saml_settings_test.go
+++ b/internal/provider/data_source_saml_settings_test.go
@@ -9,6 +9,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
+// FLAKE ALERT: SAML settings are a singleton resource shared by the entire TFE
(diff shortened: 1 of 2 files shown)


nfagerlund:
NOTE FOR REVIEWERS: This PR claims to be +213 / -194, but use the gear icon to turn on "hide whitespace" to see the actual changes (which are about 16 changed lines not including comments).

laurenolivia (approved):

PR #1119, opened 2023-10-27
[IPL4893] unable to connect workspace to agent pool

laurenolivia:
## Description

This PR adds a new resource `tfe_workspace_execution_mode`.

This PR also deprecates `agent_pool_id` and `execution_mode` in  the`tfe_workspace` resource.

Problem:
When a workspace and agent pool are created in the same `terraform apply`, a workspace cannot attach that agent pool's ID if the agent pool resource has prevented all workspace access via the `organization_scoped` [arg](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/agent_pool), upon creation.
See example below.

```
resource "tfe_agent_pool" "test-agent-pool" {
  organization_scoped = false
}
```

_Remember to:_

## Testing plan

1. This config requires executing 2 `terraform apply` commands to update state.

```
terraform {
  required_providers {
    tfe = {
      source = "hashicorp/tfe"
      version = "~> 0.48.0"
    }
  }
}

provider "tfe" {
  hostname = "your-hostname.io"
  token    = "xxxxxxxxx"
}

data "tfe_organization" "org" {
  name = "hashicorp"
}

resource "tfe_workspace" "workspace" {
  count             = 2
  name              = "pool-allowed-workspace-${count.index}"
  organization      = data.tfe_organization.org.name
}

resource "tfe_agent_pool" "pool" {
  name                = "test-6-IPL-4893"
  organization        = data.tfe_organization.org.name
  organization_scoped = false
}

# grants workspace permission to access agent pool
resource "tfe_agent_pool_allowed_workspaces" "permit" {
  agent_pool_id         = tfe_agent_pool.pool.id
  allowed_workspace_ids = [for key, value in tfe_workspace.workspace.*.id :value]
}

# attaches pool to workspace
resource "tfe_workspace_execution_mode" "attach" {
  workspace_id = tfe_workspace.workspace[0].id
  agent_pool_id = tfe_agent_pool_allowed_workspaces.permit.agent_pool_id
  execution_mode = "agent"
}
```

## External links

```

...
```
tfe_workspace_execution_mode:
Screenshot 2023-11-15 at 1 34 29 PM

tfe_workspace doc update:
Screenshot 2023-11-17 at 2 08 11 PM


internal/provider/resource_tfe_workspace_agent_pool_execution_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"testing"
+
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+// create and update
+// when workspace does not exist
+// when agent pool ID does not exist
```
Uk1288, line 18: to remove unused comment

internal/provider/resource_tfe_workspace_agent_pool_execution_test.go
```diff
@@ -0,0 +1,178 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccTFECheckWorkspaceAgentPoolNotExists("tfe_agent_pool.pool"),
+				),
+			},
+		},
+	})
+}
+
+func testAccTFECheckWorkspaceAgentPoolExists(resource string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
```
Uk1288, line 58: it seems like `testAccTFECheckWorkspaceAgentPoolExists` and `testAccTFECheckWorkspaceAgentPoolNotExists` are the same function, doing the same thing...

the check functions should be checking for the attachment/detachment as opposed to existence of agent pool.

internal/provider/resource_tfe_workspace_agent_pool_execution_test.go
```diff
@@ -0,0 +1,178 @@
...
+  name         = "agent-pool-updated"
+  organization = "%s"
+  organization_scoped = false
+}
+
+resource "tfe_agent_pool_allowed_workspaces" "permit"{
+  agent_pool_id 		= tfe_agent_pool.foobar.id
+  allowed_workspace_ids = [
+	tfe_workspace.foobar.id,
+	tfe_workspace.test-workspace.id
```
Uk1288, line 120: workspace `tfe_workspace.test-workspace` does not exists in the config

internal/provider/resource_tfe_workspace_agent_pool_execution.go
```diff
@@ -0,0 +1,135 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"agent_pool_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"workspace_id": {
+				Type:     schema.TypeSet,
```
Uk1288, line 33: should be,
```suggestion
				Type:     schema.TypeString,
```

internal/provider/resource_tfe_workspace_agent_pool_execution.go
```diff
@@ -0,0 +1,135 @@
...
+	var poolID string
+	if workspace.AgentPool != nil {
+		poolID = workspace.AgentPool.ID
+	}
+	d.Set("agent_pool_id", poolID)
+
+	return nil
+}
+
+func resourceTFEWorkspaceAgentPoolExecutionUpdate(d *schema.ResourceData, meta interface{}) error {
```
Uk1288, line 88: My thought is that since the 2 params, i.e `agent_pool_id` and `workspace_id` are both `ForceNew: true`, this update function might not be needed because the resources will always be recreated as opposed to updated in-place.

laurenolivia: That's what the error message also noted in the CI, but I left it to have a discussion first. Seems like most of our resources have `Update` methods so I wondered what others think? Remove the `ForceNew: true` field or remove the `Update` method?

internal/provider/resource_tfe_workspace_agent_pool_execution.go
```diff
@@ -0,0 +1,135 @@
...
+	return resourceTFEWorkspaceAgentPoolExecutionRead(d, meta)
+}
+
+func resourceTFEWorkspaceAgentPoolExecutionDelete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(ConfiguredClient)
+
+	poolID := d.Get("agent_pool_id").(string)
+	workspaceID := d.Get("workspace_id").(string)
+
+	log.Printf("[DEBUG] Delete the agent pool %s attached to workspace %s", poolID, workspaceID)
```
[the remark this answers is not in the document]
laurenolivia: Yes, this language is definitely clearer. Thanks!

internal/provider/resource_tfe_workspace_agent_pool_execution_test.go
laurenolivia: good eye

internal/provider/resource_tfe_workspace_agent_pool_execution_test.go
laurenolivia: oh I see, I actually pointed to `name` inside the workspace resource. Thanks.

internal/provider/resource_tfe_workspace_agent_pool_execution.go
laurenolivia: Or, rather, most of our resources follow CRUD.

brandonc (commented):
OK I took it out for a spin and came up with a list of little problems to solve together. I didn't get to the testing issues because I think we have some major schema changes to consider...
1. Unless you specify execution_mode = "agent", the initial resource computes the execution mode as "remote", which means that after a successful apply, there is an unexpected change planned on the next apply because it tries to reset the execution_mode and agent_pool_id back to the defaults
2. You can't specify execution_mode = "agent" on the workspace to try and counteract above--- it's an error because it requires agent_pool_id, which is deprecated. The combination feels a little hostile -- I want to specify agent execution so I created an agent_pool_execution resource but I certainly don't want "remote" or "local" execution mode.
3. I thought the warning on update was odd because warning logs aren't typically shown to users, but also because we are setting the execution mode to agent so I would expect the update response attrbute to always be "agent'

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 <!-- Add CHANGELOG entry to this section for any PR awaiting the next release -->
 <!-- Please also include if this is a Bug Fix, Enhancement, or Feature -->
 
+FEATURES:
+* **New Resource**: `r/tfe_workspace_agent_pool_execution` is a new resource to attach/detach an agent pool to a workspace, by @laurenolivia [1119](https://github.com/hashicorp/terraform-provider-tfe/pull/1119)
```
brandonc, line 6:
```suggestion
* **New Resource**: `r/tfe_workspace_execution_mode` is a new resource to modify the execution mode of a workspace, by @laurenolivia [1119](https://github.com/hashicorp/terraform-provider-tfe/pull/1119)
```

internal/provider/resource_tfe_workspace.go
```diff
@@ -80,9 +73,12 @@ func resourceTFEWorkspace() *schema.Resource {
 			},
 
 			"agent_pool_id": {
+				Deprecated: `Please do not use this argument to set an agent pool ID.
+				Instead, configure tfe_workspace_execution_mode to attach an agent pool to a workspace.
+				This attribute may be removed in the next major version of the provider.`,
```
brandonc, line 78: This is fine but it strikes me as a little pleading instead of prescriptive.

```suggestion
				Deprecated: `Use tfe_workspace_execution_mode to configure a workspace for agent
				pool execution. This attribute may be removed in the next major version of the provider.`,
```

internal/provider/resource_tfe_workspace.go
```diff
@@ -105,6 +101,10 @@ func resourceTFEWorkspace() *schema.Resource {
 			},
 
 			"execution_mode": {
+				Deprecated: `Please do not use this argument to set execution_mode to "agent".
+				Instead, configure tfe_workspace_execution_mode to set execution_mode = "agent",
+				if attaching an agent pool to a workspace.
+				This attribute may be removed in the next major version of the provider.`,
```
brandonc, line 107:
```suggestion
				Deprecated: `Use tfe_workspace_execution_mode to configure the workspace execution
				mode. This attribute may be removed in the next major version of the provider.`,
```

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,179 @@
...
+		Delete: resourceTFEWorkspaceExecutionModeDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"agent_pool_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
brandonc, line 30: I don't think this ForceNew option is necessary because if execution_mode is "agent", you should be able to update the agent_pool_id on a workspace.

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,179 @@
...
+	executionMode := d.Get("execution_mode").(string)
+	agentPoolID := d.Get("agent_pool_id").(string)
+
+	if executionMode == "agent" && agentPoolID == "" {
+		return fmt.Errorf(`agent_pool_id must be provided when execution_mode is "agent"`)
+	}
+
+	if executionMode != "agent" && agentPoolID != "" {
+		return fmt.Errorf(`execution_mode must be set to "agent" to assign agent_pool_id`)
+	}
```
brandonc, line 67: We should be able to move these errors into the plan phase by creating schema validation or, as a last resort, a `CustomizeDiff` handler. It would be a drag for users to have to wait until apply to find out they didn't configure the resource correctly.

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,179 @@
...
+						"remote",
+					},
+					false,
+				),
+			},
+		},
+	}
+}
+
+func resourceTFEWorkspaceExecutionModeCreate(d *schema.ResourceData, meta interface{}) error {
```
brandonc, line 55: I feel like either resourceTFEWorkspaceExecutionModeCreate should call resourceTFEWorkspaceExecutionModeUpdate. They both perform the exact same operation.

brandonc (changes requested):
Initial feedback. We also need to add documentation and an execution mode example to the deprecated fields on the website/docs/r/workspace docs

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 <!-- Add CHANGELOG entry to this section for any PR awaiting the next release -->
 <!-- Please also include if this is a Bug Fix, Enhancement, or Feature -->
 
+FEATURES:
+* **New Resource**: `r/tfe_workspace__execution_mode` is a new resource to manage the execution mode on a workspace, by @laurenolivia [1119](https://github.com/hashicorp/terraform-provider-tfe/pull/1119)
```
laurenolivia, line 6: what about `manage` instead of `modify` since it's required? @brandonc

internal/provider/resource_tfe_workspace.go
laurenolivia: Lol it's polite 🤷‍♀️

internal/provider/resource_tfe_workspace_execution_mode.go
laurenolivia: I wondered if actually the `Update` should use the d.HasChange() on `agent_pool_id` and if there's been no change on the pool ID then throw an error? Otherwise, stuff the new pool ID into the options struct? @brandonc

CHANGELOG.md
laurenolivia: It's also the language I use in our registry doc

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,179 @@
...
+	log.Printf("[DEBUG] Create attachment on workspace with agent pool ID: %s", agentPoolID)
+	workspace, err := config.Client.Workspaces.UpdateByID(ctx, workspaceID, options)
+	if err != nil {
+		return fmt.Errorf("error attaching agent pool ID %s to workspace ID %s: %w", agentPoolID, workspaceID, err)
+	}
+
+	d.SetId(workspace.ID)
+	// d.Set() will update state file on tfe_workspace for execution_mode and agent pool ID
+	d.Set("execution_mode", workspace.ExecutionMode)
+	d.Set("agent_pool_id", workspace.AgentPool.ID)
```
brandonc, line 147: It's possible for AgentPool to be null, causing a panic if the execution mode is not "agent"

internal/provider/resource_tfe_workspace_execution_mode.go
laurenolivia: `ValidateFunc` on execution_mode won't allow me to do a d.Get on agent_pool_id so I went with `CustomizeDiff`

internal/provider/resource_tfe_workspace_execution_mode.go
laurenolivia: I added validation to the customizediff, is that the best place for this check?

internal/provider/resource_tfe_workspace_execution_mode.go
brandonc: I think it's OK to send everything, just like on create. I think you'll want to conditionally send the agent pool ID, though, to only send it if you have one. Whatever makes the API and state file happy.

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return fmt.Errorf("error attaching agent pool ID %s to workspace ID on CREATE %s: %w", agentPoolID, workspaceID, err)
+	}
+
+	d.SetId(workspace.ID)
+	// d.Set() will update the resource state file for execution_mode and agent pool ID
+	d.Set("execution_mode", workspace.ExecutionMode)
+	d.Set("agent_pool_id", workspace.AgentPool.ID)
+
+	return resourceTFEWorkspaceExecutionModeRead(d, meta)
```
brandonc, line 81:
```suggestion
	return resourceTFEWorkspaceExecutionModeUpdate(d, meta)
```

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func resourceTFEWorkspaceExecutionModeRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(ConfiguredClient)
+
+	log.Printf("[DEBUG] Read configuration: %s", d.Id())
```
brandonc, line 87: I think we do debug logs on the actual request, but NBD

```suggestion
```

internal/provider/resource_tfe_workspace_execution_mode.go
brandonc: No, I would just set it to nil if the workspace.AgentPool is nil. There won't always be an AgentPool (like if the execution mode is "remote")

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,186 @@
...
+			log.Printf("[DEBUG] Workspace %s no longer exists", workspaceID)
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("error reading configuration of workspace %s: %w", workspaceID, err)
+	}
+
+	_, errs := config.Client.Workspaces.UpdateByID(ctx, workspaceID, tfe.WorkspaceUpdateOptions{
+		AgentPoolID:   tfe.String(""),
+		ExecutionMode: tfe.String("remote"),
```
brandonc, line 158: Now that there is an organization default exection mode, I wonder if we should set it to _that_

brandonc (changes requested):

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if d.HasChange("execution_mode") {
+		options.ExecutionMode = tfe.String(d.Get("execution_mode").(string))
+	}
+
+	if d.HasChange("agent_pool_id") {
+		options.AgentPoolID = tfe.String(d.Get("agent_pool_id").(string))
+	}
+
+	log.Printf("[DEBUG] Update attachment on workspace with agent pool ID: %s", d.Get("agent_pool_id"))
```
Uk1288, line 124: not all execution mode will be attaching agents, this should either be a generic message for all execution mode update or made conditional for agent_pool_id only

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,186 @@
...
+	workspace, err := config.Client.Workspaces.UpdateByID(ctx, workspaceID, options)
+	if err != nil {
+		return fmt.Errorf("error attaching agent pool ID %s to workspace ID %s: %w", d.Get("agent_pool_id"), workspaceID, err)
```
Uk1288, line 127: same here, not all execution mode will be attaching agents

internal/provider/resource_tfe_workspace_execution_mode.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+
+	if executionMode != "agent" && agentPoolID != "" {
+		return fmt.Errorf(`execution_mode must be set to "agent" to assign agent_pool_id`)
+	}
+
+	if executionMode != "agent" && d.GetRawConfig().GetAttr("agent_pool_id").IsNull() {
+		return fmt.Errorf(`agent_pool_id cannot be null, while execution_mode is not set to "agent"`)
+	}
```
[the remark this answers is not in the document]
laurenolivia: whoops, yep!

internal/provider/resource_tfe_workspace_execution_mode.go
laurenolivia: Not sure. That PR isn't merged yet and we will likely merge this one, first. @brandonc

brandonc:
We've decided to go in a different direction with the design

PR #1123, opened 2023-10-31
Add auto_apply_run_trigger attribute to workspace resource and data source

nfagerlund:
## Description

This adds provider support for the new auto-apply-run-trigger workspace attribute. This setting has shipped to the general public on TFC prod, but is not in TFE yet. 

_Remember to:_

## Testing plan

1. Set the attribute to true.
2. Set the attribute to false.
3. Leave the attribute blank when creating a workspace (defaults to hard-false, instead of allowing the API to decide for itself).

Something like this: 

```terraform
terraform {
  required_providers {
    tfe = {
      source = "hashicorp/tfe"
      version = "0.49.2" # EXCEPT, DO A BUILD AND SET YOUR DEV OVERRIDES CONFIG
    }
  }
}

provider "tfe" {
  hostname = "app.terraform.io"
  organization = "YOUR ORG"
}

data tfe_project "one-offs" {
    name = "One-offs" # YOUR PROJECT OF CHOICE FOR USELESS WORKSPACES
}

resource tfe_workspace "auto-trigger-test" {
    name = "auto-trigger-test"
    project_id = data.tfe_project.one-offs.id
    auto_apply = false
    auto_apply_run_trigger = true
    description = "Nov 2023 - Just testing provider support for this new attr."
}
```

## External links

- [go-tfe PR](https://github.com/hashicorp/go-tfe/pull/798)
- [Docs PR](https://github.com/hashicorp/terraform-docs-common/pull/377)

I'm internal, so see CI.


internal/provider/data_source_workspace_test.go
```diff
@@ -204,6 +205,13 @@ func TestAccTFEWorkspaceDataSource_readProjectIDNonDefault(t *testing.T) {
 }
 
 func testAccTFEWorkspaceDataSourceConfig(rInt int) string {
```
brandonc, line 207: I think its OK to specify `auto_apply_run_trigger` even if the API doesn't support it. Many enterprise installations will not.

internal/provider/resource_tfe_workspace.go
```diff
@@ -98,6 +98,12 @@ func resourceTFEWorkspace() *schema.Resource {
 				Default:  false,
 			},
 
+			"auto_apply_run_trigger": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
brandonc, line 104: Is the default needed for some behavior you are trying to elicit? I wonder if optional is enough

internal/provider/data_source_workspace_test.go
brandonc: I think it's cool that this feature was enabled in the test environment

brandonc (dismissed):
Everything was great. I'm just requesting changes until go-tfe is released and bumped

brandonc:
In order to test this, I had to update go-tfe to `fbcc54a62e4a6265d8395a477148d72e689ba2e9`

internal/provider/resource_tfe_workspace.go
nfagerlund: Good question! I was following the existing behavior of `auto_apply`, since it's basically the exact same setting.

Torn between my user brain saying "unmanaged is fine," my dev brain saying "Optional + Computed always causes weirdness on unset, love 2 avoid that when possible," and both brains saying "just be consistent with the other thing it's basically identical to."

internal/provider/resource_tfe_workspace_test.go
```diff
@@ -2293,6 +2296,20 @@ func testAccCheckTFEWorkspaceHasRemoteConsumers(ws string, wsConsumers []string)
...
 
+// Helper to to stub attribute checks for auto_apply_run_trigger until it's supported in TFE.
+func testAccCheckTFEWorkspaceAutoApplyRunTrigger(res, value string) resource.TestCheckFunc {
+	if enterpriseEnabled() {
+		return func(s *terraform.State) error {
+			return nil
+		}
+	}
+	return resource.TestCheckResourceAttr(res, "auto_apply_run_trigger", value)
+}
```
sebasslash, line 2307: This is a fantastic helper. I'm wondering if we could refactor this method as a generic helper for any attribute we should test TFC-only. This could avoid having to duplicate this logic again or worse, force the parent test to be flagged entirely. 👍

nfagerlund: 👀 that's an interesting thought. I'll be going back to this branch anyway to update go-tfe, may as well refactor this to a general helper while I'm in there.

internal/provider/testing.go
```diff
@@ -268,6 +270,20 @@ func skipIfUnitTest(t *testing.T) {
...
+//
+//nolint:unparam
+func testCheckResourceAttrUnlessEnterprise(name, key, value string) resource.TestCheckFunc {
+	if enterpriseEnabled() {
+		return func(s *terraform.State) error {
+			return nil
+		}
+	}
+	return resource.TestCheckResourceAttr(name, key, value)
+}
```
sebasslash, line 285: 🚀

sebasslash (approved):

PR #1125, opened 2023-11-01
Bump go-slug

brandonc:


nfagerlund:
Not needed anymore; this bump rode in on an unrelated go-tfe upgrade 👍🏼

PR #1127, opened 2023-11-02
[TF-10363] skip SAML resource tests when CI runs

laurenolivia:
## Description

Skip `tfe_saml_settings` resource tests in CI workflow.

_Remember to:_

## Testing plan

1. Invoke CI test suite
2. Verify that `TestAccTFESAMLSettings_omnibus` does not appear in logs
3. OR run `go test ./... -list` in terminal and list of tests that output should not include `TestAccTFESAMLSettings_omnibus`

## External links

```

...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,10 @@ jobs:
           admin_subscription_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.subscription }}
           admin_support_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.support }}
           admin_version_maintenance_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.version-maintenance }}
+          # Run terminal cmd 'go help testflag' to learn more about -list flag
+          # action.yml uses https://github.com/hashicorp-forge/go-test-split-action/blob/main/action.yml to split acceptance tests
+          # which runs against all tests using the list arg
+          # lists_tests regex is used to skip the TestAccTFESAMLSettings_omnibus test suite for CI tests only
           list_tests: "[^(TestAccTFESAMLSettings_omnibus)]"
 
   tests-combine-summaries:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
           admin_subscription_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.subscription }}
           admin_support_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.support }}
           admin_version_maintenance_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.version-maintenance }}
-          list_tests: "!^TestAccTFESAMLSettings_omnibus"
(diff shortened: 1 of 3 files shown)


sebasslash (changes requested):
I find this change to be interesting. Primarily because we normally skip tests only once the tests are already executing, i.e using `skipIfSomething(t *testing.T)` helpers. It allows us to provide detail as to why a particular test was skipped. Why are we omitting certain tests at the GH workflow level? 

If the intent here is to skip these tests temporarily (perhaps we are waiting on a fix), we should use [skipUnlessAfterDate()](https://github.com/hashicorp/go-tfe/blob/04571b5828ff1de4cc42257c9df54c41ec036738/helper_test.go#L2806-L2813) and set some arbitrary date in the future to fail tests (so we don't forget we've completely omitted this test).

If the intent is to never run this test, we should delete it. But I personally feel specifying it in the workflow file is bound to be forgotten.

laurenolivia:
> I find this change to be interesting. Primarily because we normally skip tests only once the tests are already executing, i.e using `skipIfSomething(t *testing.T)` helpers. It allows us to provide detail as to why a particular test was skipped. Why are we omitting certain tests at the GH workflow level?
> 
> If the intent here is to skip these tests temporarily (perhaps we are waiting on a fix), we should use [skipUnlessAfterDate()](https://github.com/hashicorp/go-tfe/blob/04571b5828ff1de4cc42257c9df54c41ec036738/helper_test.go#L2806-L2813) and set some arbitrary date in the future to fail tests (so we don't forget we've completely omitted this test).
> 
> If the intent is to never run this test, we should delete it. But I personally feel specifying it in the workflow file is bound to be forgotten.

cc @nfagerlund

laurenolivia:
@sebasslash Based on the ticket, I don't think the intention is to skip these tests temporarily. These tests will still run as part of our nightlies, just not in the CI.

.github/workflows/ci.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           admin_subscription_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.subscription }}
           admin_support_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.support }}
           admin_version_maintenance_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.version-maintenance }}
+          list_tests: "[^(TestAccTFESAMLSettings_omnibus)]"
```
brandonc, line 54: I thought this was a pretty clever way skip, given that the issue with the test is that it cannot run concurrently.

I also think this will be hard to understand in the future without explanation.

A self-explaining method might be to set an environment variable when running go test and then checking that variable inside the test itself. Something like "CI_PUSH" or "CI_NIGHTLY" or something that could be used to "skipUnlessNightly"

Of course, you could also just document this option here inside the workflow, although I find the test name itself to be a very brittle form of test selection.

brandonc (approved):

.github/workflows/ci.yml
sebasslash: `skipUnlessEnterprise()` is already our `skipUnlessNightly` so i don't understand what we're adding that's new here?

laurenolivia: I think using `skipIfEnterprise` is a little misleading because we are not skipping the tests because the feature is unavailable in Terraform Enterprise. Seems like we are solving for a different problem. Preventing these tests from running where we build our test setup seems clearer than having to maintain a custom solution, in my very humble opinion.

laurenolivia: RE: brittle name
I don't think we change tests names often but I can add a comment in the test file that notes we skip this test and should the test name ever change we need to update it in this file too?

PR #1142, opened 2023-11-16
TF-9704 Added organization_scoped resource to oauth client

Netra2104:
## Description

OrganizationScoped was added to the oauth client so that the user can scope it to an organization optionally.
By default the value of organization scoped is true when it is first created.

_Remember to:_

## Testing plan

Run the tests locally

## External links

- [Jira](https://hashicorp.atlassian.net/browse/TF-9704)
- [Related PR](https://github.com/hashicorp/go-tfe/pull/812)

```
=== RUN   TestAccTFEOAuthClient_basic
--- PASS: TestAccTFEOAuthClient_basic (9.48s)
PASS
```

## Documentation 
Screenshot 2023-11-17 at 11 15 12 AM

Screenshot 2023-11-17 at 11 15 21 AM


hs26gill (approved):
Please fix the lint issue else LGTM!

internal/provider/resource_tfe_oauth_client_test.go
```diff
@@ -153,6 +187,7 @@ resource "tfe_oauth_client" "foobar" {
   http_url         = "https://github.com"
   oauth_token      = "%s"
   service_provider = "github"
+  organization_scoped = true
```
Netra2104, line 190: @laurenolivia the CI tests that are failing right now are part of test files that I haven't modified. The reason they are failing is because the resource needs an organization_scoped attribute which is defined under beta flag in atlas. This attribute is optional. Locally, when I add line 190 to all the other failing tests they pass, but I would still have to skip all the tests which is a bad idea. Is there a way to get around this or do I have to remove the feature flag in atlas first?

laurenolivia: It sounds like the current options we've identified are either:
1. make the feature GA
2. use our `skipUnlessBeta()` helper

And neither are an option for your current workflow? Can you say more about why option #2 is a bad idea?

Netra2104: It not just this test, I would have to skip all the test that are failing by the CI tests right now (there are several). If the cli team is ok with this then I can go ahead and add it but...in the past the team has told me that skipping tests can be risky with other code changes that are added in.

jbonhag: Would it be possible to exercise the `organization_scoped` attribute only in its own set of tests?

brandonc (approved):

PR #1143, opened 2023-11-16
[r/tfe_workspace] Support `organization_default` value for the `execution_mode` argument

### LINKED ISSUE #1118: Support `execution_mode` of "Organization Default" for `tfe_workspace`
#### Use-cases

Allow a `tfe_workspace` to set `execution_mode` to "Organization Default". Currently the only options provided by this provider are `remote`, `local` or `agent`. TF Cloud recently added a new feature to set a default execution mode for all workspaces, and the workspace settings for execution mode has a new "Organization Default" option.

#### Attempted Solutions

Setting `execution_mode` to Organization Default is currently not possible as the provider only supports `remote`, `local` or `agent`.

#### Proposal

In addition to `remote`, `local` or `agent`, add `default`, which will respect the organization default.

To help you with implementation - via the API, I was able to achieve the "Organization Default" setting with the follow request data:
```
{"data":{"attributes":{"setting-overwrites":{"execution-mode":false,"agent-pool":false}}}}
```
So I imagine the provider can use a similar request body when implementing this feature.

Screenshot 2023-10-27 at 12 46 12 PM

tmatilai:
## Description

When the `execution_mode` is set to `organization_default`, the default execution mode of the organization is used.

The API is not very nice to be used with Terraform, and the UX with a separate `setting_overwrites` block feels equally clumsy. But YMMV, so feel free to judge.

Fixes #1118

_Remember to:_

## Testing plan

```terraform
  name = "test"

  execution_mode = "organization_default"
}
```

## External links

- [go-tfe documentation](https://pkg.go.dev/github.com/hashicorp/go-tfe#WorkspaceSettingOverwrites)
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces)

```

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 FEATURES:
 * `d/tfe_registry_module`: Add `vcs_repo.tags` and `vcs_repo.branch` attributes to allow configuration of `publishing_mechanism`. Add `test_config` to support running tests on `branch`-based registry modules, by @hashimoon [1096](https://github.com/hashicorp/terraform-provider-tfe/pull/1096)
+* `r/tfe_workspace`: Support `organization_default` value for the `execution_mode` argument, by @tmatilai [1143](https://github.com/hashicorp/terraform-provider-tfe/pull/1143)
 
 ## v0.50.0
 
--- a/internal/provider/resource_tfe_workspace.go
+++ b/internal/provider/resource_tfe_workspace.go
@@ -112,6 +112,7 @@ func resourceTFEWorkspace() *schema.Resource {
 				ValidateFunc: validation.StringInSlice(
 					[]string{
 						"agent",
+						"organization_default",
 						"local",
 						"remote",
 					},
@@ -313,6 +314,7 @@ func resourceTFEWorkspaceCreate(d *schema.ResourceData, meta interface{}) error 
 		SpeculativeEnabled:         tfe.Bool(d.Get("speculative_enabled").(bool)),
 		StructuredRunOutputEnabled: tfe.Bool(d.Get("structured_run_output_enabled").(bool)),
 		WorkingDirectory:           tfe.String(d.Get("working_directory").(string)),
+		SettingOverwrites:          &tfe.WorkspaceSettingOverwritesOptions{},
 	}
 
 	// Send global_remote_state if it's set; otherwise, let it be computed.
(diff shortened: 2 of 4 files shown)


SwiftEngineer:
I have a few concerns with the PR 👋 

1. Adding `organization_default` as a possible option makes adding future defaults tricky. If we ever want the ability to set the default execution mode of projects for example.
2. This PR might break in a few different cases, like, if the organization default execution mode is agent for example, we have validations in place that check the agent pool id. If it is set (which it would be during the refresh step if the organization default execution mode is agent), then I believe this PR might trigger a failure. Perhaps you've already tested this? Hopefully I'm wrong 🙏 
3. This change means users are no longer able to read the value of the `execution_mode` in order to conditionally modify their infrastructure. For example, if a customer wants to dynamically provision agents based on the number of workspaces that are using the agent execution mode. For this reason, I'd greatly prefer if the `execution_mode` property always matched the `execution_mode` that the workspace is actually set to.

tmatilai:
> 1. Adding `organization_default` as a possible option makes adding future defaults tricky. If we ever want the ability to set the default execution mode of projects for example.

Agreed. But I didn't like just the `default` as it's not very clear. Suggestions?

> 2\. [...] if the organization default execution mode is agent for example, we have validations in place that check the agent pool id. If it is set (which it would be during the refresh step if the organization default execution mode is agent), then I believe this PR might trigger a failure.

The validation of the resource doesn't let the agent pool id to be unless `execution_mode` is agent. If that's what you meant.

> 3\. This change means users are no longer able to read the value of the `execution_mode` in order to conditionally modify their infrastructure. For example, if a customer wants to dynamically provision agents based on the number of workspaces that are using the agent execution mode.

I don't see how it would be better to need to parse the `setting_overwrites` either? In both cases they would anyway need to check the organization (or project in the future) default, too.

Anyway, this is just one suggestion. Maintainers are free to do what they want.  🙂 

But I must say that I don't like the API too much about this. Makes things complicated.

SwiftEngineer:
> The validation of the resource doesn't let the agent pool id to be set unless execution_mode is agent. If that's what you meant.

Ah, I think I was confused because I see now that your `agent_pool_id` is not computed! My bad 👍 

> I don't see how it would be better to need to parse the setting_overwrites either? In both cases they would anyway need to check the organization (or project in the future) default, too.

Users shouldn't need to parse the `setting_overwrites` though, correct? It's just the provider itself that needs to do this. 

> But I must say that I don't like the API too much about this. Makes things complicated.

Tell me about it! Nobody is a fan of the whole `setting_overwrites` thing, it's been tricky, but it did allow us to avoid breaking backwards compatibility with a LOT of systems that rely on TFC and TFE's APIs. The biggest silver-lining I've found in it is that the majority of users will never need to know anything about it, so at least the complexity CAN be avoided.

tmatilai:
All good. And thanks for taking the time to explain the decisions. 👍

PR #1144, opened 2023-11-16
TF-9700 Add a new project_oauth_client resource

Netra2104:
## Description

New endpoints in atlas have been added that add a project relationship to an oauth client. Supporting functions have been added to go-tfe as well.

**Note**: this is a beta feature as this PR might be merged before the TFC launch. There will be a PR in the future to remove the beta flags.

_Remember to:_

- [x] _Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)_
- [x] _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)_

## Testing plan

Run the unit tests locally.

## External links

_Include any links here that might be helpful for people reviewing your PR. If there are none, feel free to delete this section._

- [Related PR](https://hashicorp.atlassian.net/browse/TF-9700)

## Output from acceptance tests

_Please run applicable acceptance tests locally and include the output here. See [testing.md](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/testing.md) to learn how to run acceptance tests._

_If you are an external contributor, your contribution(s) will first be reviewed before running them against the project's CI pipeline._

```
--- PASS: TestAccTFEProjectOAuthClient_basic (10.16s)
PASS

--- PASS: TestAccTFEProjectOAuthClient_incorrectImportSyntax (8.24s)
PASS
```

## Documentation Output
Screenshot 2023-11-17 at 9 52 30 AM
Screenshot 2023-11-17 at 9 52 38 AM

PR #1148, opened 2023-11-20
TF-9703 project_ids and organization_scoped to oauth client

Netra2104:
## Description

Add project_ids and organization scoped to oauth client data source.

Needs to be merged after: https://github.com/hashicorp/terraform-provider-tfe/pull/1144 

_Remember to:_

## Testing plan

Run the tests without skip beta flag locally.
CI test passes

## External links

- [Jira](https://hashicorp.atlassian.net/browse/TF-9703)

```

...
```

## Documentation Output
Screenshot 2023-11-20 at 11 35 21 AM


internal/provider/data_source_oauth_client.go
```diff
@@ -132,6 +141,7 @@ func dataSourceTFEOAuthClientRead(d *schema.ResourceData, meta interface{}) erro
 	}
 	d.Set("service_provider", oc.ServiceProvider)
 	d.Set("service_provider_display_name", oc.ServiceProviderName)
+	d.Set("organization_scoped", oc.OrganizationScoped)
```
brandonc, line 144: Will this behave correctly on TFE where this key is missing/false (This would be the default empty value when the key) ?

[The resource](https://github.com/hashicorp/terraform-provider-tfe/blob/main/internal/provider/resource_tfe_oauth_client.go#L119) has the benefit of having a default true value.

But it would seem like you need to invert it if the field is fully missing. Unfortunately, your go-tfe field is not a pointer so it's not easy to know whether the field was missing or literally false.

What are your thoughts?

brandonc (commented):
❔ 👇

internal/provider/data_source_oauth_client.go
netramali: In atlas, organization_scoped boolean DEFAULT true NOT NULL is set as a restriction. So then the missing case shouldn't be possible correct.

brandonc: I was talking about when you are using this against TFE that haven't added this API yet and therefore organization-scoped would be missing from the response altogether

brandonc: I see two options:

1. Release a known 'incorrect' value for TFE customers until they upgrade to latest release and get the API
2. Change go-tfe beta field (this is why they are beta!) to be a pointer that can be nil. The nil value will indicate the API is missing.
3. Change the API altogether to be "project-scoped", which makes missing/default/false values all alike. (This is obviously a crazy choice at this stage)

netramali: Which one would you suggest? I would have to confirm with my PM if option 1 is ok to do.

brandonc: I think the best compromise is to change the field to be a pointer attribute in the underlying go-tfe interface. It's hard to justify releasing a known bug when it can be fixed relatively easily

netramali: I think I am going to go ahead and implement #2, that one seems like the best solution in this case.

hs26gill (approved):
LGTM!

PR #1159, opened 2023-12-01
Introduce tfe_workspace_settings (deprecates execution_mode, agent_pool_id)

brandonc:
## Description

Recently, we merged #1137, which added `tfe_organization_default_execution` mode resource as well as `setting_overwrites` field on the `tfe_workspace` resource.

This added quite some complexity to `tfe_workspace` because of the previous deprecation of `operations`, the size of the resource in general, and the special handling required for setting_overwrites in the plugin SDK v2. It also brushed up against `agent_pool_id`, where we were already having challenges with a circular dependency between tfe_workspace and tfe_agent_pool / _allowed_workspaces.

So this change does three things to simplify the resource API:

1. extracted execution_mode, agent_pool_id, and setting_overwrites from tfe_workspace into a new resource written for the plugin framework, tfe_workspace_settings. This simplified the plan modifiers required by the computed fields, but also created a nice symmetry with organization defaults since these fields in particular are subject to being defined by the organization. More of these kinds of overwrite settings are expected in the future, so I also...
2. renamed tfe_organization_default_execution_mode to tfe_organization_default_settings.
3. deprecated agent_pool_id and execution_mode on tfe_workspace

## IMPORTANT

Please review commit-wise!

## Testing plan

1. Use the two new resources to default and overwrite workspace execution
2. ...Especially with an agent pool and agent pool allowed workspaces

```
$ TESTARGS="-run TestAccTFEWorkspaceSettings" make testacc
=== RUN   TestAccTFEWorkspaceSettings
2023/12/01 10:18:17 [DEBUG] Configuring client for host "tfcdev-5839ba17.ngrok.io"
2023/12/01 10:18:17 [WARN] Unable to read CLI config or credentials file /Users/brandonc/.terraformrc: open /Users/brandonc/.terraformrc: no such file or directory
2023/12/01 10:18:17 [DEBUG] Service discovery for tfcdev-5839ba17.ngrok.io at https://tfcdev-5839ba17.ngrok.io/.well-known/terraform.json
--- PASS: TestAccTFEWorkspaceSettings (21.75s)
=== RUN   TestAccTFEWorkspaceSettingsImport
--- PASS: TestAccTFEWorkspaceSettingsImport (13.23s)
=== RUN   TestAccTFEWorkspaceSettingsImport_ByName
--- PASS: TestAccTFEWorkspaceSettingsImport_ByName (14.17s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/provider	49.589s
...
```


internal/provider/resource_tfe_workspace_settings.go
```diff
@@ -0,0 +1,410 @@
...
+			log.Printf("[DEBUG] Workspace %s no longer exists", workspaceID)
+			return nil, errWorkspaceNoLongerExists
+		}
+		return nil, fmt.Errorf("couldn't read workspace %s: %s", workspaceID, err.Error())
+	}
+
+	return workspaceSettingsModelFromTFEWorkspace(ws), nil
+}
+
+func (r *workspaceSettings) updateSettings(ctx context.Context, data *modelWorkspaceSettings, state *tfsdk.State) error {
```
SwiftEngineer, line 302: Shouldn't we add handling here for outdated versions of TFE that do not support setting overwrites? I believe in this case, the provider would previously use `remote` as the default execution_mode, whereas I believe this code would simply NOT set the execution_mode.

A part of me really wants to force users to set the `execution_mode`... purely based on how that would drastically reduce the complexity of the provider from BOTH a user standpoint and a developer standpoint. That being said, it feels wrong to oppose the API and (effectively) disallow users the ability to use default execution modes if they use the terraform provider.

SwiftEngineer (changes requested):
I found some bugs during testing!

That being said, the biggest of the bugs is the one that happens when unsetting the deprecated `execution_mode`, so maybe it's something you already knew about and decided it is acceptable? The other only happens with outdated versions of TFE (that don't support setting-overwrites).

If you already knew about these issues and deemed them acceptable, please let me know and I'll re-review with that in mind 🙇

internal/provider/resource_tfe_workspace.go
```diff
@@ -41,16 +41,6 @@ func resourceTFEWorkspace() *schema.Resource {
...
-			// NOTE: execution mode and agent_pool_id must be set to default first before calling
-			// the validation functions
-			if err := setComputedDefaults(c, d); err != nil {
-				return err
-			}
-
-			if err := overwriteDefaultExecutionMode(c, d); err != nil {
-				return err
-			}
-
```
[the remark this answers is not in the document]
brandonc: It does seem to be a tradeoff when extracting this argument into a separate resource.  Removing the computed default does break the unsetting case out of necessity because otherwise it would continuously conflict when you unset it to add `tfe_workspace_settings`.

I highlighted this as a breaking change in the release notes, but it deserves more attention, so I added a note in the documentation.

CHANGELOG.md
```diff
@@ -2,16 +2,17 @@
...
-* **New Resource**: `r/tfe_organization_default_execution_mode` is a new resource to set the `default_execution_mode` and `default_agent_pool_id` for an organization, by @SwiftEngineer [1137](https://github.com/hashicorp/terraform-provider-tfe/pull/1137)'
-* `r/tfe_workspace`: Now uses the organization's `default_execution_mode` and `default_agent_pool_id` as the default `execution_mode`, by @SwiftEngineer [1137](https://github.com/hashicorp/terraform-provider-tfe/pull/1137)'
+* **New Resource**: `r/tfe_organization_default_settings` is a new resource to set the `default_execution_mode` and `default_agent_pool_id` for an organization, by @SwiftEngineer [1137](https://github.com/hashicorp/terraform-provider-tfe/pull/1137)'
+* **New Resource**: `r/tfe_workspace_settings` Uses the `tfe_organization_default_settings` `default_execution_mode` and `default_agent_pool_id` as the default `execution_mode` by @brandonc [1159](https://github.com/hashicorp/terraform-provider-tfe/pull/1159)
```
sebasslash, line 15: Can we also credit @laurenolivia for the initial work done migrating execution mode setting in #1119

sebasslash (changes requested):

internal/provider/resource_tfe_workspace_settings.go
SwiftEngineer: Awesome 💯

SwiftEngineer (approved):

website/docs/r/workspace_settings.markdown
```diff
@@ -0,0 +1,113 @@
+---
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_workspace_setting"
+description: |-
+  Manages workspace settings.
+---
+
+# tfe_workspace_settings
+
+Manages or reads execution mode and agent pool settings for a workspace. If [tfe_organization_default_settings](organization_default_settings.html) are used, those settings may be read using a combination of the read-only `overwrites` argument and the setting itself.
```
sebasslash, line 10: 💅 We should probably add an example for using `overwrites`.

sebasslash (approved):
LGTM

website/docs/r/workspace_settings.markdown
brandonc: Since it's a read-only attribute, all the examples I can think of feel a little contrived. I don't think you'd be interested in this unless you were also reading the execution_mode but needed to know whether or not a default was imposed. I'll leave it out for now but if a clear example emerges I will add it.

website/docs/d/workspace.html.markdown
```diff
@@ -60,7 +60,7 @@ In addition to all arguments above, the following attributes are exported:
 * `trigger_patterns` - List of [glob patterns](https://developer.hashicorp.com/terraform/cloud-docs/workspaces/settings/vcs#glob-patterns-for-automatic-run-triggering) that describe the files Terraform Cloud monitors for changes. Trigger patterns are always appended to the root directory of the repository.
 * `vcs_repo` - Settings for the workspace's VCS repository.
 * `working_directory` - A relative path that Terraform will execute within.
-* `execution_mode` - Indicates the [execution mode](https://developer.hashicorp.com/terraform/cloud-docs/workspaces/settings#execution-mode) of the workspace. Possible values include `remote`, `local`, or `agent`.
+* `execution_mode` - **Deprecated** Indicates the [execution mode](https://developer.hashicorp.com/terraform/cloud-docs/workspaces/settings#execution-mode) of the workspace. Use the `tfe_workspace_settings` resource instead.
```
nfagerlund, line 63: Hmm, this probably isn't called for in the data-source, since that's read-only... right?

brandonc: I did this because it's not the full picture any longer because `execution_mode` can be set from the organization default or from the workspace (as an override) `tfe_workspace_settings` gives access to both pieces of information and contains only optional attributes.

website/docs/r/organization_default_settings.html.markdown
```diff
@@ -20,14 +20,20 @@ resource "tfe_organization" "test" {
...
+  organization           = tfe_organization.test.name
   default_execution_mode = "agent"
-  default_agent_pool_id = tfe_agent_pool.my_agents.id
+  default_agent_pool_id  = tfe_agent_pool.my_agents.id
+}
+
+resource "tfe_workspace" "my_workspace" {
+  name       = "my-workspace"
+  # Ensures this workspace will inherit the org defaults
+  depends_on = [tfe_organization_default_settings.org_default]
```
nfagerlund, line 36: (Recording this for posterity, no one needs to respond to this.) I was confused about why this was necessary, because the workspace will just inherit the default anyway if you don't specify an overridden value. But the issue is that the workspace reports the inherited default as the value of its attribute of the same name, so without the ordering, you get needless churn — the workspace first reports it inherited the old default, then on the second run it reports that it inherited the new default as a change to the computed value. So the ordering _is_ important, because it'll give you a stable result on the first apply. 👍🏼

website/docs/r/workspace.html.markdown
```diff
@@ -9,6 +9,8 @@ description: |-
 
 Provides a workspace resource.
 
+~> **NOTE:** Setting the execution mode and agent pool affinity directly on the workspace has been deprecated in favor of using both [tfe_workspace_settings](workspace_settings) and [tfe_organization_default_settings](organization_default_settings). Use caution when unsetting `execution_mode` because the default value `"remote"` is no longer applied when the `execution_mode` is unset.
```
nfagerlund, line 12: As a user reading this, it's not clear to me what _will_ happen instead when I unset `execution_mode`. 🤔 Tho, as someone who's been in here reviewing recently, I think the answer is "it'll revert to the configured org default (whatever that happens to be) rather than a hardcoded default chosen by this provider".

nfagerlund: Oh wait, but actually, after reading the code: it won't revert to _anything,_ it'll just leave it what it was previously managed to (until something else changes it). And that's because, we _have_ to do it that way in order to continue supporting the deprecated attributes properly while allowing the new settings resource to take precedence when the deprecated attrs are absent. GOT IT.

website/docs/r/workspace_settings.markdown
nfagerlund: Hmm. The description in this paragraph is a little disorienting, mostly just because the whole overwrites scheme is a bit challenging to get a grip on. I might whip up a proposed reword.

website/docs/d/workspace.html.markdown
nfagerlund: Ah ok, thx for that context — I think I still disagree slightly, but I'll make a case in a docs PR.

PR #1163, opened 2023-12-04
Fix 'no changes detected' when provider default organization changes

### LINKED ISSUE #1152: `organization` change in the provider configuration is not detected in the resouces
#### Terraform version

```plaintext
Terraform v1.6.3
TFE provider v0.50.0
```

First create a workspace to an organization configured in the provider:

```terraform
provider "tfe" {
  organization = "example-1"
}

resource "tfe_workspace" "example" {
  name = "foo"
}
```

Then change the organization in the provider configuration:
```terraform
provider "tfe" {
  organization = "example-2"
}

resource "tfe_workspace" "example" {
  name = "foo"
}
```

#### Terraform Output

```plaintext
$ terraform apply 
tfe_workspace.example: Refreshing state... [id=ws-123456789]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

#### Expected Behavior

The resource should have been recreated to the other organization.

#### Actual Behavior

The organization change was not detected.

#### Additional Context

Probably not very essential in normal operations. But I hit the [same issue](https://github.com/hashicorp/terraform-plugin-framework/issues/878) when started to implement some new resources and data sources using the plugin framework, and realized that it already exists with the SDKv2 resources.

I guess also with the SDK the `ConfiguredClient.schemaOrDefaultOrganization()` uses the state instead of configuration for verifying if the organization has been set in the resource.

brandonc:
## Description

Closes #1152 

_Remember to:_

## Testing plan

1. Create a workspace or any other resource that has an "organization" argument using a default org from the provider
2. Change the organization string in the provider config
3. Resources will now will be replaced. Previously, no change would be detected.

I couldn't think of a way to automate the testing of this customized diff. Ideas?

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,7 +24,6 @@ linters:
     - gosec #https://github.com/securego/gosec
     - predeclared #https://github.com/nishanths/predeclared
     - unconvert #https://github.com/mdempsky/unconvert
-    - wrapcheck #https://github.com/tomarrell/wrapcheck
 
 issues:
   exclude-rules:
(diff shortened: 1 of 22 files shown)


nfagerlund (approved):
I ran it and it works! 🙌🏼

I don't believe there's a sensible way to test this via the standard provider acceptance test harness? I think the state gets wholly owned by the respective `resource.TestCase`, and I don't think it offers you any hook where you could mutate the providers in between the `resource.TestStep`s. 🤔 

Anyway, given that complexity, I'm fine shipping this without tests to fix the bug, rather than blocking on re-implementing the test harness. 

OH WAIT

ACTUALLY

It might be possible to abuse the import state test steps for this? It would take an experiment to be sure, tho, because all the examples I'm seeing are going like,

1. Apply a config
2. Import into fresh state the resource the config created. 
3. Compare to prior state and ensure no changes.

But, maybe if you were to 

1. Use the default org test provider.
1. Make a workspace outside config with the raw client, with an org that doesn't match the default org.
2. Set `ImportStateVerify` false and `ImportStatePersist` true and use that as the first step.
3. Apply the config for the first time and ensure there's a planned change to org.

??? 

Well, anyway, I still won't block the PR on going down that entirely speculative rabbit hole.

brandonc:
@nfagerlund Thanks for the ideas! I was in the middle of implementing it when I found a way to use PreCheck() to mutate the provider config between steps. Using the same config twice, I just ensure the organization attribute changed. https://github.com/hashicorp/terraform-provider-tfe/pull/1163/commits/6e976234a30ffd3c7bd3500b80e2d9a95b181ccc

PR #1171, opened 2023-12-14
Docs tweaks for tfe_workspace_settings and friends

nfagerlund:
## Description

Updating several things I found unclear as a user, plus one thing I found unclear as a fellow maintainer.

Also, an important adjustment to some messaging: `r/tfe_workspace_settings` _cannot_ safely be used as a data source, because it interprets absence of each optional argument to explicitly set the workspace's corresponding overwrite to false. This makes me believe we need an overwrites object attribute added to `d/workspace`, but this PR defers that for later.

This is a follow-on from some post-merge comments I left in #1159. 

- [x] _Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)_ N/A
- [x] _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)_ N/A

## Testing plan

1.  n/a

## External links

- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/1159)

### DIFF
--- a/internal/provider/resource_tfe_workspace.go
+++ b/internal/provider/resource_tfe_workspace.go
@@ -598,6 +598,12 @@ func resourceTFEWorkspaceUpdate(d *schema.ResourceData, meta interface{}) error 
 			}
 		}
 
+		// NOTE: since agent_pool_id and execution_mode are both deprecated on
+		// tfe_workspace and we want tfe_workspace_settings to be authoritative,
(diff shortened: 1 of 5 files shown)


brandonc (approved):
Very great

PR #1191, opened 2023-12-21
Improve tfe_agent_pool_allowed_workspaces docs

brandonc:
## Description

This example was inadequate because it didn't reference workspace_settings which can now break the dependency cycle between these resource. I've also highlighted how to ensure that execution mode is assigned after (and destroyed before) allowed workspace permissions on the agent pool

_Remember to:_

- [ ] ~_Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)_~


sebasslash (approved):
Great improvement. Much clearer dependency chain.

PR #1197, opened 2024-01-02
Update workspace_settings.markdown

brandonc:
Quick correction to the tfe_workspace_settings resource docs to match the recommendation from the tfe_agent_pool docs.


laurenolivia (approved):
👍

Issue #1203, opened 2024-01-04
Wrong error message when agent pool id is set and when execution is remote


Erouan50:
Hi!

I noticed a small bug when using the resource "tfe_workspace" with the field `agent_pool_id` set and when the execution mode is remote. Let's assume the following workspace:

```hcl
resource "tfe_workspace" "this" {
  name         = "my-workspace"
  organization = "My org"

  execution_mode = "remote"
  agent_pool_id = "my-agent"
}
```
When I do a plan I get the following error:

```plaintext
╷
│ Error: agent_pool_id must be provided when execution_mode is 'agent'
│ 
│   with tfe_workspace.this,
│   on main.tf line 1, in resource "tfe_workspace" "this":
│    1: resource "tfe_workspace" "this" {
│ 
╵
```

I think the error message should be `execution_mode must be set to 'agent' to assign agent_pool_id` instead of `agent_pool_id must be provided when execution_mode is 'agent'`

I'll open a pull request where I'll swap the error message in that condition here -> https://github.com/hashicorp/terraform-provider-tfe/blob/30ef66248f4ba1dcf1524282d5233033f67b3d84/internal/provider/resource_tfe_workspace.go#L906

Cheers,
Antoine

PR #1204, opened 2024-01-04
fix: Swap error messages in validateAgentExecution

Erouan50:
The message "agent_pool_id must be provided when execution_mode is 'agent'" was displayed when the execution mode was remote, and the agent pool ID was set, which was confusing.

Fixes: #1203


brandonc:
Thank you for fixing this!

PR #1254, opened 2024-02-13
Adds `ignore_additional_tag_names` to tfe_workspace resource

brandonc:
## Description

A boolean value that, when set to true, allows additional unmanaged tags to be applied to this workspace outside of terraform without being overwritten by `tag_names` config. This creates exceptional behavior in terraform with respect to `tag_names` and is generally not recommended. But given that the alternative is to ignore all updates to tags after they are created, or not be able to mix unmanaged and managed tags at all, it seems like a reasonable feature until we separate the tag resource altogether

### DIFF
--- a/internal/provider/resource_tfe_workspace.go
+++ b/internal/provider/resource_tfe_workspace.go
@@ -209,6 +209,11 @@ func resourceTFEWorkspace() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 
+			"ignore_additional_tag_names": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
 			"terraform_version": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -529,8 +534,11 @@ func resourceTFEWorkspaceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("agent_pool_id", agentPoolID)
 
 	var tagNames []interface{}
+	managedTags := d.Get("tag_names").(*schema.Set)
 	for _, tagName := range workspace.TagNames {
-		tagNames = append(tagNames, tagName)
+		if managedTags.Contains(tagName) || !d.Get("ignore_additional_tag_names").(bool) {
+			tagNames = append(tagNames, tagName)
+		}
 	}
 	d.Set("tag_names", tagNames)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## UNRELEASED
+
+FEATURES:
+* `r/tfe_workspace`: Add `ignore_additional_tag_names` which explicitly ignores tag_names _not_ defined by config to exist on the workspace without being overwritten when the resource is refreshed, by @brandonc and @mbillow
+
 ## v0.52.0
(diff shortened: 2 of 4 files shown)


ctrombley (approved):
Looks good.

internal/provider/resource_tfe_workspace.go
```diff
@@ -529,8 +534,11 @@ func resourceTFEWorkspaceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("agent_pool_id", agentPoolID)
 
 	var tagNames []interface{}
+	managedTags := d.Get("tag_names").(*schema.Set)
```
Uk1288, line 537: Generally, worked as expected. A scenario which I came accross:

Step 1: Create 2 workspace each with `ignore_additional_tag_names` set to true/false like below and apply
```
# Expectation: should mutate/destroy all unmanaged tags ✅ 
resource "tfe_workspace" "from_false_to_true" {
  name                        = "workspace-from_false_to_true"
  organization                = "hashicorp"
  auto_apply                  = true
  tag_names                   = ["afoo", "abar"]
  ignore_additional_tag_names = false
}

# Expectation: should not change any unmanaged tags ✅ 
resource "tfe_workspace" "from_true_to_false" {
  name                        = "workspace-from_true_to_false"
  organization                = "hashicorp"
  auto_apply                  = true
  tag_names                   = ["bfoo", "bbar"]
  ignore_additional_tag_names = true
}
```

Step 2: add tag names to both workspaces using terraform cloud.
Step 3: run an apply, the workspaces tags are mutated as expected  as stated in the above expectation statements
Step 4: add tag names to both workspaces again using terraform cloud.
Step 5: change `ignore_additional_tag_names` to `true` in `tfe_workspace.from_false_to_true` and vice versa in `tfe_workspace.from_true_to_false`. The config should now look like this:

```
# Expectation: should not change any unmanaged tags ❌ 
resource "tfe_workspace" "from_false_to_true" {
  name                        = "workspace-from_false_to_true"
  organization                = "hashicorp"
  auto_apply                  = true
  tag_names                   = ["afoo", "abar"]
  ignore_additional_tag_names = true
}

# Expectation: should mutate/destroy all unmanaged tags  ❌ 
resource "tfe_workspace" "from_true_to_false" {
  name                        = "workspace-from_true_to_false"
  organization                = "hashicorp"
  auto_apply                  = true
  tag_names                   = ["bfoo", "bbar"]
  ignore_additional_tag_names = false
}
```
Step 6: apply the config.

Notice, that the apply uses previous values of `ignore_additional_tag_names` when deciding to ignore tag_names.
However, a second apply acts as expected, although unexpected tag_names would have been deleted. If tags are recreated, subsequent applies act as expected.

brandonc: Thanks for the repro instructions. I can summarize the issue like this: "the config value is ignored before it is applied because the value is used when reading (refreshing) the resource prior to the plan being generated".

I can not read the raw configuration values from this point in the provider lifecycle so I think we have to depend on the state value alone. I can put a note in the docs. I added "This value must be applied before is will be respected."

Uk1288 (approved):

PR #1267, opened 2024-02-28
[TF-13376] Refactor registry module delete method

laurenolivia:
## Description

The Registry Module `Delete` API deprecated and replaced three endpoints (see [docs](https://developer.hashicorp.com/terraform/enterprise/api-docs/private-registry/modules#delete-a-module)). This ticket is a follow-on, after refactoring the API calls in go-tfe ([PR](https://github.com/hashicorp/go-tfe/pull/847)) and releasing those changes in [v1.45.0](https://github.com/hashicorp/go-tfe/releases/tag/v1.45.0).

_Remember to:_

## Testing plan

1. Create two registry module resources
2.  Make `Name` attr for both resources identical
3.  Make `Provider` attr for both resources different
4.  Delete one registry module
5. Verify the resource was deleted

## External links

```

...
```

### DIFF
--- a/internal/provider/resource_tfe_registry_module.go
+++ b/internal/provider/resource_tfe_registry_module.go
@@ -10,6 +10,7 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -414,15 +415,30 @@ func resourceTFERegistryModuleRead(d *schema.ResourceData, meta interface{}) err
 func resourceTFERegistryModuleDelete(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(ConfiguredClient)
 
-	log.Printf("[DEBUG] Delete registry module: %s", d.Id())
-	organization := d.Get("organization").(string)
-	name := d.Get("name").(string)
-	err := config.Client.RegistryModules.Delete(ctx, organization, name)
-	if err != nil {
-		if err == tfe.ErrResourceNotFound {
-			return nil
+	// Fields required to delete registry module by provider
+	// To delete by name, Provider field is not required
+	rModID := tfe.RegistryModuleID{
+		Organization: d.Get("organization").(string),
+		Name:         d.Get("name").(string),
+		Provider:     d.Get("module_provider").(string),
+		Namespace:    d.Get("namespace").(string),
+		RegistryName: tfe.RegistryName(d.Get("registry_name").(string)),
+	}
+
+	if v, ok := d.GetOk("module_provider"); ok && v.(string) != "" {
+		log.Printf("[DEBUG] Delete registry module by provider: %s", d.Id())
+
+		err := config.Client.RegistryModules.DeleteProvider(ctx, rModID)
+		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
+			return fmt.Errorf("error deleting registry module provider: %w", err)
 		}
-		return fmt.Errorf("Error deleting registry module %s: %w", d.Id(), err)
+	} else {
+		log.Printf("[DEBUG] Delete registry module by name: %s", d.Id())
+
+		err := config.Client.RegistryModules.DeleteByName(ctx, rModID)
+		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
+			return fmt.Errorf("Error deleting registry module %s: %w", d.Id(), err)
+		}
 	}
 
 	return nil
--- a/internal/provider/resource_tfe_registry_module_test.go
+++ b/internal/provider/resource_tfe_registry_module_test.go
@@ -803,6 +803,7 @@ func TestAccTFERegistryModule_invalidWithRegistryNameAndNoModuleProvider(t *test
 		},
 	})
 }
+
 func testAccCheckTFERegistryModuleExists(n string, rmID tfe.RegistryModuleID, registryModule *tfe.RegistryModule) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		config := testAccProvider.Meta().(ConfiguredClient)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ BUG FIXES:
 * `r/tfe_workspace`: Change the error message returned when a workspace cannot be safe-deleted to reflect that the error can happen when the latest state is still being processed @uturunku1 [1274](https://github.com/hashicorp/terraform-provider-tfe/pull/1274)
 
 
+ENHANCEMENTS:
+* `r/tfe_registry_module`: Update `Delete` method to call `DeleteByName` when `module_provider` is not present, and `DeleteProvider` when `module_provider` exists @laurenolivia[1267](https://github.com/hashicorp/terraform-provider-tfe/pull/1267)
+
 ## v0.52.0
 
 FEATURES:


internal/provider/resource_tfe_registry_module.go
```diff
@@ -416,15 +417,30 @@ func resourceTFERegistryModuleRead(d *schema.ResourceData, meta interface{}) err
...
+		Provider:     d.Get("module_provider").(string),
+		Namespace:    d.Get("namespace").(string),
+		RegistryName: tfe.RegistryName(d.Get("registry_name").(string)),
+	}
+
+	if v, ok := d.GetOk("module_provider"); ok && v.(string) != "" {
+		log.Printf("[DEBUG] Delete registry module by provider: %s", d.Id())
+
+		err := config.Client.RegistryModules.DeleteProvider(ctx, rModID)
+		if err != nil && errors.Is(err, tfe.ErrResourceNotFound) {
```
Uk1288, line 434:
```suggestion
		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
```

internal/provider/resource_tfe_registry_module.go
```diff
@@ -416,15 +417,30 @@ func resourceTFERegistryModuleRead(d *schema.ResourceData, meta interface{}) err
...
+			return fmt.Errorf("error deleting registry module provider: %w", err)
+		}
+	} else {
+		log.Printf("[DEBUG] Delete registry module by name: %s", d.Id())
+
+		err := config.Client.RegistryModules.DeleteByName(ctx, rModID)
+		if err != nil && errors.Is(err, tfe.ErrResourceNotFound) {
```
Uk1288, line 441:
```suggestion
		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
```

internal/provider/resource_tfe_registry_module_test.go
```diff
@@ -803,6 +805,107 @@ func TestAccTFERegistryModule_invalidWithRegistryNameAndNoModuleProvider(t *test
...
+						}, registryModule),
+					testAccCheckTFERegistryModuleExists("tfe_registry_module.baz",
+						tfe.RegistryModuleID{
+							Organization: orgName,
+							Provider:     expectedRegistryModuleAttributes.Provider,
+							Name:         expectedRegistryModuleAttributes.Name,
+							RegistryName: expectedRegistryModuleAttributes.RegistryName,
+							Namespace:    expectedRegistryModuleAttributes.Namespace,
+						}, registryModule),
+					testAccTFERegistryModule_deleteModuleProvider("tfe_registry_module.foo",
```
Uk1288, line 854: this check function seems to delete the registry module as opposed to checking for an expected condition, can you say more about this check?

internal/provider/resource_tfe_registry_module.go
brandonc: Good eye

internal/provider/resource_tfe_registry_module_test.go
laurenolivia: Yep, I interpreted the language here to mean I can use `Check` to make a custom API call. In this case, I am testing the `DeleteProvider` endpoint. Realizing `check the status of things` might only mean existence or not.
```
// Check is called after the Config is applied. Use this step to
// make your own API calls to check the status of things, and to
// inspect the format of the ResourceState itself.
```

internal/provider/resource_tfe_registry_module.go
```diff
@@ -416,15 +417,30 @@ func resourceTFERegistryModuleRead(d *schema.ResourceData, meta interface{}) err
...
+		err := config.Client.RegistryModules.DeleteProvider(ctx, rModID)
+		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
+			return fmt.Errorf("error deleting registry module provider: %w", err)
+		}
+	} else {
+		log.Printf("[DEBUG] Delete registry module by name: %s", d.Id())
+
+		err := config.Client.RegistryModules.DeleteByName(ctx, rModID)
+		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
+			return fmt.Errorf("Error deleting registry module %s: %w", d.Id(), err)
```
sebasslash, line 442: So I have some pretty strong opinions against mix & matching different endpoints within a single resource operation depending on the attributes defined. While it's totally possible to move forward with this, I would prefer not to introduce the footgun of having either `module_provider` defined resulting in "the specified provider for the given module is deleted along with all its versions" ([docs](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/private-registry/modules#parameters-2)). Or, omitting the `module_provider` attribute which results in the entire module being deleted. 

With any specific Terraform resource, I expect providers to manage a single, real "thing"... be it a relationship between two API resources or the API resource itself. I'd prefer to avoid introducing side effects that can cause undesirable behavior. 

In my opinion, a better approach would be to add new `tfe_registry_module_version`, which manages a specific version for a module, and `tfe_registry_module_provider`, which manages a specific provider for a module. We can then leave this existing resource `tfe_registry_module` to manage the entire module. Each resource would then call one of the three possible delete endpoints for modules. 

cc @brandonc

sebasslash (changes requested):
Apologies, blocking following a pending discussion. ⬇️

internal/provider/resource_tfe_registry_module.go
laurenolivia: FWIW, the original intent of this PR was to fix a bug where the `Delete()` method was calling a deprecated endpoint (see [docs](https://developer.hashicorp.com/terraform/enterprise/api-docs/private-registry/modules#delete-a-module)). I think for the purpose of resolving that bug for the customer, we can address it by updating the Delete() method to call `DeleteByName` - and then separate any additional work into new tickets.

brandonc: We had an offline conversation and decided that, while it would be better to have discreet records, this is still a valuable bug fix

internal/provider/resource_tfe_registry_module_test.go
```diff
@@ -803,6 +803,27 @@ func TestAccTFERegistryModule_invalidWithRegistryNameAndNoModuleProvider(t *test
...
+
+func TestAccTFERegistryModule_deleteModuleProvider(t *testing.T) {
+	org, _ := setupDefaultOrganization(t)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckTFERegistryModuleDestroy,
```
laurenolivia, line 815: A question that has come up is whether we need to build a test case where  resources are destroyed if we have `CheckDestroy`?

sebasslash, line 815: Given that each top-level acceptance test makes use of `resource.Test()` which executes `terraform destroy` as part of the test cleanup, ensuring there are no failures when `Delete()` is called, it feels somewhat redundant to test explicitly for a destroy. `testAccCheckTFERegistryModuleDestroy` is a check that more appropriately belongs in go-tfe integration tests.

sebasslash (approved):
Great work 👍

internal/provider/resource_tfe_registry_module_test.go
laurenolivia: Thanks @sebasslash and @uturunku1. In that case, let's avoid adding code we don't need, or creating tech debt. This was thoroughly tested in go-tfe, and after the merge I can monitor any incoming requests/bugs related to this refactor. 👍

internal/provider/resource_tfe_registry_module.go
```diff
@@ -414,15 +415,30 @@ func resourceTFERegistryModuleRead(d *schema.ResourceData, meta interface{}) err
...
+
+		err := config.Client.RegistryModules.DeleteProvider(ctx, rModID)
+		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
+			return fmt.Errorf("error deleting registry module provider: %w", err)
+		}
+	} else {
+		log.Printf("[DEBUG] Delete registry module by name: %s", d.Id())
+
+		err := config.Client.RegistryModules.DeleteByName(ctx, rModID)
+		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
```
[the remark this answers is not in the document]
laurenolivia: This is a check for err `!` + `tfe.ErrResourceNotFound`

sebasslash (approved):
Apologies. I completely missed the `!`.

PR #1296, opened 2024-03-25
Bump nightly instance job timeout to 20m

sebasslash:
Timeouts are resulting workflow failures since runs are taking longer than 10m to complete.


laurenolivia (approved):
Sure, let's try it.

PR #1330, opened 2024-04-23
Add Global Run Task support 

glennsarti:
## Description

This PR;

* Updates the workspace run task resource and datasource to deprecate the `stage` attribute
  in favour of the new `stages` attribute. This also includes a state version update
* Adds a new organization_run_task_global_settings resource and datasource for managing the
  global settings of a Run task. This does require the HCP Terraform/TFE instance to support
  global run tasks. The provider does attempt to error gracefully if a user is trying to use
  this feature if it's unsupported.

_Remember to:_

## Testing plan

Handled by acceptance tests

## External links

- [go-tfe documentation](https://github.com/hashicorp/go-tfe/pull/865)
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/run-tasks/run-tasks)

In this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ FEATURES:
 * `r/tfe_team`: Add attribute `manage_agent_pools` to `organization_access` on `tfe_team` by @emlanctot [#1358](https://github.com/hashicorp/terraform-provider-tfe/pull/1358)
 * `r/tfe_workspace`: Add an `auto_destroy_activity_duration` attribute for automatic scheduling of auto-destroy runs based off of workspace activity, by @notchairmk [#1377](https://github.com/hashicorp/terraform-provider-tfe/pull/1377)
 * `d/tfe_workspace`: Add an `auto_destroy_activity_duration`, by @notchairmk [#1377](https://github.com/hashicorp/terraform-provider-tfe/pull/1377)
+* `d/tfe_organization_run_task_global_settings`: Add a datasource to retrieve the global settings of Run tasks, by @glennsarti [#1328](https://github.com/hashicorp/terraform-provider-tfe/pull/1330)
+* `r/tfe_organization_run_task_global_settings`: Add a resource to manage the global settings of Run tasks, by @glennsarti [#1328](https://github.com/hashicorp/terraform-provider-tfe/pull/1330)
+
+DEPRECATIONS and BREAKING CHANGES:
+* `r/_workspace_run_task`: The `stage` attribute has been deprecated in favor of the `stages` attribute, by @glennsarti [#1328](https://github.com/hashicorp/terraform-provider-tfe/pull/1330)
+* `d/_workspace_run_task`: The `stage` attribute has been deprecated in favor of the `stages` attribute, by @glennsarti [#1328](https://github.com/hashicorp/terraform-provider-tfe/pull/1330)
 
 ## v0.56.0
 
--- a/internal/provider/resource_tfe_organization_run_task_global_settings.go
+++ b/internal/provider/resource_tfe_organization_run_task_global_settings.go
@@ -1,0 +1,298 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package provider
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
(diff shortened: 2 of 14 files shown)


glennsarti:
Added compliance team for review as a once-over.

internal/provider/resource_tfe_organization_run_task_global_settings.go
brandonc, line 1: Can you help me understand why this schema was not added to organization_run_task?

In the case of tfe_organization_default_settings, it's very common to not manage the organization itself with the provider... and in the case of tfe_workspace_settings, the primary issue was breaking the circular dependency between agent pools and workspaces but it was also a convenient symmetry with organization_settings because it  overrides 'inheritability' of those settings.

In the case of organization run tasks, neither problem seems to appear so it seems like it would be much straightforward to put all schema within the same resource. Is there a concept I am overlooking?

brandonc (commented):
Let's start with a question 👇

internal/provider/resource_tfe_organization_run_task_global_settings.go
brandonc: Another concern is "enabled" schema would appear to be managing the same thing on both resources

glennsarti: So there a couple of things that forced me down this path;

1. Global Run Tasks are entitlement based.
So it's possible for their functionality to be removed/added outside of the provider.

    * Imagine the run task resource includes global information in it's state.
    * A run task resource is created _without_ global run tasks being available.
    * An admin comes along and upgrades their TFC subscription to premium
    * Now the next run the resource will be out of sync because its state has changed. (from null to _something_)

Same applies when a resource wants to manage a global run task, but the subscription was demoted. (from _something_ to null)

2. We are still using [Plugin Protocol v5](https://github.com/hashicorp/terraform-provider-tfe/blob/5192223788f67f732ca8f3aa85859f1e24d91c0b/internal/provider/plugin_provider.go#L12) so I don't have have access to nested attributes. I tried so many different combinations of [hacks](https://github.com/hashicorp/terraform-provider-tfe/blob/5192223788f67f732ca8f3aa85859f1e24d91c0b/internal/provider/resource_tfe_workspace_settings.go#L239-L241) but I could never get it quite right.

---

I desperately wanted to use a nested block syntax, but I just couldn't, given the dynamic nature of the subscription turning on/off the global run task feature.

I'd prefer

```
org_run_task "thing" {
  ...
  global = {
    enabled = true
    enforcement-level = "advisory"
  }
}
```

website/docs/r/organization_run_task_global_settings.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+Basic usage:
+
+```hcl
+resource "tfe_organization_run_task_global_settings" "example" {
+		task_id = tfe_organization_run_task.example.id
+		enabled           = false
+		enforcement_level = "advisory"
+		stages            = ["pre_plan", "post_plan"]
+}
```
brandonc, line 24: Can you provide the whole example where `tfe_organization_run_task.example` is included as well?

Does `enabled` need to be an argument in both resources?

glennsarti: Added the missing example, and set enabled to true in both to avoid confusion.

> Does enabled need to be an argument in both resources?
Yes.

website/docs/d/organization_run_task_global_settings.html.markdown
```diff
@@ -0,0 +1,34 @@
+---
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_organization_run_task_global_settings"
+description: |-
+  Get information on a Run task global settings.
```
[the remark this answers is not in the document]
glennsarti: Fixed

CHANGELOG.md
```diff
@@ -6,10 +6,16 @@ ENHANCEMENTS:
 * `d/tfe_workspace`: Add an `auto_destroy_at` attribute for reading a scheduled auto-destroy, by @notchairmk [1354](https://github.com/hashicorp/terraform-provider-tfe/pull/1354)
 * `r/tfe_registry_module`: Add `initial_version` support for Branch Based Modules by @aaabdelgany [#1363](https://github.com/hashicorp/terraform-provider-tfe/pull/1363)
 * `r/tfe_oauth_client`: Add `agent_pool_id` as an optional argument to enable Private VCS support, by @roleesinhaHC [1255](https://github.com/hashicorp/terraform-provider-tfe/pull/1255)
+* `d/tfe_organization_run_task_global_settings`: Added a datasource to retrieve the global settings of Run tasks, by @glennsarti [#1328](https://github.com/hashicorp/terraform-provider-tfe/pull/1330)
```
mrinalirao, line 9: nit: are these new data source/resources? Should it be under features instead of enhancement. Also just for consistency with the other changelog entries, we could active voice instead of passive (Adds instead of Added).

internal/provider/resource_tfe_workspace_run_task_schemas.go
```diff
@@ -0,0 +1,164 @@
...
+			ElementType: types.StringType,
+			Description: fmt.Sprintf("The stages to run the task in. Valid values are %s.", sentenceList(
+				workspaceRunTaskStages(),
+				"`",
+				"`",
+				"and",
+			)),
+			// Default: listdefault.StaticValue(basetypes.NewListValueMust(types.StringType, []attr.Value{
+			// 	basetypes.NewStringValue(string(tfe.PostPlan)),
+			// })),
```
mrinalirao, line 152: Should stages have a default? 🤔 If not, we should probably remove this^

CHANGELOG.md
glennsarti: Done

internal/provider/resource_tfe_workspace_run_task_schemas.go
glennsarti: Good call!  fixed

glennsarti:
@brandonc or @ctrombley Am I ok to merge?

glennsarti:
Rebased due to changelog

PR #1385, opened 2024-06-14
Support for Data Retention Policies

SwiftEngineer:
## Description

Add support for Data Retention Policies, only available in Terraform Enterprise

```
=== RUN   TestAccTFEDataRetentionPolicy_basic
--- PASS: TestAccTFEDataRetentionPolicy_basic (13.06s)
=== RUN   TestAccTFEDataRetentionPolicy_dontdelete_basic
--- PASS: TestAccTFEDataRetentionPolicy_dontdelete_basic (8.32s)
=== RUN   TestAccTFEDataRetentionPolicy_explicit_organization
--- PASS: TestAccTFEDataRetentionPolicy_explicit_organization (7.81s)
=== RUN   TestAccTFEDataRetentionPolicy_update_type
--- PASS: TestAccTFEDataRetentionPolicy_update_type (10.22s)
=== RUN   TestAccTFEDataRetentionPolicy_implicit_organization
--- PASS: TestAccTFEDataRetentionPolicy_implicit_organization (8.06s)
=== RUN   TestAccTFEDataRetentionPolicy_dontdelete_organization_level
--- PASS: TestAccTFEDataRetentionPolicy_dontdelete_organization_level (7.44s)
PASS
```


internal/provider/resource_tfe_data_retention_policy.go
```diff
@@ -0,0 +1,368 @@
...
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"organization": schema.StringAttribute{
+				Description: "Name of the organization. If omitted, organization must be defined in the provider config.",
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
```
JarrettSpiker, line 58: Would it be possible to add a validation that either organization or workspace is specified?
Edit: Or maybe both? I see that both is possible now...I feel like it would be acceptable to say just one or the other should be specified, to reduce ambiguity. But I dont feel strongly

internal/provider/resource_tfe_data_retention_policy.go
```diff
@@ -0,0 +1,368 @@
...
+					stringplanmodifier.UseStateForUnknown(),
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"workspace_id": schema.StringAttribute{
```
JarrettSpiker, line 63: I am able to generate errors by creating a config with multiple policies specified for the same workspace.

```
resource "tfe_data_retention_policy" "workspace1_policy" {
  workspace_id = tfe_workspace.workspace1.id
  delete_older_than {
    days = 3
  }
}

resource "tfe_data_retention_policy" "workspace1_policy2" {
  workspace_id = tfe_workspace.workspace1.id
  delete_older_than {
    days = 6
  }
}
```


If I start with one policy and then add the other, the policy is replaced with the new value (e.g. it goes from 3 days to 6 days). Then every subsequent apply detects drift, and tries to update the policy back to the other value (3 -> 6 -> 3 -> 6 -> .....)

If I try to modify both policies at once, I get an error as it attempts to delete the same policy twice.

JarrettSpiker: If I delete one of the configs, it deletes the policy on TFE (as expected). Then further applies crash with the error

```
│ The plugin.(*GRPCProvider).ReadResource request was cancelled.
╵

Stack trace from the terraform-provider-tfe plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x1027f0434]

```

SwiftEngineer: Good catch 👍 I knew having multiple resources "overriding each other" on a given workspace or organization would be problematic, but we've ignored that problem in the past. That being said, I didn't expect the removal to cause a nasty error like that. I pushed a "fix" that at least allows for users to remove policies attached to the same workspace/organization.

internal/provider/resource_tfe_data_retention_policy.go
```diff
@@ -0,0 +1,368 @@
...
+	// Save data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &result)...)
+}
+
+func (r *resourceTFEDataRetentionPolicy) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	// If the resource does not support modification and should always be recreated on
+	// configuration value updates, the Update logic can be left empty and ensure all
+	// configurable schema attributes implement the resource.RequiresReplace()
+	// attribute plan modifier.
+	resp.Diagnostics.AddError("Update not supported", "The update operation is not supported on this resource. This is a bug in the provider.")
```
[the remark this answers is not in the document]
SwiftEngineer: Dang, I thought having the requirereplace modifier on the number would be enough 🤔 good catch 👍

internal/provider/resource_tfe_data_retention_policy.go
SwiftEngineer: I left in it because I saw that seemed to be the pattern for some other resources, but I think you're right that it makes a lot more sense this way 👍

internal/provider/resource_tfe_data_retention_policy.go
```diff
@@ -0,0 +1,368 @@
...
+			},
+			"workspace_id": schema.StringAttribute{
+				Description: "ID of the workspace that the data retention policy should apply to. If omitted, the data retention policy will apply to the entire organization.",
+				Optional:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+		},
+		Blocks: map[string]schema.Block{
```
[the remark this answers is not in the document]
SwiftEngineer: I prefer that too, but I moved away for two reasons:
1. Manuel and the core team had already agreed on a design and I didn't wanna stir the pot.
2. Far less "importantly", but I kinda like the idea of reducing the number TFE-only resources that we have.

JarrettSpiker (approved):
Tested again, and looks good! Thanks Taylor

PR #1398, opened 2024-07-01
Add AllowMemberTokenManagement to Team

juliannatetreault:
## Description

This PR adds a new team permission - `AllowMemberTokenManagement` - to the `Team` block. `AllowMemberTokenManagement` is `true` by default and can only be managed by Owners or users with "Manage Teams" permissions. Additionally, this PR adds tests and documentation for the new permission.

_Remember to:_

## Testing plan

1. `allow_member_token_management` should default to `true` for new teams.
2. If `allow_member_token_management` is omitted, it should default to`true`.
3. `allow_member_token_management` should be able to be set or unset via the provider.

## External links

- [go-tfe PR](https://github.com/hashicorp/go-tfe/pull/922)
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/permissions#allow-member-token-management)

```

...
```

### DIFF
--- a/internal/provider/resource_tfe_team.go
+++ b/internal/provider/resource_tfe_team.go
@@ -144,7 +144,6 @@ func resourceTFETeam() *schema.Resource {
 			"allow_member_token_management": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  true,
 			},
 		},
 	}
--- a/internal/provider/resource_tfe_team.go
+++ b/internal/provider/resource_tfe_team.go
@@ -195,9 +195,7 @@ func resourceTFETeamCreate(d *schema.ResourceData, meta interface{}) error {
 		options.SSOTeamID = tfe.String(v.(string))
 	}
 
-	if v, ok := d.GetOk("allow_member_token_management"); ok {
-		options.AllowMemberTokenManagement = tfe.Bool(v.(bool))
-	}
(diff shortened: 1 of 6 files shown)


JarrettSpiker:
This looks good to me. Will just need a rebase once the go-tfe changes are available

internal/provider/resource_tfe_team.go
```diff
@@ -141,6 +141,11 @@ func resourceTFETeam() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"allow_member_token_management": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
brandonc, line 147: Does this need a provider default? The side effect of this choice is that when it is not defined in config, it will be set to true, overriding anything that may be preexisting in the API. Since the API defines its own default, maybe it's best to accept whatever value it returns... Because older TFE may not understand this attribute and will always appear to be false, which (I think) will constantly be a planned change that cannot be changed even if the attribute is not in config.

juliannatetreault: Ah, that's a fair point! Thank you for this feedback! I think it would make sense to go without a default then and rely on the API's value.

Any objections to this, @emlanctot?

emlanctot: i agree with @brandonc that there should not be a default on this!

emlanctot (approved):
looks good to me ! ✨ 🛹 💯

PR #1448, opened 2024-08-23
Bump go-tfe for stack decoding fix and prep release

brandonc:
## Description

Actual fix is in associated upstream go-tfe release

I intend to release this as v0.58.1


ctrombley (approved):
🚢

PR #1521, opened 2024-11-12
Update enforcement mode for OPA policy resources

### LINKED ISSUE #1407: tfe_policy does not actually alter enforce_mode on OPA policies

TFCB

#### Terraform version

```plaintext
Terraform v1.9.0
on darwin_amd64
+ provider registry.terraform.io/hashicorp/tfe v0.56.0
```

```terraform
resource "tfe_policy" "policy" {
  organization = "..."
  description  = "My policy"
  # Set this to `advisory` during create, then try applying a new plan after changing it to `mandatory`.
  enforce_mode = "advisory"
  kind         = "opa"
  name         = "test_policy"
  query        = "data.whatever"
  policy       = "deny {\n  false\n}"
}
```

#### Debug Output

The salient output from the debug log seems to be the following:

```plaintext
2024-07-05T18:50:56.601-0700 [WARN]  Provider "provider[\"registry.terraform.io/hashicorp/tfe\"]" produced an unexpected new value for tfe_policy.policy, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .enforce_mode: was cty.StringVal("mandatory"), but now cty.StringVal("advisory")
```

I'll send a complete log in a support ticket.

#### Expected Behavior

The provider should configure the policy in `mandatory` enforcement mode.

#### Actual Behavior

The provider configures the policy in `advisory` enforcement mode.

#### Additional Context

I've tried this from our own Terraform Cloud agents, and tried running this locally. No dice.

glennsarti:
## Description

Fixes #1407

Previously the tfe_policy resource could only set policies to advisory. This commit updates the update function to properly switch path for the policy enforcement name based on the kind. This was due to the previous comparison not comparing the same types so it always failed.

_Remember to:_

## Testing plan

Handled by additional acceptance test

### DIFF
--- a/internal/provider/resource_tfe_policy.go
+++ b/internal/provider/resource_tfe_policy.go
@@ -271,6 +271,11 @@ func resourceTFEPolicyRead(d *schema.ResourceData, meta interface{}) error {
 func resourceTFEPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(ConfiguredClient)
 
+	var kind string
+	if v, ok := d.GetOk("kind"); ok {
+		kind = v.(string)
+	}
+
 	// nolint:nestif
 	if d.HasChange("description") || d.HasChange("enforce_mode") || d.HasChange("query") {
 		// Create a new options struct.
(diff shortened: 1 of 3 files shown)


internal/provider/resource_tfe_policy.go
```diff
@@ -283,7 +283,7 @@ func resourceTFEPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
 		path := d.Get("name").(string) + ".sentinel"
 		vKind, ok := d.GetOk("kind")
```
sebasslash, line 284: Alternatively:
```go
vKind, ok := d.GetOK("kind").(string)
...
```
It also seems to be cast again in L313 under `if d.HasChange("policy") {}`. Maybe we can consolidate this a bit.

sebasslash (approved):
Looks good!

internal/provider/resource_tfe_policy.go
glennsarti: Done

mrinalirao (approved):
🚢

PR #1524, opened 2024-11-14
Adds remote state consumers to tfe_workspace_settings

### LINKED ISSUE #331: tfe_workspace has possible self-referencial block -> pull remote_state_consumer_ids out into separate resource
#### Use-cases
My personal case is I have a `tfe_workspace` for every microservice+environment combination.
Workspaces are dependent on each other and across environments. An example section of my `.tf` file looks like this:

```
  for_each                  = toset(["qa","production"])
  name                      = "${each.value}-foo"
  global_remote_state       = false
  terraform_version         = "1.0.0"
  remote_state_consumer_ids = toset(compact([tfe_workspace.bar[each.value].id]))
  # ...
}

resource "tfe_workspace" "bar" {
  for_each                  = toset(["qa","production"])
  name                      = "${each.value}-bar"
  global_remote_state       = false
  terraform_version         = "1.0.0"
  remote_state_consumer_ids = toset(compact([each.value == "production" ? tfe_workspace.bar["qa"].id : ""]))
}
```

Note in the first case, `qa-foo` and `production-foo` are dependent on `qa-bar` and `production-bar` respectively. This works as planned.

The second case, `qa-bar` has a cross-environment dependency on `production-bar`. This causes the plan to error. This error is triggered by `production-bar` being defined in the same top-level block as `qa-bar`. So `production-bar` cannot reference `qa-bar` in it's own block (even though they are separate resources).

_(Some terminology clarification) The real world dependency is `qa-bar` depends on `production-bar`. In terraform `production-bar` must declare this dependency, so that makes `production-bar` dependent on `qa-bar` in this case_

The actual error is that `production-bar["qa"]` cannot reference `production-bar["qa]`, despite the variable not being used in that "execution" of the ternary.
(Name changed from real case)
```
Error: Self-referential block
# ...file location...
Configuration for tfe_workspace.bar["qa"] may not refer to itself.
```

Considering that. `tfe_workspace` can logically depend on other resources of the same type, it would be useful to have the option to declare this connection in a _separate_ resource, so terraform can deterministically figure out dependencies.

#### Attempted Solutions

1) `remote_state_consumer_ids ` is only necessary when `global_remote_state=false`. So I _could_ just open all workspaces up to one-another. For general security/design reasons I would prefer to not do that in name of [POLA](https://en.wikipedia.org/wiki/Principle_of_least_privilege).
2) Hard-code that edge case. In my case these workspaces already exist, so `id` can be known in advance by me and hard-coded. So the line would change to something like `each.value == "production" ? "abc123" : ""`. This is my current work around.
3) Don't use `for_each` to group resources, and declare each separately. In hindsight this would be the best design for this solution, but I don't have that currently implemented. Migrating to this would mean manually updating each resource pragmatically.

#### Proposal
Create a new resource `tfe_workspace_remote_consumers`. It will take only two parameters of `workspace_id` and `remote_state_consumer_ids`.

This ensures any workspace can depend on any other workspace. For backwards compatibility and simplicity, the `tfe_workspace` `remote_state_consumer_ids` can continue to act as normal.

This pattern of multiple types of resources being able to manipulate the same attributes is used other places like:
`aws_sqs_queue` and `aws_sqs_queue_policy` https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue_policy
`aws_security_group` and `aws_security_group_rule` https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule

Similar to those implementations, a note should be included that users should only pick one. Similarly the `tfe_workspace` `remote_state_consumer_ids` should expect its value to be possibly set elsewhere if not explicitly defined. Although `lifecycle { ignore_changes: [remote_state_consumer_ids] }` should do the trick.

#### Request
While this feature would solve my issue, it seems possible there are other solutions around this. If someone could help with those that would be great.

Two thoughts are:
1) Somehow use the `self` value. `provisioner` can access its own resource with the `self` value. I guess a feature request to terraform core would be granting some `self` equivalent to the resource itself. Except I don't really want the resource... I want the group one level higher, with the other resource keyed by environment.
2) Only generate the reference sometimes... My understanding of the _actual_ error is that when `each.value == "production" ? tfe_workspace.bar["qa"].id : ""` is calculated, `tfe_workspace.bar["qa"].id`'s value is calculated even though the value is not eventually referenced when `each.value == "qa"`. If there was some way of _not_ doing that and instead tell terraform "don't calculate this value until absolutely necessary" that could also potentially solve this issue. Again this is probably a feature request to terraform core instead.

brandonc:
## Description

Moving this attribute to the auxiliary workspace resource removes possible mutual dependence when adding workspace IDs to tfe_workspace resources.

_Remember to:_

- [X] _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)_

This is the most 👍  open issue on the repo.

## Testing plan

Try this config:

```terraform
provider "tfe" {}

resource "tfe_organization" "example" {
  name  = "example-remote-state"
  email = "you@example.com"
}

  for_each     = toset(["qa","production"])
  name         = "${each.value}-test"
  organization = tfe_organization.example.name
}

resource "tfe_workspace_settings" "test-settings" {
  for_each                  = toset(["qa","production"])
  workspace_id              = tfe_workspace.test[each.value].id
  global_remote_state       = false
  remote_state_consumer_ids = toset(compact([each.value == "production" ? tfe_workspace.test["qa"].id : ""]))
}

```

If remote_state_consumer_ids were defined on tfe_workspace, this kind of config would be invalid.

Closes #331 

```
$ TESTARGS="-run TestAccTFEWorkspaceSettings" make testacc
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEWorkspaceSettings -timeout 15m
=== RUN   TestAccTFEWorkspaceSettings
2024/11/14 10:38:04 [DEBUG] Configuring client for host "tfcdev-xxx.ngrok.app"
2024/11/14 10:38:04 [DEBUG] Service discovery for tfcdev-xxx.ngrok.app at https://tfcdev-xxx.ngrok.app/.well-known/terraform.json
--- PASS: TestAccTFEWorkspaceSettings (10.96s)
=== RUN   TestAccTFEWorkspaceSettingsRemoteState
--- PASS: TestAccTFEWorkspaceSettingsRemoteState (8.72s)
=== RUN   TestAccTFEWorkspaceSettingsImport
--- PASS: TestAccTFEWorkspaceSettingsImport (6.13s)
=== RUN   TestAccTFEWorkspaceSettingsImport_ByName
--- PASS: TestAccTFEWorkspaceSettingsImport_ByName (6.22s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider   32.669s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+DEPRECATIONS:
+* `r/tfe_workspace`: `global_remote_state` and `remote_state_consumer_ids` have been **deprecated** and moved to `tfe_workspace_settings` (see ENHANCEMENTS below for more details)
+
 FEATURES:
 * `r/tfe_audit_trail_token` is a new resource for managing audit trail tokens in organization, by @glensarti and @c4po [1533](https://github.com/hashicorp/terraform-provider-tfe/pull/1533)
 
@@ -7,6 +10,9 @@ FEATURES:
 
 BUG FIXES:
 * `r/tfe_policy`: enforcement level can be updated on OPA policies by @glennsarti [#1521](https://github.com/hashicorp/terraform-provider-tfe/pull/1521)
+
+ENHANCEMENTS:
+* `r/tfe_workspace_settings`: `global_remote_state` and `remote_state_consumer_ids` can now be managed using `tfe_workspace_settings`. This enhancement avoids the possibility of a mutual dependency between two or more workspaces that may access each others' state by @brandonc [#1524](https://github.com/hashicorp/terraform-provider-tfe/pull/1524)
 
 ## v0.60.0
 
--- a/internal/provider/resource_tfe_workspace.go
+++ b/internal/provider/resource_tfe_workspace.go
@@ -51,10 +51,6 @@ func resourceTFEWorkspace() *schema.Resource {
 				return err
 			}
 
-			if err := validateRemoteState(c, d); err != nil {
-				return err
-			}
-
 			if err := validateTagNames(c, d); err != nil {
 				return err
 			}
@@ -156,16 +152,18 @@ func resourceTFEWorkspace() *schema.Resource {
 			},
 
 			"global_remote_state": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Computed: true,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Computed:   true,
+				Deprecated: "Use resource `tfe_workspace_settings` to modify the workspace `global_remote_state`. `global_remote_state` on `tfe_workspace` is no longer validated properly and will be removed in a future release of the provider.",
 			},
 
 			"remote_state_consumer_ids": {
-				Type:     schema.TypeSet,
-				Optional: true,
-				Computed: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
+				Type:       schema.TypeSet,
+				Optional:   true,
+				Computed:   true,
+				Elem:       &schema.Schema{Type: schema.TypeString},
+				Deprecated: "Use resource `tfe_workspace_settings` to modify the workspace `remote_state_consumer_ids`. `remote_state_consumer_ids` on `tfe_workspace` is no longer validated properly on this resource and This attribute will be removed in a future release of the provider.",
 			},
 
 			"assessments_enabled": {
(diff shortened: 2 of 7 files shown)


ctrombley (dismissed):
Looks good!

brandonc:
@ctrombley had to fix a conflict in the changelog. Re-✅ ?

ctrombley (approved):
✨

PR #1543, opened 2024-12-11
Add `tags` attribute to tfe_project, tfe_workspace, and tfe_workspace_ids

sebasslash:
## Description

Adding the new KV tagging attribute `tags` to each of the following resources: `rd/tfe_project`, `rd/tfe_workspace`, and `d/tfe_workspace_ids`. This change also deprecates the single value legacy tags in favor of the new `tags` attribute. 

_Remember to:_

## Testing plan

1.  _the conditions_

## External links

```

...
```


internal/provider/resource_tfe_workspace.go
```diff
@@ -513,6 +532,21 @@ func resourceTFEWorkspaceRead(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error reading configuration of workspace %s: %w", id, err)
 	}
 
+	tagBindings, err := config.Client.Workspaces.ListTagBindings(ctx, workspace.ID)
+	if err != nil {
+		if errors.Is(err, tfe.ErrResourceNotFound) {
```
brandonc, line 537: This would be a problem for any Terraform Enterprise version that does not have this endpoint. Maybe this particular 404 should not remove the workspace from state

internal/provider/resource_tfe_workspace.go
```diff
@@ -722,6 +757,15 @@ func resourceTFEWorkspaceUpdate(d *schema.ResourceData, meta interface{}) error
 			}
 		}
 
+		if tagBindings, ok := d.Get("tag_bindings").(map[string]interface{}); ok {
```
brandonc, line 760: Removing tag_bindings config or setting it to `{}` did not unset the tags.

brandonc (changes requested):
Found a couple of issues 👇

internal/provider/resource_tfe_workspace.go
sebasslash: Investigating a bit further the culprit is actually the `omitempty` in go-tfe, which omits empty slices from being sent over the wire. This would require a change to add a `RemoveTagBindings()` receiver for projects/workspaces that explicitly sends an empty array to remove all bindings.

sebasslash: Fixed. Outright removing the field won't do anything since the attribute is also computed; but setting it to `{}` will intentionally delete all fields.

internal/provider/data_source_project.go
```diff
@@ -58,6 +58,12 @@ func dataSourceTFEProject() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+
+			"effective_tags": {
+				Type:     schema.TypeMap,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
```
sebasslash, line 66: For the reviewer: Thinking about this a little more, workspace/project data sources should simply have `tags` be the list of effective tags and leave direct tag bindings for each resource. Therefore there would be no `effective_tags` attribute on these. Wdyt?

website/docs/r/project.html.markdown
```diff
@@ -34,6 +52,12 @@ The following arguments are supported:
     *  TFE versions v202405-1 and later support between 3-40 characters
 * `organization` - (Optional) Name of the organization. If omitted, organization must be defined in the provider config.
 * `description` - (Optional) A description for the project.
+* `tags` - (Optional) A map of key-value tags to add to the project.
+* `ignore_additional_tags` - (Optional) Explicitly ignores `tags`
```
ctrombley, line 56: If we don't recommend it, should we even include it?

Is there enough precedent elsewhere for this kind of configurability to back up including it in the initial k/v tags release?

ctrombley (commented):
Nice work, getting close! Got a few comments for discussion.

website/docs/r/project.html.markdown
sebasslash: Yes `ignore_additional_tags` prevents accidentally overriding tags that may have been added outside of configuration. This is helpful in scenarios where you want to ensure certain tags are present on the target, but you do not wish to manage all tags on the target.

internal/provider/data_source_project.go
```diff
@@ -77,39 +90,56 @@ func dataSourceTFEProjectRead(ctx context.Context, d *schema.ResourceData, meta
...
+			readOptions.PageNumber = wl.NextPage
+		}
+
+		effectiveTagBindings := make(map[string]interface{})
+		effectiveBindings, err := config.Client.Projects.ListEffectiveTagBindings(ctx, proj.ID)
+		if err != nil && !errors.Is(err, tfe.ErrResourceNotFound) {
+			return diag.Errorf("Error retrieving effective tag bindings for project %s: %v", proj.ID, err)
 		}
+		if err != nil {
+			// This endpoint may not be supported against a given TFE instance.
```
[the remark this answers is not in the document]
sebasslash: Correct, range over nil results in a panic so we have to catch `404` errors and assign the bindings to an empty slice. I think the log is helpful if you are attempting to use this field against a TFE instance that does not support tags.

internal/provider/data_source_workspace_test.go
```diff
@@ -222,6 +222,8 @@ func TestAccTFEWorkspaceDataSource_readAutoDestroyAt(t *testing.T) {
 }
 
 func TestAccTFEWorkspaceDataSource_readAutoDestroyDuration(t *testing.T) {
+	skipUnlessAfterDate(t, time.Date(2025, time.February, 16, 0, 0, 0, 0, time.UTC))
```
ctrombley, line 225: What's this about?

website/docs/d/workspace.html.markdown
```diff
@@ -55,7 +55,7 @@ In addition to all arguments above, the following attributes are exported:
 * `speculative_enabled` - Indicates whether this workspace allows speculative plans.
 * `ssh_key_id` - The ID of an SSH key assigned to the workspace.
 * `structured_run_output_enabled` - Indicates whether runs in this workspace use the enhanced apply UI.
-* `tags` - A map of key-value tags associated with the workspace, including any inherited tags from the parent project.
+* `effective_tags` - A map of key-value tags associated with the workspace, including any inherited tags from the parent project.
```
ctrombley, line 58: 💅 
Consider:
```suggestion
* `effective_tags` - A map of key-value tags associated with the workspace, including any tags inherited from the parent project.
```

website/docs/d/workspace_ids.html.markdown
```diff
@@ -74,12 +74,12 @@ The following arguments are supported. At least one of `names` or `tag_names` mu
 
     To select _all_ workspaces for an organization, provide a list with a single
     asterisk, like `["*"]`. The asterisk also supports partial matching on prefix and/or suffix, like `[*-prod]`, `[test-*]`, `[*dev*]`.
-* `filter_tags` - (Optional) A set of key-value tag filters to search for workspaces.
+* `tag_filters` - (Optional) A set of key-value tag filters to search for workspaces.
```
ctrombley, line 77: 💅 
```suggestion
* `tag_filters` - (Optional) A set of key-value tags used to filter the search results.
```

ctrombley (commented):
Good stuff, seems like the last changed simplified things a bit. I only have some minor cosmetic suggestions.

internal/provider/data_source_workspace_test.go
sebasslash: These are temporary skips since there are some known issues with auto-destroy-duration ILM is patching. I've set an arbitrary date to fail these tests again if we forget to remove these skips once the fix is introduced.

brandonc (approved):

brandonc (approved):

brandonc (approved):
high-five-to-myself-high-five

ctrombley (approved):

PR #1604, opened 2025-02-12
[tfe_workspace_run] Allow configuration without `apply` and `destroy` blocks

tmatilai:
## Description

Allow configuring the `tfe_workspace_run` resource without `apply` and `destroy` blocks. This makes it possible to remove the `destroy` block from a resource without an `apply` block, to avoid a destroy run to be scheduled like when removing the whole resource.

This would be especially useful when managing workspaces and workspace runs with automation, allowing to leave the empty resources in state without worrying about accidental destroy runs.

_Remember to:_

- [x] _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)_ [No changes, the documentation doesn't say that at least one block is required] 

## Testing plan

Example configuration, which errors out without this change.

```terraform
  name = "foo"
}

resource "tfe_workspace_run" "test" {
  workspace_id = tfe_workspace.test.id
}
```

[The test setup seems to have changed, disallowing running the test suite using own HCP Terraform organization]

```
$ TESTARGS="-run TestAccTFEWorkspaceRun" make testacc

...
```


tmatilai:
This would be especially useful when changing the execution mode of a workspace from remote/agent to local. For locally executed workspaces the API obviously fails if runs are tried to be created. So the only option now would be to remove the `tfe_workspace_run` resource, but that would trigger a DESTROY run for the workspace, assuming that the `destroy` block was specified.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ ENHANCEMENTS:
 * resource/tfe_saml_settings: Add `private_key_wo` write-only attribute, by @uturunku1 ([#1660](https://github.com/hashicorp/terraform-provider-tfe/pull/1660))
 
 * resource/tfe_ssh_key: Add `key_wo` write-only attribute, by @ctrombley ([#1659](https://github.com/hashicorp/terraform-provider-tfe/pull/1659))
```
tmatilai, line 28: Was the changelog item convention change intentional? At least I find the old one more readable.

ctrombley: This was unintentional, good catch! Much appreciated ✨ 

We'll address this in a separate PR.

PR #1607, opened 2025-02-12
Prep release for v0.64.0

Uk1288:
## Description

Prep release for v0.64.0


Maed223 (approved):
Might want to add the `Unreleased` part on top of `v0.64.0` but shouldn't be a problem to add it when a changelog needs to be added

PR #1634, opened 2025-03-05
Add default project to organization default settings

ctrombley:
## Description

In atlas, we can now update the default project hence we need to add default project id to the organization default settings.

Note:
- I have added default project to the importer as well, is this a needed change?
- Cleaning up resources for a default project is a little tricky because default project cannot be deleted so, while testing I needed to set the default project to be the original default project to be able to delete the resource.

_Remember to:_

## External links

- [go-tfe documentation](https://github.com/hashicorp/go-tfe/pull/1056)
- [API documentation](https://github.com/hashicorp/terraform-docs-common/pull/683)
- [Jira](https://hashicorp.atlassian.net/browse/TF-14884)

## Output from Documentation Preview
Screenshot 2025-02-11 at 12 21 20 PM
Screenshot 2025-02-24 at 1 09 25 PM

Screenshot 2025-03-04 at 5 06 01 PM

### DIFF
--- a/internal/provider/resource_tfe_organization_default_settings.go
+++ b/internal/provider/resource_tfe_organization_default_settings.go
@@ -5,6 +5,7 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 
@@ -280,6 +281,11 @@ func (r *resourceTFEOrganizationDefaultSettings) Read(ctx context.Context, req r
 	// Get organization
 	o, err := r.config.Client.Organizations.Read(ctx, orgName)
 	if err != nil {
+		if errors.Is(err, tfe.ErrResourceNotFound) {
+			resp.Diagnostics.AddError("Organization not found", err.Error())
+			return
+		}
+
 		resp.Diagnostics.AddError("Unable to read organization", err.Error())
 		return
 	}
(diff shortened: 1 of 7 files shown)


internal/provider/resource_tfe_organization_default_settings.go
```diff
@@ -15,159 +10,299 @@ import (
...
+	if !planData.DefaultProjectID.IsNull() {
+		project, err := r.config.Client.Projects.Read(ctx, planData.DefaultProjectID.ValueString())
+		if err != nil {
+			resp.Diagnostics.AddError("Unable to read project", err.Error())
+			return
+		}
+		options.DefaultProject = jsonapi.NewNullableRelationshipWithValue(project)
+	} else if !stateData.DefaultProjectID.IsNull() {
+		// If the project ID is being removed, we need to set it to null
+		options.DefaultProject = jsonapi.NewNullNullableRelationship[*tfe.Project]()
```
netramali, line 224: @ctrombley After thinking about it again, if !stateData.DefaultProjectID.IsNull() and planData.DefaultProjectID.IsNull() --> basically if the user didn't provide default project id but it already exists, we shouldn't remove it. 

An approach we can take is to make the default project id required field. What do you think about that?

The other option is that we only ever do this `options.DefaultProject = jsonapi.NewNullNullableRelationship[*tfe.Project]()` when the plan has an explicit null set for the default project id

ctrombley:
Closing as this work has been deprioritized.

PR #1639, opened 2025-03-10
create a write only mode attribute to set a tfe_variable and tfe_test_variable value

uturunku1:
## Description

Add a write-only attributes to `tfe_variable`, so that a user can pass sensitive data to the resource securely and temporarily without storing it in the Terraform state file or having it show up in the plan.

More info about write-only :

https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments
https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only

_Remember to:_

## Testing plan

1. Apply a configuration that uses a tfe_variable and the new value_wo attribute:

```
variable "session_token" {
  type      = string
  ephemeral = true
}

resource "tfe_variable" "agent_token" {
  key          = "my_agent_token_name"
  value_wo = var.session_token
  sensitive    = false
  category     = "terraform"
  workspace_id = data.tfe_workspace.testing.id
  description  = "a useful description"
}
```

```
    "value": "",
    "value_wo": null,
```

The value that was written into TFC's Workspace variables will show up in the UI because the value of sensitive was set to `false` in this case.

2. Apply the same configuration and pass the same value to the session_token variable. There should be no changes to infrastructure.

3. Now pass a different value to session_token and when you apply, the tfe_variable resource will be recreated.

## External links

```

...
```

### DIFF
--- a/internal/provider/resource_tfe_test_variable.go
+++ b/internal/provider/resource_tfe_test_variable.go
@@ -273,11 +273,11 @@ func (r *resourceTFETestVariable) Create(ctx context.Context, req resource.Creat
 	if !config.ValueWO.IsNull() {
 		// Use the resource's private state to store secure hashes of write-only argument values, the provider during planmodify will use the hash to determine if a write-only argument value has changed in later Terraform runs.
 		hashedValue := generateSHA256Hash(config.ValueWO.ValueString())
-		diags := resp.Private.SetKey(ctx, "value_wo", fmt.Appendf(nil, `"%s"`, hashedValue))
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, fmt.Appendf(nil, `"%s"`, hashedValue))
 		resp.Diagnostics.Append(diags...)
 	} else {
 		// if the value is not configured as write-only, then remove valueWO key from private state. Setting a key with an empty byte slice is interpreted by the framework as a request to remove the key from the ProviderData map.
-		diags := resp.Private.SetKey(ctx, "value_wo", []byte(""))
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, []byte(""))
 		resp.Diagnostics.Append(diags...)
 	}
 
@@ -418,11 +418,11 @@ func (r *resourceTFETestVariable) updatePrivateState(ctx context.Context, resp *
 	if !configValueWO.IsNull() {
 		// Use the resource's private state to store secure hashes of write-only argument values, planModify will use the hash to determine if a write-only argument value has changed in later Terraform runs.
 		hashedValue := generateSHA256Hash(configValueWO.ValueString())
-		diags := resp.Private.SetKey(ctx, "value_wo", fmt.Appendf(nil, `"%s"`, hashedValue))
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, fmt.Appendf(nil, `"%s"`, hashedValue))
 		resp.Diagnostics.Append(diags...)
 	} else {
 		// if value is not configured as write-only, remove valueWO key from private state
-		diags := resp.Private.SetKey(ctx, "value_wo", []byte(""))
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, []byte(""))
 		resp.Diagnostics.Append(diags...)
 	}
 }
--- a/internal/provider/resource_tfe_test_variable.go
+++ b/internal/provider/resource_tfe_test_variable.go
@@ -214,7 +214,7 @@ func (r *resourceTFETestVariable) Schema(ctx context.Context, req resource.Schem
 		Description:         "",
 		MarkdownDescription: "",
 		DeprecationMessage:  "",
-		Version:             0,
+		Version:             1,
 	}
 }
 
--- a/internal/provider/resource_tfe_test_variable.go
+++ b/internal/provider/resource_tfe_test_variable.go
@@ -37,6 +37,7 @@ type modelTFETestVariable struct {
 	ID             types.String `tfsdk:"id"`
 	Key            types.String `tfsdk:"key"`
 	Value          types.String `tfsdk:"value"`
+	ValueWO        types.String `tfsdk:"value_wo"`
 	ReadableValue  types.String `tfsdk:"readable_value"`
 	Category       types.String `tfsdk:"category"`
 	Description    types.String `tfsdk:"description"`
@@ -49,7 +50,7 @@ type modelTFETestVariable struct {
 
 // modelFromTFETestVariable builds a modelTFETestVariable struct from a tfe.TestVariable
 // value (plus the last known value of the variable's `value` attribute).
-func modelFromTFETestVariable(v tfe.Variable, lastValue types.String, moduleID tfe.RegistryModuleID) modelTFETestVariable {
+func modelFromTFETestVariable(v tfe.Variable, lastValue types.String, moduleID tfe.RegistryModuleID, isWriteOnlyValue bool) modelTFETestVariable {
 	// Initialize all fields from the provided API struct
 	m := modelTFETestVariable{
 		ID:             types.StringValue(v.ID),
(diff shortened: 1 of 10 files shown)


internal/provider/resource_tfe_variable.go
```diff
@@ -475,11 +534,28 @@ func (r *resourceTFEVariable) updateWithWorkspace(ctx context.Context, req resou
...
+	updatePrivateState(ctx, resp, config.ValueWO)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
 	diags = resp.State.Set(ctx, &result)
 	resp.Diagnostics.Append(diags...)
 }
 
+func updatePrivateState(ctx context.Context, resp *resource.UpdateResponse, configValueWO types.String) {
```
ctrombley, line 547: Consider making this a receiver method so that the `updatePrivateState` method name can be used in other resources.

internal/provider/resource_tfe_variable.go
```diff
@@ -475,11 +534,28 @@ func (r *resourceTFEVariable) updateWithWorkspace(ctx context.Context, req resou
...
+	if !configValueWO.IsNull() {
+		// Use the resource's private state to store secure hashes of write-only argument values, planModify will use the hash to determine if a write-only argument value has changed in later Terraform runs.
+		diags := resp.Private.SetKey(ctx, "valueWO", fmt.Appendf(nil, `"%s"`, configValueWO.ValueString()))
```
ctrombley, line 550: Here's a simple hashing function we can use:

```
func getSHA256Hash(data string) string {
	hasher := sha256.New()
	hasher.Write([]byte(data))
	return hex.EncodeToString(hasher.Sum(nil))
}
```

internal/provider/resource_tfe_variable.go
```diff
@@ -717,6 +793,58 @@ func (r *resourceTFEVariable) ImportState(ctx context.Context, req resource.Impo
...
+
+func (v *replaceValueWOPlanModifier) Description(ctx context.Context) string {
+	return "The resource will be replaced when the value of value_wo has changed"
+}
+
+func (v *replaceValueWOPlanModifier) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v *replaceValueWOPlanModifier) PlanModifyString(ctx context.Context, request planmodifier.StringRequest, response *planmodifier.StringResponse) {
```
ctrombley, line 806: I *think* this can be replaced by an attribute-focused `RequiresReplaceIf`.

ctrombley: [Here is what is looks like in my PR](https://github.com/hashicorp/terraform-provider-tfe/blob/f5b62f85ef39042a97627837338c1d9b9c6e42f1/internal/provider/resource_tfe_policy_set_parameter.go#L146-L169). The update case seems to be working ok, does it look like I am [missing any test coverage](https://github.com/hashicorp/terraform-provider-tfe/blob/f5b62f85ef39042a97627837338c1d9b9c6e42f1/internal/provider/resource_tfe_policy_set_parameter_test.go#L137-L200) to you?

internal/provider/resource_tfe_variable.go
```diff
@@ -475,11 +537,35 @@ func (r *resourceTFEVariable) updateWithWorkspace(ctx context.Context, req resou
...
+
 	diags = resp.State.Set(ctx, &result)
 	resp.Diagnostics.Append(diags...)
 }
 
+func generateSHA256Hash(data string) string {
+	hasher := sha256.New()
+	hasher.Write([]byte(data))
+	return hex.EncodeToString(hasher.Sum(nil))
+}
```
ctrombley, line 554: We should probably find a shared location for this... maybe `attribute_helpers.go`?

internal/provider/resource_tfe_variable_test.go
```diff
@@ -192,6 +194,38 @@ func TestAccTFEVariable_update_key_sensitive(t *testing.T) {
...
+	variableValue1 := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
+	variableValue2 := variableValue1 + 42
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.SkipBelow(version.Must(version.NewVersion("1.11.0"))),
+		},
+		ProtoV5ProviderFactories: testAccMuxedProviders,
+		CheckDestroy:             testAccCheckTFEVariableDestroy,
+		Steps: []resource.TestStep{
```
ctrombley, line 210: Consider including a test for the `ConflictsWith` validator?

We can also use `resource.TestCheckNoResourceAttr` to assert that there's no value in state.

ctrombley (commented):
Looking really good, nice work! Just a few small questions but it is really close.

internal/provider/resource_tfe_test_variable.go
```diff
@@ -235,7 +268,19 @@ func (r *resourceTFETestVariable) Create(ctx context.Context, req resource.Creat
...
+		hashedValue := generateSHA256Hash(config.ValueWO.ValueString())
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, fmt.Appendf(nil, `"%s"`, hashedValue))
+		resp.Diagnostics.Append(diags...)
+	} else {
+		// if the value is not configured as write-only, then remove valueWO key from private state. Setting a key with an empty byte slice is interpreted by the framework as a request to remove the key from the ProviderData map.
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, []byte(""))
+		resp.Diagnostics.Append(diags...)
+	}
+
 	diags = resp.State.Set(ctx, &result)
```
ctrombley, line 284: I wonder should we return early if there are diag errors before we set state?

internal/provider/resource_tfe_test_variable.go
```diff
@@ -235,7 +268,19 @@ func (r *resourceTFETestVariable) Create(ctx context.Context, req resource.Creat
...
+		// Use the resource's private state to store secure hashes of write-only argument values, the provider during planmodify will use the hash to determine if a write-only argument value has changed in later Terraform runs.
+		hashedValue := generateSHA256Hash(config.ValueWO.ValueString())
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, fmt.Appendf(nil, `"%s"`, hashedValue))
+		resp.Diagnostics.Append(diags...)
+	} else {
+		// if the value is not configured as write-only, then remove valueWO key from private state. Setting a key with an empty byte slice is interpreted by the framework as a request to remove the key from the ProviderData map.
+		diags := resp.Private.SetKey(ctx, ValueWOHashedPrivateKey, []byte(""))
+		resp.Diagnostics.Append(diags...)
+	}
+
```
ctrombley, line 283: Can this make use of r.updatePrivateState?

internal/provider/resource_tfe_test_variable_test.go
```diff
@@ -50,6 +53,40 @@ func TestAccTFETestVariable_basic(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckTFETestVariableExists(
+						"tfe_test_variable.foobar", variable),
+					resource.TestCheckNoResourceAttr(
+						"tfe_test_variable.foobar", "value_wo"),
+					resource.TestCheckResourceAttr(
+						"tfe_test_variable.foobar", "sensitive", "false"),
+				),
+			},
+		},
```
ctrombley, line 86: There's an opportunity here to use a ValueComparer to assert that the resource is being replaced under certain conditions.

[Here's an example](https://github.com/hashicorp/terraform-provider-tfe/pull/1641/files#diff-3cc46359081aee5c828d76dcb1d6a834a9cbf1a5ff32af6e344e82fce6bdc51aR140-R225) from my PR.

ctrombley (dismissed):
Awesome work @uturunku1!

ctrombley (approved):
👍

shwetamurali (approved):

PR #1640, opened 2025-03-11
feat: upgrade tfe_outputs to the provider framework

ctrombley:
This PR upgrades `tfe_outputs` data source to the provider framework, and introduces a new `tfe_outputs` ephemeral resource.

Users can now [create an ephemeral resource](https://developer.hashicorp.com/terraform/plugin/framework/ephemeral-resources) for `tfe_outputs`, so that the sensitive data in the state version outputs is not stored in the Terraform state file and can be handled securely and temporarily during Terraform operations.

Upgrading the `tfe_outputs` data source to the provider framework allows us to remove a direct dependency on the `go-plugin` module (and also an entire muxed provider implementation).

_Remember to:_

### `tfe_outputs` data source
1. Acc tests
1. `tfe_outputs` data source should continue to function as usual

### `tfe_outputs` ephemeral resource
1. Create a terraform configuration that has child module
1. In that module, try out a config similar to this:

```
ephemeral "tfe_outputs" "this" {
   organization   = "my-org-name"
   workspace   = "my-org-name"
}

output "my-org-token" {
  value       = ephemeral.tfe_outputs.this.values.my_workspace_output
  ephemeral   = true
}
```

Then init and apply the configuration.
The state file should not include the ephemeral resource.

```
$ TESTARGS="-run TestAccTFEOutputs" make testacc
--- PASS: TestAccTFEOutputs (12.76s)
=== RUN   TestAccTFEOutputs_ReadAllNonSensitiveValues
--- PASS: TestAccTFEOutputs_ReadAllNonSensitiveValues (11.31s)
--- PASS: TestAccTFEOutputs_emptyOutputs (6.53s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider 31.142s

$ TESTARGS="-run TestAccOutputsEphemeralResource" make testacc
--- PASS: TestAccOutputsEphemeralResource_basic (12.70s)
=== RUN   TestAccOutputsEphemeralResource_readAllNonSensitiveValues
--- PASS: TestAccOutputsEphemeralResource_readAllNonSensitiveValues (12.54s)
=== RUN   TestAccOutputsEphemeralResource_emptyOutputs
--- PASS: TestAccOutputsEphemeralResource_emptyOutputs (6.82s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider 32.567s
```


main.go
```diff
@@ -47,9 +47,8 @@ func main() {
 	//   not proven that out yet.
 	nextProvider := providerserver.NewProtocol5(provider.NewFrameworkProvider())
 	classicProvider := provider.Provider().GRPCProvider
-	lowLevelProvider := provider.PluginProviderServer
 	mux, err := tf5muxserver.NewMuxServer(
-		ctx, nextProvider, classicProvider, lowLevelProvider,
```
brandonc, line 52: 👀 👀 nice

internal/provider/provider_test.go
```diff
@@ -40,7 +40,7 @@ func init() {
 			nextProvider := providerserver.NewProtocol5(NewFrameworkProvider())
 
 			mux, err := tf5muxserver.NewMuxServer(
-				ctx, nextProvider, PluginProviderServer, testAccProvider.GRPCProvider,
```
brandonc, line 43: I had lots of test failures unless `testAccProvider.GRPCProvider` was included in this list

brandonc (commented):
Quick pass. I need to smoke test next

internal/provider/provider_test.go
ctrombley: Thanks, added back in c954bff5.

brandonc (commented):
This is great! A big deal. From my smoke testing, the only thing I noticed was that before the old data source had an optional `organization` because it could also use the organization that is configured on the provider.

ctrombley:
> This is great! A big deal. From my smoke testing, the only thing I noticed was that before the old data source had an optional `organization` because it could also use the organization that is configured on the provider.

That [should still be in place](https://github.com/hashicorp/terraform-provider-tfe/pull/1640/files#diff-d352a56ccdb393bb7cc63294fc2f67f98cc10c161b03de0a44d1ff8591b13a97R114-R119), do you see that it's not working now?

website/docs/ephemeral-resources/outputs.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+```hcl
+ephemeral "tfe_outputs" "foo" {
+  organization = "my-org"
+  workspace = "my-workspace"
+}
+
+resource "random_id" "vpc_id" {
+  keepers = {
+    # Generate a new ID any time the value of 'bar' in workspace 'my-org/my-workspace' changes.
+    bar = ephemeral.tfe_outputs.foo.values.bar
```
brandonc, line 42: Is this a valid example? I got the error `Ephemeral values are not valid for "keepers", because it is not a write-only attribute and must be persisted to state.`

website/docs/ephemeral-resources/outputs.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+
+* `organization` - (Required) The name of the organization.
+* `workspace` - (Required) The name of the workspace.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `values` - The current output values for the specified workspace.
+* `nonsensitive_values` - The current non-sensitive output values for the specified workspace, this is a subset of all output values.
```
brandonc, line 61: Do we need this property on this resource? My hunch is that the restrictions are so high for ephemeral values that it would eclipse sensitive value restrictions. You tell me

ctrombley: Hmm, good question, I can see an argument removing it, but also keeping it around. From a data perspective, the information provided by this resource is lossy if we don't include both attributes - the HCPT-defined Sensitive flag is dropped. I could see that maybe being useful to users in certain situations.

website/docs/ephemeral-resources/outputs.html.markdown
ctrombley: Dang, I'll fix this. Thanks

ctrombley: Updated in af82e4a4.

website/docs/ephemeral-resources/outputs.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+description: |-
+  Get output values from another organization/workspace without writing
+  sensitive data to state.
+---
+
+# Ephemeral: tfe_organization_token
+
+Terraform ephemeral resource for managing a TFE organization token. This
+resource is used to generate a new organization token that is guaranteed not to
+be written to state. Since organization tokens are singleton resources, using this ephemeral resource will replace any existing organization token, including those managed by `tfe_organization_token`.
```
brandonc, line 13: I think this description is copied from another resource...

website/docs/ephemeral-resources/outputs.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `organization` - (Required) The name of the organization.
```
brandonc, line 55:
```suggestion
* `organization` - (Optional) The name of the organization.
```

ctrombley: Addressed in 7d7be8ec.

website/docs/ephemeral-resources/outputs.html.markdown
ctrombley: Addressed in 7d7be8ec.

PR #1641, opened 2025-03-12
[TF-23089] implement write only values for tfe_policy_set_parameter

ctrombley:
## Description

Add a write-only attributes to `tfe_policy_set_parameter`, so that a user can pass sensitive data to the resource securely and temporarily without storing it in the Terraform state file or having it show up in the plan.

_Remember to:_

## Testing plan

1. Apply a configuration that uses a `tfe_policy_set_parameter` and the new `value_wo` attribute:

```
variable "secret" {
  type      = string
  ephemeral = true
}

data "tfe_policy_set" "this" {
  name         = "my-policy-set-name"
  organization = "my-org-name"
}

resource "tfe_policy_set_parameter" "secret" {
  key          = "my_secret"
  value_wo = var.secret
  sensitive    = false
  policy_set_id = data.tfe_policy_set.this.id
}
```

```
    "value": "",
    "value_wo": null,
```

The value that was written into TFC's Workspace variables will show up in the UI because the value of sensitive was set to `false` in this case.

2. Apply the same configuration and pass the same value to the session_token variable. There should be no changes to infrastructure.

3. Now pass a different value to session_token and when you apply, the tfe_policy_set_parameter resource will be recreated.

## External links

- [write-only valuesin the provider framework](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments)
- [Related PR: upgrade tfe_policy_set_parameter to provide framework](https://github.com/hashicorp/terraform-provider-tfe/pull/1637)

```
$ TESTARGS="-run TestAccTFEPolicySetParameter" make testacc

=== RUN   TestAccTFEPolicySetParameter_basic
2025/03/11 18:21:21 [DEBUG] Configuring client for host "tfcdev-781a2fe5.ngrok.io"
2025/03/11 18:21:21 [DEBUG] Service discovery for tfcdev-781a2fe5.ngrok.io at https://tfcdev-781a2fe5.ngrok.io/.well-known/terraform.json
--- PASS: TestAccTFEPolicySetParameter_basic (5.98s)
=== RUN   TestAccTFEPolicySetParameter_update
--- PASS: TestAccTFEPolicySetParameter_update (9.20s)
=== RUN   TestAccTFEPolicySetParameter_import
--- PASS: TestAccTFEPolicySetParameter_import (6.42s)
=== RUN   TestAccTFEPolicySetParameter_valueWO
--- PASS: TestAccTFEPolicySetParameter_valueWO (12.57s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,11 @@ FEATURES:
 
 ENHANCEMENTS:
 
-* resource/tfe_variable: Add `value_wo` write-only attribute ([#1639](https://github.com/hashicorp/terraform-provider-tfe/pull/1639))
+* resource/tfe_variable: Add `value_wo` write-only attribute, by @uturunku ([#1639](https://github.com/hashicorp/terraform-provider-tfe/pull/1639))
 
-* resource/tfe_test_variable: Add `value_wo` write-only attribute ([#1639](https://github.com/hashicorp/terraform-provider-tfe/pull/1639))
+* resource/tfe_test_variable: Add `value_wo` write-only attribute, by @uturunku ([#1639](https://github.com/hashicorp/terraform-provider-tfe/pull/1639))
+
+* resource/tfe_policy_set_parameter: Add `value_wo` write-only attribute, by @ctrombley ([#1641](https://github.com/hashicorp/terraform-provider-tfe/pull/1641))
 
 ## v.0.64.0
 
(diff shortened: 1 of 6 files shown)


brandonc (commented):
Expected to see some new tests for value_wo configurations.

ctrombley:
> Expected to see some new tests for value_wo configurations.

[They're in there](https://github.com/hashicorp/terraform-provider-tfe/pull/1641/files#diff-3cc46359081aee5c828d76dcb1d6a834a9cbf1a5ff32af6e344e82fce6bdc51aR137-R201), did you mean something other than acc tests?

shwetamurali (approved):

PR #1646, opened 2025-03-17
[TF-24656] implement write only values for tfe_organization_run_task

shwetamurali:
## Description

Add a write-only attribute for HMAC key (optional value to verify run task) to `tfe_organization_run_task`, so that a user can pass it in to the resource without seeing it in the diff in plan or storing it in state.

_Remember to:_

## Testing plan

1. Write a terraform configuration with the new key
```
resource "tfe_organization_run_task" "example" {
  organization = "org-name"
  url          = "my-url.com"
  name         = "task-name"
  enabled      = true
  description  = "An example task"
  hmac_key_wo = "example-key"
}
```

2. The state will store an empty string for both the `hmac_key` and the `hmac_key_wo`.
3. A note that it is difficult to test updating hmac key as that would require coordination between this test suite and the external Run Task service to "magically" allow a different Key, so that would require manual testing.

## External links

- [organization_run_task documentation](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/organization_run_task)

```

...
```


internal/provider/resource_tfe_organization_run_task.go
```diff
@@ -146,6 +178,67 @@ func (r *resourceOrgRunTask) Schema(ctx context.Context, req resource.SchemaRequ
...
+		// when `hmac_key_wo` was previously set in the config, but the config switched to either `value` or no value whatsoever
+		response.RequiresReplace = true
+	}
+}
+
+func isWriteOnlyValueInPrivateState(req resource.ReadRequest, resp *resource.ReadResponse) bool {
+	storedValueWO, diags := req.Private.GetKey(ctx, "value_wo")
+	resp.Diagnostics.Append(diags...)
+	return len(storedValueWO) != 0
+}
```
ctrombley, line 240: The linter is complaining since there's another instance of `isWriteOnlyValueInPrivateState` in the `provider` module. Pretty much all of the code is in this module, so it feels a bit like a global namespace.

You have a few options:
- Make this a receiver method on `resourceOrgRunTask`. Probably the quickest, this is what I would recommend.
- Reuse the existing method. This feels a little awkward since the other definition is in another unrelated resource. 
- Check out [my PR](https://github.com/hashicorp/terraform-provider-tfe/pull/1641) and make use of the shared code refactorings I did there. This will be more involved but I'm happy to help you with it. 

Also, no matter what you choose there's a bug here since this method is setting the key `value_wo`, but down below you are reading the key `hmac_key_wo`. They will need to be the same for the hash comparison to work properly.

internal/provider/resource_tfe_organization_run_task.go
```diff
@@ -194,24 +302,48 @@ func (r *resourceOrgRunTask) Create(ctx context.Context, req resource.CreateRequ
...
 
 	// Save data into Terraform state
 	resp.Diagnostics.Append(resp.State.Set(ctx, &result)...)
 }
 
+func generateSHA256Hash(data string) string {
+	hasher := sha256.New()
+	hasher.Write([]byte(data))
+	return hex.EncodeToString(hasher.Sum(nil))
+}
```
ctrombley, line 345: This one is easier. @uturunku1 already moved this method to a shared location, so you can just delete this copy.

internal/provider/resource_tfe_organization_run_task.go
```diff
@@ -251,7 +390,13 @@ func (r *resourceOrgRunTask) Update(ctx context.Context, req resource.UpdateRequ
 		return
 	}
 
-	result := modelFromTFEOrganizationRunTask(task, plan.HMACKey)
+	// isWriteOnlyValue := isWriteOnlyValueInPrivateState(req, resp) // to avoid reading from written-only values
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
ctrombley, line 396: I think this can be deleted?

internal/provider/resource_tfe_organization_run_task_test.go
```diff
@@ -184,6 +187,66 @@ func TestAccTFEOrganizationRunTask_Read(t *testing.T) {
...
+					resource.TestCheckNoResourceAttr("tfe_organization_run_task.foobar", "hmac_key_wo"),
+				),
+				// Ensure that the resource has been replaced
+				ConfigStateChecks: []statecheck.StateCheck{
+					compareValuesDiffer.AddStateValue(
+						"tfe_organization_run_task.foobar", tfjsonpath.New("id"),
+					),
+				},
+			},
+		},
```
ctrombley, line 245: There's one other test step we should add here - it's when the user switched from the write-only value to the regular one.  That's another case where we expect the resource to be recreated. Should be easy enough to reuse the `testAccTFEOrganizationRunTask_basic` config for this one.

ctrombley (changes requested):
This is looking great @shwetamurali! I left a few comments for consideration.

It looks like the linter errors are also responsible for the acceptance test failures.

internal/provider/attribute_helpers.go
```diff
@@ -45,9 +43,3 @@ func (c *ConfiguredClient) dataOrDefaultOrganization(ctx context.Context, data A
 
 	return diags
 }
-
-func generateSHA256Hash(data string) string {
-	hasher := sha256.New()
-	hasher.Write([]byte(data))
-	return hex.EncodeToString(hasher.Sum(nil))
-}
```
ctrombley, line 53: We should leave this in place for now since the `tfe_variable` and `tfe_test_variable` resources still depend on it.

internal/provider/resource_tfe_organization_run_task.go
```diff
@@ -267,7 +393,6 @@ func (r *resourceOrgRunTask) Delete(ctx context.Context, req resource.DeleteRequ
 
 	taskID := state.ID.ValueString()
 
-	tflog.Debug(ctx, fmt.Sprintf("Delete task %s", taskID))
```
ctrombley, line 270: This seems ok to leave in, any reason we removed it here?

ctrombley (changes requested):

ctrombley (dismissed):
🚀

PR #1659, opened 2025-03-21
feat: add write-only `key_wo` attribute to `tfe_ssh_key`

ctrombley:
## Description

This PR migrates `tfe_ssh_key` to the provider framework and adds the write-only attribute `key_wo`, so that a user can pass sensitive private key data to the resource securely and temporarily without storing it in the Terraform state file or having it show up in the plan.

_Remember to:_

## Testing plan

1. Apply a configuration that uses a `tfe_policy_set_parameter` and the new `value_wo` attribute:

```
variable "private_key" {
  type      = string
  ephemeral = true
}

resource "tfe_ssh_key" "this" {
  name            = "my_key"
  organization    = "my_org"
  key_wo          = var.secret
}
```

```
    "key": null,
    "key_wo": null,
```

The value that was written into HCPT will not show up in the UI due to the way the API works.

2. Apply the same configuration and pass the same value to the `private_key` variable. There should be no changes to infrastructure.

3. Now pass a different value to `private_key` and when you apply, the `tfe_ssh_key` resource will be recreated.

## External links

```
$ TESTARGS="-run TestAccTFESSHKey_" make testacc

--- PASS: TestAccTFESSHKey_basic (5.04s)
--- PASS: TestAccTFESSHKey_update (6.13s)
=== RUN   TestAccTFESSHKey_keyWO
--- PASS: TestAccTFESSHKey_keyWO (13.32s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider   25.257s

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ ENHANCEMENTS:
 
 * resource/tfe_saml_settings: Add `private_key_wo` write-only attribute, by @uturunku1 ([#1660](https://github.com/hashicorp/terraform-provider-tfe/pull/1660))
 
+* resource/tfe_ssh_key: Add `key_wo` write-only attribute, by @ctrombley ([#1659](https://github.com/hashicorp/terraform-provider-tfe/pull/1659))
 ## v.0.64.0
 
 FEATURES:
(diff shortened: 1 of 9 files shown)


ctrombley:
> Could you update the docs for datasource? https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/data-sources/ssh_key It says that the `organization` argument is required but it isn't. Your PR changes worked great 👍

Good catch! Addressed in bd87b71d.

PR #1660, opened 2025-03-21
feat: add write-only private_key_wo attribute to tfe_saml_settings

uturunku1:
## Description

This PR adds the write-only attribute `private_ key_wo`, so that a user can pass sensitive private key data to the resource securely and temporarily without storing it in the Terraform state file or having it show up in the plan.

_Remember to:_

## Testing plan

1. Apply a configuration that uses a `tfe_saml_settings` and the new `private_key_wo` attribute:

```
variable "private_key" {
  type      = string
  ephemeral = true
}

resource "tfe_saml_settings" "this" {
  idp_cert         = "foobarCertificate"
  slo_endpoint_url = "https://example.com/slo_endpoint_url"
  sso_endpoint_url = "https://example.com/sso_endpoint_url"
  private_key_wo = var.private_key
 }
```

The state should not include the private_key value:

Screenshot 2025-03-21 at 12 30 31 PM

2. Apply the same configuration and pass the same value to the `private_key_wo` variable. There should be no changes to infrastructure.

3. Now pass a different value to `private_key` and when you apply, the resource will be recreated.

## External links

```
$ TESTARGS="-run TestAccTFESSHKey_" make testacc

--- PASS: TestAccTFESSHKey_basic (5.04s)
--- PASS: TestAccTFESSHKey_update (6.13s)
=== RUN   TestAccTFESSHKey_keyWO
--- PASS: TestAccTFESSHKey_keyWO (13.32s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider   25.257s

```

### DIFF
--- a/internal/provider/resource_tfe_saml_settings.go
+++ b/internal/provider/resource_tfe_saml_settings.go
@@ -94,7 +94,7 @@ func modelFromTFEAdminSAMLSettings(v tfe.AdminSAMLSetting, privateKey types.Stri
 		m.PrivateKey = privateKey
 	}
 
-	// Don't retrieve values if write-only is being used. Unset the hmac key field before updating the state.
+	// Don't retrieve values if write-only is being used. Unset the private key field before updating the state.
 	if isWriteOnly {
 		m.PrivateKey = types.StringValue("")
 	}


ctrombley (approved):
Smoke tested, this looks great!

PR #1664, opened 2025-03-25
migrate tfe_notification_configuration to the provider framework and add write-only token

uturunku1:
## Description

- migrate the tfe_notification_configuration resource to the newer
framework. It uses a schema v0 as there is no difference in schema.
- add write-only for token attribute.

As this is a no-op as far as customer changes go, no changelog is required.

_Remember to:_

## Testing plan from framework migration changes:

- This should be a noop. so all acceptance and unit tests will pass.
- Create a configuration that uses this resource https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/notification_configuration . Confirm that you are able to apply and the functionality has not changed.

## Testing plan from write-only argument changes:

Create a configuration that uses the token_wo attribute. Example:

```
resource "tfe_notification_configuration" "test" {
  name             = "my-test-notification-configuration"
  enabled          = true //optional
  token_wo        = "my-token"
  destination_type = "generic"
  triggers         = ["run:created", "run:planning"] //optional
  url              = "https://httpstat.us/200"
  workspace_id     = data.tfe_workspace.test.id
}
```

Apply the configuration. Unlike the argument `token`, `token_wo` should display `null` as its value in the state.

## External links

[write-only values in the provider framework](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments)
[write-only values for ephemeral resources](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only)

```

...
```

### DIFF
--- a/internal/provider/resource_tfe_notification_configuration.go
+++ b/internal/provider/resource_tfe_notification_configuration.go
@@ -5,6 +5,7 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	tfe "github.com/hashicorp/go-tfe"
@@ -328,11 +329,10 @@ func (r *resourceTFENotificationConfiguration) Create(ctx context.Context, req r
 
 	// Add email_addresses set to the options struct
 	emailAddresses := make([]types.String, len(plan.EmailAddresses.Elements()))
-	if diags := plan.EmailAddresses.ElementsAs(ctx, &emailAddresses, true); diags != nil && diags.HasError() {
-		resp.Diagnostics.Append(diags...)
+	resp.Diagnostics.Append(plan.EmailAddresses.ElementsAs(ctx, &emailAddresses, true)...)
+	if resp.Diagnostics.HasError() {
 		return
 	}
-
 	options.EmailAddresses = []string{}
 	for _, emailAddress := range emailAddresses {
 		options.EmailAddresses = append(options.EmailAddresses, emailAddress.ValueString())
@@ -340,16 +340,18 @@ func (r *resourceTFENotificationConfiguration) Create(ctx context.Context, req r
 
 	// Add email_user_ids set to the options struct
 	emailUserIDs := make([]types.String, len(plan.EmailUserIDs.Elements()))
-	if diags := plan.EmailUserIDs.ElementsAs(ctx, &emailUserIDs, true); diags != nil && diags.HasError() {
-		resp.Diagnostics.Append(diags...)
+	resp.Diagnostics.Append(plan.EmailUserIDs.ElementsAs(ctx, &emailUserIDs, true)...)
+	if resp.Diagnostics.HasError() {
 		return
 	}
+
 	options.EmailUsers = []*tfe.User{}
 	for _, emailUserID := range emailUserIDs {
 		options.EmailUsers = append(options.EmailUsers, &tfe.User{ID: emailUserID.ValueString()})
 	}
 
 	tflog.Debug(ctx, "Creating notification configuration")
+
 	nc, err := r.config.Client.NotificationConfigurations.Create(ctx, workspaceID, options)
 	if err != nil {
 		resp.Diagnostics.AddError("Unable to create notification configuration", err.Error())
@@ -366,8 +368,8 @@ func (r *resourceTFENotificationConfiguration) Create(ctx context.Context, req r
 
 	// We got a notification, so set state to new values
 	result, diags := modelFromTFENotificationConfiguration(nc, !config.TokenWO.IsNull())
-	if diags != nil && diags.HasError() {
-		resp.Diagnostics.Append((diags)...)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
 		return
 	}
 
@@ -395,7 +397,12 @@ func (r *resourceTFENotificationConfiguration) Read(ctx context.Context, req res
 	tflog.Debug(ctx, fmt.Sprintf("Reading notification configuration %q", state.ID.ValueString()))
 	nc, err := r.config.Client.NotificationConfigurations.Read(ctx, state.ID.ValueString())
 	if err != nil {
-		resp.Diagnostics.AddError("Unable to read notification configuration", err.Error())
+		if errors.Is(err, tfe.ErrResourceNotFound) {
+			tflog.Debug(ctx, fmt.Sprintf("`Notification configuration %s no longer exists", state.ID))
+			resp.State.RemoveResource(ctx)
+		} else {
+			resp.Diagnostics.AddError("Error reading notification configuration", "Could not read notification configuration, unexpected error: "+err.Error())
+		}
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


ctrombley (commented):
Before I dig in, I see we need a changelog entry & updated docs.

website/docs/r/notification_configuration.html.markdown
```diff
@@ -124,6 +124,8 @@ The following arguments are supported:
   is `email`.
 * `workspace_id` - (Required) The id of the workspace that owns the notification configuration.
 
+-> **Note:** Write-Only argument `token_wo` is available to use in place of `token`. Write-Only arguments are supported in HashiCorp Terraform 1.11.0 and later. [Learn more](https://developer.hashicorp.com/terraform/language/v1.11.x/resources/ephemeral#write-only-arguments).
```
ctrombley, line 127: I'm not sure if we came to a consensus on this, but I feel (mildly) that this note should either be supplemented with a full treatment for a `value_wo` attribute above, or at the very least a short reference in the description for `value`. Only because I think it might be easy for users to miss if they don't scroll down far enough. 

More importantly, I want to give a consistent treatment across the docs for all these new write-only attributes. I don't feel strongly about which way we decide but there are a few places I did it differently. Thoughts?

internal/provider/resource_tfe_notification_configuration_test.go
```diff
@@ -47,14 +47,48 @@ func TestAccTFENotificationConfiguration_basic(t *testing.T) {
...
+						"tfe_notification_configuration.foobar", "destination_type", "generic"),
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "name", "notification_basic"),
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "triggers.#", "0"),
+					resource.TestCheckResourceAttr(
+						"tfe_notification_configuration.foobar", "url", runTasksURL()),
+					resource.TestCheckNoResourceAttr("tfe_notification_configuration.foobar", "token_wo"),
+				),
+			},
```
ctrombley, line 79: Consider adding test steps and value comparers to make sure the resource is replaced when moving from `value` to `value_wo` and vice-versa.

internal/provider/resource_tfe_notification_configuration_test.go
```diff
@@ -804,6 +850,49 @@ resource "tfe_notification_configuration" "foobar" {
...
+
+resource "tfe_workspace" "foobar" {
+  name         = "workspace-test"
+  organization = tfe_organization.foobar.id
+}
+
+resource "tfe_notification_configuration" "foobar" {
+  name             = "notification_basic"
+  destination_type = "generic"
+  token			   = "some-token"
```
ctrombley, line 868: should this use `token_wo`?

internal/provider/resource_tfe_notification_configuration.go
```diff
@@ -1,88 +1,239 @@
...
-				ConflictsWith: []string{"token", "url"},
+			"email_addresses": schema.SetAttribute{
+				Description: "A list of email addresses. This value must not be provided if `destination_type` is `generic`, `microsoft-teams`, or `slack`.",
+				Optional:    true,
+				Computed:    true,
+				ElementType: types.StringType,
+				Validators: []validator.Set{
+					validators.AttributeValueConflictValidator(
+						"destination_type",
+						[]string{"generic", "microsoft-teams", "slack"},
```
ctrombley, line 182: ❤️

internal/provider/resource_tfe_notification_configuration.go
```diff
@@ -1,88 +1,239 @@
...
-				ConflictsWith: []string{"token", "url"},
+			"email_user_ids": schema.SetAttribute{
+				Description: "A list of user IDs. This value must not be provided if `destination_type` is `generic`, `microsoft-teams`, or `slack`.",
+				Optional:    true,
+				Computed:    true,
+				ElementType: types.StringType,
+				Validators: []validator.Set{
+					validators.AttributeValueConflictValidator(
+						"destination_type",
+						[]string{"generic", "microsoft-teams", "slack"},
```
ctrombley, line 195: ❤️

internal/provider/resource_tfe_notification_configuration.go
```diff
@@ -1,88 +1,239 @@
...
-				Optional:  true,
-				Sensitive: true,
+			"token": schema.StringAttribute{
+				Description: "A write-only secure token for the notification configuration, which can be used by the receiving server to verify request authenticity when configured for notification configurations with a destination type of `generic`. Defaults to `null`. This value _must not_ be provided if `destination_type` is `email`, `microsoft-teams`, or `slack`.",
+				Optional:    true,
+				Sensitive:   true,
+				Validators: []validator.String{
+					validators.AttributeValueConflictValidator(
+						"destination_type",
+						[]string{"email", "microsoft-teams", "slack"},
```
ctrombley, line 213: ❤️

internal/provider/resource_tfe_notification_configuration.go
```diff
@@ -94,333 +245,295 @@ func resourceTFENotificationConfiguration() *schema.Resource {
...
+				Description: "The HTTP or HTTPS URL where notification requests will be made. This value must not be provided if `email_addresses` or `email_user_ids` is present, or if `destination_type` is `email`.",
+				Optional:    true,
+				Validators: []validator.String{
+					validators.AttributeRequiredIfValueString(
+						"destination_type",
+						[]string{"generic", "microsoft-teams", "slack"},
+					),
+					validators.AttributeValueConflictValidator(
+						"destination_type",
+						[]string{"email"},
```
ctrombley, line 263: ❤️

internal/provider/resource_tfe_notification_configuration.go
```diff
@@ -94,333 +245,295 @@ func resourceTFENotificationConfiguration() *schema.Resource {
...
 	}
 
 	// Add triggers set to the options struct
-	for _, trigger := range d.Get("triggers").(*schema.Set).List() {
-		options.Triggers = append(options.Triggers, tfe.NotificationTriggerType(trigger.(string)))
+	var triggers []types.String
+	if diags := plan.Triggers.ElementsAs(ctx, &triggers, true); diags != nil && diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
```
ctrombley, line 323: Definitely no problem here, just another way I have seen this written, choose whichever you like:
```suggestion
	resp.Diagnostics.Append(plan.Triggers.ElementsAs(ctx, &triggers, true)...)
	if resp.Diagnostics.HasError() {
		return
	}
```

internal/provider/resource_tfe_notification_configuration.go
```diff
@@ -94,333 +245,295 @@ func resourceTFENotificationConfiguration() *schema.Resource {
...
-	d.Set("email_addresses", emailAddresses)
 
-	// Update the email user ids
-	var emailUserIDs []interface{}
-	for _, emailUser := range notificationConfiguration.EmailUsers {
-		emailUserIDs = append(emailUserIDs, emailUser.ID)
+	tflog.Debug(ctx, fmt.Sprintf("Reading notification configuration %q", state.ID.ValueString()))
+	nc, err := r.config.Client.NotificationConfigurations.Read(ctx, state.ID.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to read notification configuration", err.Error())
```
ctrombley, line 398: We should check for 404 here and call `resp.RemoveResource()` if it's returned.

ctrombley (changes requested):
This is looking great! 

I left a few comments about areas that might need a second look. I'm also a little confused about the changes around the `Subscribable`/`SubscribableChoice`. Looks like we are setting the workspace ID a bit differently now, and also we are using create options that we weren't using before. Would you mind summing up the intent behind those?

ctrombley (approved):
Smoke tested & looks great, nice work! 

When smoke testing, I noticed that there is a replace required even when simply changing the value of `value_wo`. Is this expected?

PR #1665, opened 2025-03-25
Add write only `token_wo` attribute to `tfe_team_notification_configuration`

shwetamurali:
## Description

This PR adds the write-only attribute `token_wo` to the `tfe_team_notification_configuration` resource, so that a user can pass sensitive private token data to the resource securely and temporarily without storing it in the Terraform state file or having it show up in the plan.

_Remember to:_

## Testing plan

1. Apply a configuration that uses a `tfe_notification_configuration` and the new `token_wo` attribute:

```
variable "secret_token" {
  type      = string
  ephemeral = true
}

  organization = tfe_organization.test.id
}

resource "tfe_team_notification_configuration" "this" {
  name             = "my-test-notification-configuration""
  destination_type = "generic"
  url 			   = "https://example.com"
  token_wo         = var.secret_token
  workspace_id          = tfe_workspace.test.id
}
```

```
    "token": null,
    "token_wo": null,
```

The value that was written into HCPT will not show up in the UI due to the way the API works.

2. Apply the same configuration and pass the same value to the `secret_token` variable. There should be no changes to infrastructure.

3. Now pass a different value to `secret_token` and when you apply, the `tfe_notification_configuration` resource will be recreated.

## External links

```
$ TESTARGS="-run TestAccTFETeamNotificationConfiguration_tokenWO" make testacc

--- PASS: TestAccTFETeamNotificationConfiguration_tokenWO (6.88s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/provider	7.525s

```


internal/provider/resource_tfe_team_notification_configuration.go
```diff
@@ -323,7 +358,18 @@ func (r *resourceTFETeamNotificationConfiguration) Create(ctx context.Context, r
 		tnc.Token = plan.Token.ValueString()
 	}
 
-	result, diags := modelFromTFETeamNotificationConfiguration(tnc)
+	result, diags := modelFromTFETeamNotificationConfiguration(tnc, isWriteOnly, plan.Token)
+	if diags != nil {
```
ctrombley, line 362: At the top of the definition of `modelFromTFETeamNotificationConfiguration`, you can instantiate a `diags` so that you can return `diag.Diagnostics` and not `*diag.Diagnostics`:

```
var diags diag.Diagnostics
```

And then here (and after other similar calls), we can check for errors the same way we do everywhere else:
```suggestion
	if diags.HasError() {
```

website/docs/r/team_notification_configuration.html.markdown
```diff
@@ -126,6 +126,7 @@ The following arguments are supported:
   be used by the receiving server to verify request authenticity when configured for notification
   configurations with a destination type of `generic`. Defaults to `null`.
   This value _must not_ be provided if `destination_type` is `email`, `microsoft-teams`, or `slack`.
+- `TokenWO` - (Optional) A write-only secure token for the notification configuration, guaranteed not to be written to plan or state artifacts. Only one of `token` or `token_wo` can be provided.
```
ctrombley, line 129: Small change needed here: 

```suggestion
- `token_wo` - (Optional) A write-only secure token for the notification configuration, guaranteed not to be written to plan or state artifacts. Only one of `token` or `token_wo` can be provided.
```

internal/provider/resource_tfe_team_notification_configuration.go
```diff
@@ -323,7 +358,18 @@ func (r *resourceTFETeamNotificationConfiguration) Create(ctx context.Context, r
 		tnc.Token = plan.Token.ValueString()
 	}
 
-	result, diags := modelFromTFETeamNotificationConfiguration(tnc)
+	result, diags := modelFromTFETeamNotificationConfiguration(tnc, isWriteOnly, plan.Token)
+	if diags.HasError() {
```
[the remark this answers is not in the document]
ctrombley: We can also keep this as is, but change the signature of `modelFromTFETeamNotificationConfiguration` to:

```
func modelFromTFETeamNotificationConfiguration(v *tfe.NotificationConfiguration, isWriteOnly bool, lastValue types.String) (*modelTFETeamNotificationConfiguration, diag.Diagnostics) {
  var diags diag.Diagnostics
...
```

ctrombley (approved):
🚀

PR #1701, opened 2025-04-19
Use the correct form of `sha256.Sum256`.

mcsaucy:
## Description

`sha256.New().Sum(foo)` appends the SHA-256 hash for nil to foo. See https://go.dev/play/p/vSW0U3Hq4qk

I think was only used for comparisons, which meant it worked just fine but your comparisons were less efficient and the SHA256 did nothing for you.

I'm making this change as part of a cleanup of invalid hash.Hash uses across Github.

## Testing plan

Acceptance tests in the CI should cover it.

I spent a while trying to work out how to set up a test environment and gave up after it took substantially longer than the actual fixing of the bug. Seems like I'd need to set up a Terraform instance. If this is insufficient, lemme know and I'll close the PR and flip this to an issue for someone else to take.

### DIFF
--- a/internal/client/config.go
+++ b/internal/client/config.go
@@ -52,7 +52,8 @@ type ClientConfiguration struct {
 
 // Key returns a string that is comparable to other ClientConfiguration values
 func (c ClientConfiguration) Key() string {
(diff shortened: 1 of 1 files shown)


sebasslash:
Thank you for this quick fix :heart:. I've gone ahead and rebased your changes onto a local branch so it can run against CI, see #1703

Issue #1700, opened 2025-04-19
tfe_team_token token value is changed to "" in state - Version 0.65.1


twallac10:
#### Terraform version

```plaintext
Terraform v1.9.8
```

#### Terraform Configuration Files

```terraform
resource "tfe_team_token" "example" {
  team_id = data.tfe_team.example.id
  }

data "tfe_team" "example" {
  name = "owners"
  organization = "Exampleorg"
}
```

#### Debug Output

```plaintext
2025-04-19T07:55:21.058-0500 [TRACE] DiffTransformer: found NoOp change for tfe_team_token.example
```

#### Expected Behavior

 
Any runs using `0.65.1` should maintain the token value in the state file for the standard resource type.

#### Actual Behavior

The token attribute in the state file gets changed to: `"token": ""` after running an apply on version `0.65.1`.

#### Additional Context

Downgrading to version `0.64.0` has the expected behavior of the token remaining in the state file for the resource.

New resources created using `0.65.1` will populate the token in the state file. 

Any subsequent apply that occurs on the resource using `0.65.0` or `0.65.1` will remove the token from the state file.

reed-hanger:
I looked at updating to 0.65.1 of the tfe provider using 1.11.4 of Terraform and saw this planned change. I reverted the tfe provider to 0.64.0, and the planned state changes went away:

```
Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the
last "terraform apply" which may have affected this plan:

  # tfe_team_token.this[0] has changed
  ~ resource "tfe_team_token" "team_tokens" {
        id      = "team-aBcDeFg123456789"
      - token   = (sensitive value) -> null
        # (1 unchanged attribute hidden)
    }
```

ctrombley:
Fixed in [v0.65.2](https://github.com/hashicorp/terraform-provider-tfe/releases/tag/v0.65.2).

Issue #1696, opened 2025-04-18
url attribute validation doesn't work as expected for tfe_notification_configuration - Version 0.65.0


cclaes-dsi:
#### Terraform version

```plaintext
Terraform v1.11.4
```

#### Terraform Configuration Files

```terraform
variable "url" {
  default = "http://example.com"
}

resource "tfe_notification_configuration" "this" {
  destination_type = "slack"
  enabled          = true
  name             = "Slack"
  triggers         = []
  url              = var.url
  workspace_id     = tfe_workspace.this.id
}
```

#### Debug Output

```plaintext
│ Error: Missing Required Attribute
│ 
│   with module.mod.tfe_notification_configuration.this,
│   on .terraform/modules/mod/main.tf line 143, in resource "tfe_notification_configuration" "this":
│  143:   url              = var.url
│ 
│ The attribute 'url' is required when 'destination_type' is 'slack'
```

#### Expected Behavior

The notification configuration should be successfully validated.

#### Actual Behavior

Validation fails with the following message: `The attribute 'url' is required when 'destination_type' is 'slack'`.

#### Additional Context

Downgrading to `0.64.0` or not using a variable, works.

0.64.0 output:
```plaintext
Success! The configuration is valid.
```

tyu0912:
Seeing the same.

ctrombley:
Resolved by #1705.

PR #1706, opened 2025-04-21
Fix several regressions in v0.65.x

ctrombley:
## Description

The recent releases of v0.65.0 and v0.65.1 contain several regressions:

- `tfe_team_token` clears token from state after upgrade (#1700)
-  `tfe_policy_set_parameter` errors attempting to unnecessarily upgrade schema version
- `url` attribute validation doesn't work as expected for `tfe_notification_configuration` (#1696)

This PR provides resolution for the first two issues, and #1705 resolves the third.

##### Test config
```
resource "tfe_policy_set" "test" {
  name         = "my-policy-set-name"
}

resource "tfe_policy_set_parameter" "test" {
  key          = "my_key_name"
  value        = "my_value_name"
  policy_set_id = tfe_policy_set.test.id
}
```
1. Use v0.64 to create a policy set param with the config above
1. Re-apply the config with the provider version in this PR
1. Ensure no errors are thrown relating to the schema version

##### Test config
```
resource "tfe_team" "test" {
  name         = "test-team"
}

resource "tfe_team_token" "test" {
  team_id = tfe_team.test.id
}
```
1. Use v0.64 to create a team token with the config above
1. Re-apply the config with the provider version in this PR
1. Check the state file and ensure the token string is still visible

### DIFF
--- a/internal/provider/resource_tfe_policy_set_parameter.go
+++ b/internal/provider/resource_tfe_policy_set_parameter.go
@@ -97,7 +97,7 @@ func (r *resourceTFEPolicySetParameter) Metadata(_ context.Context, req resource
 func (r *resourceTFEPolicySetParameter) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
 	resp.Schema = schema.Schema{
 		Description: "Creates, updates and destroys policy set parameters.",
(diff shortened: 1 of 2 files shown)


Maed223 (approved):
Smoke tested, looking good 👍

PR #1708, opened 2025-04-22
chore: onboard repo to Common Release Tooling

ctrombley:
## Description
This PR onboards the TFE provider to use Common Release Tooling.


.github/workflows/build.yml
```diff
@@ -0,0 +1,188 @@
...
+          reproducible: report
+          # TODO: Customize `go build` flags for your provider. Compare to existing .goreleaser.yml.
+          # TODO: Customize `ldflags` for your provider. Ensure that the import path and name in the -X flag
+          # are correct for your provider.
+          instructions: |
+            go build \
+              -o "$BIN_PATH" \
+              -trimpath \
+              -buildvcs=false \
+              -ldflags "-s -w -X 'main.version=${{ needs.set-product-version.outputs.product-version }}'"
```
bbasata, line 168: I see that the provider declares a `version.ProviderVersion` variable here:

https://github.com/hashicorp/terraform-provider-tfe/blob/608dfddea25595430bfefe73e743974a8e3b1fb8/version/version.go#L9

If you'd like, you can set that at build-time with the following change. Curiously, I do not see this in [this provider's GoReleaser build command](https://github.com/hashicorp/terraform-provider-tfe/blob/608dfddea25595430bfefe73e743974a8e3b1fb8/.goreleaser.yml#L39), so I wonder if `version.ProductVersion` is being set at all in current releases.

```suggestion
              -ldflags "-s -w -X 'version.ProductVersion=${{ needs.set-product-version.outputs.product-version }}'"
```

ctrombley: Thanks for noticing this! Addressed in 0f08ecea. I'm not sure that this is still being used anywhere but I'd rather this PR be consistent with the old code.

PR #1776, opened 2025-07-12
Fix `tfe_agent_pool_allowed_workspaces` example using resources that …aren't defined

AleksaC:
## Description

`tfe_workspace.test` is not defined, should use `tfe_workspace.test-workspace` instead


sebasslash:
Thank you for this contribution! I've gone ahead and rebased the changes onto a local branch in #1781

PR #1835, opened 2025-09-08
[TF-28672] Add Vault, AWS, GCP, and Azure OIDC configuration resources

helenjw:
## Description

Add the following resources to manage OIDC configurations:

- `tfe_vault_oidc_configuration`
- `tfe_aws_oidc_configuration`
- `tfe_gcp_oidc_configuration`
- `tfe_azure_oidc_configuration`

_Remember to:_

## Testing plan

1.  Using a version of [go-tfe with HYOK support](https://github.com/hashicorp/go-tfe/pull/1201) and using an organization with access to HYOK
2. Use the following terraform configuration to add OIDC resources via `terraform plan` and `terraform apply`
3. Try to update these values and apply them (updating attributes should not destroy and replace the resources)
4. Run `terraform destroy` on these resources to make sure they can be deleted

```terraform
resource "tfe_aws_oidc_configuration" "aws_oidc_tfe_provider" {
  role_arn = "arn:aws:iam::111111111111:role/example-role-arn"
}

resource "tfe_gcp_oidc_configuration" "gcp_oidc_tfe_provider" {
  service_account_email     = "myemail@gmail.com"
  project_number            = "11111111"
  workload_provider_name    = "projects/1/locations/global/workloadIdentityPools/1/providers/1"
}

resource "tfe_azure_oidc_configuration" "azure_oidc_tfe_provider" {
  client_id         = "application-id1"
  subscription_id   = "subscription-id1"
  tenant_id         = "tenant-id1"
}

resource "tfe_vault_oidc_configuration" "vault_oidc_tfe_provider" {
  address           = "https://my-vault-cluster-public-vault-659decf3.b8298d98.z1.hashicorp.cloud:8200"
  role_name         = "vault-role-name"
  namespace         = "admin"
  auth_path         = "jwt-auth-path"
}
```

## External links

- API documentation:
   - [Vault](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/hold-your-own-key/oidc-configurations/vault)
   - [AWS](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/hold-your-own-key/oidc-configurations/aws)
   - [Azure](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/hold-your-own-key/oidc-configurations/azure)
   - [GCP](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/hold-your-own-key/oidc-configurations/gcp)
- [JIRA](https://hashicorp.atlassian.net/browse/TF-28672)

```shell
TESTARGS="-run TestAccTFEVaultOIDCConfiguration" envchain local make testacc     14:58:34   5.672s
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEVaultOIDCConfiguration -timeout 15m
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/client	0.340s [no tests to run]
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/logging	0.167s [no tests to run]
=== RUN   TestAccTFEVaultOIDCConfiguration_basic
--- PASS: TestAccTFEVaultOIDCConfiguration_basic (3.01s)
PASS
```

```shell
TESTARGS="-run TestAccTFEAWSOIDCConfiguration" envchain local make testacc       15:00:37   5.392s
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEAWSOIDCConfiguration -timeout 15m
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/client	(cached) [no tests to run]
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/logging	(cached) [no tests to run]
=== RUN   TestAccTFEAWSOIDCConfiguration_basic
--- PASS: TestAccTFEAWSOIDCConfiguration_basic (3.28s)
PASS
```

```shell
TESTARGS="-run TestAccTFEGCPOIDCConfiguration" envchain local make testacc       15:07:26   1.089s
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEGCPOIDCConfiguration -timeout 15m
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/client	0.298s [no tests to run]
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/logging	0.416s [no tests to run]
=== RUN   TestAccTFEGCPOIDCConfiguration_basic
--- PASS: TestAccTFEGCPOIDCConfiguration_basic (3.16s)
PASS
```

```shell
TESTARGS="-run TestAccTFEAzureOIDCConfiguration" envchain local make testacc     15:07:52   5.436s
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run TestAccTFEAzureOIDCConfiguration -timeout 15m
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/client	0.334s [no tests to run]
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/logging	0.166s [no tests to run]
=== RUN   TestAccTFEAzureOIDCConfiguration_basic
--- PASS: TestAccTFEAzureOIDCConfiguration_basic (2.82s)
PASS
```

### DIFF
--- a/internal/provider/resource_tfe_aws_oidc_configuration_test.go
+++ b/internal/provider/resource_tfe_aws_oidc_configuration_test.go
@@ -13,6 +13,9 @@ func TestAccTFEAWSOIDCConfiguration_basic(t *testing.T) {
 	skipUnlessHYOKEnabled(t)
 
 	orgName := os.Getenv("HYOK_ORGANIZATION_NAME")
+	if orgName == "" {
+		t.Skip("Skipping test. Set HYOK_ORGANIZATION_NAME environment to enable test.")
+	}
 
 	originalRoleARN := "arn:aws:iam::123456789012:role/terraform-provider-tfe-example-1"
 	newRoleARN := "arn:aws:iam::123456789012:role/terraform-provider-tfe-example-2"
--- a/internal/provider/resource_tfe_azure_oidc_configuration_test.go
+++ b/internal/provider/resource_tfe_azure_oidc_configuration_test.go
@@ -13,6 +13,9 @@ func TestAccTFEAzureOIDCConfiguration_basic(t *testing.T) {
 	skipUnlessHYOKEnabled(t)
 
 	orgName := os.Getenv("HYOK_ORGANIZATION_NAME")
+	if orgName == "" {
+		t.Skip("Skipping test. Set HYOK_ORGANIZATION_NAME environment to enable test.")
+	}
 
 	originalClientID := "client-id-1"
 	updatedClientID := "client-id-2"
--- a/internal/provider/resource_tfe_gcp_oidc_configuration_test.go
+++ b/internal/provider/resource_tfe_gcp_oidc_configuration_test.go
@@ -13,6 +13,9 @@ func TestAccTFEGCPOIDCConfiguration_basic(t *testing.T) {
 	skipUnlessHYOKEnabled(t)
 
 	orgName := os.Getenv("HYOK_ORGANIZATION_NAME")
+	if orgName == "" {
+		t.Skip("Skipping test. Set HYOK_ORGANIZATION_NAME environment to enable test.")
+	}
 
 	originalServiceAccountEmail := "service-account@example.iam.gserviceaccount.com"
 	updatedServiceAccountEmail := "updated-service-account@example.iam.gserviceaccount.com"
--- a/internal/provider/resource_tfe_vault_oidc_configuration_test.go
+++ b/internal/provider/resource_tfe_vault_oidc_configuration_test.go
@@ -13,6 +13,9 @@ func TestAccTFEVaultOIDCConfiguration_basic(t *testing.T) {
 	skipUnlessHYOKEnabled(t)
 
 	orgName := os.Getenv("HYOK_ORGANIZATION_NAME")
+	if orgName == "" {
+		t.Skip("Skipping test. Set HYOK_ORGANIZATION_NAME environment to enable test.")
+	}
 
 	originalAddress := "https://vault.example.com"
 	updatedAddress := "https://vault.example2.com"


website/docs/r/aws_oidc_configuration.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+page_title: "Terraform Enterprise: tfe_aws_oidc_configuration"
+description: |-
+  Manages AWS OIDC configurations.
+---
+
+# tfe_aws_oidc_configuration
+
+Defines an AWS OIDC configuration resource.
+
+~> **NOTE:** Hold your own key is available on the HCP Terraform Premium edition. Refer to [HCP Terraform pricing](https://www.hashicorp.com/en/pricing?product_intent=terraform&tab=terraform) for details.
```
helenjw: Not sure if I should be directly referencing HYOK here, or just say something like:
```html
OIDC configuration management is available on the HCP Terraform Premium edition. Refer to HCP Terraform pricing for details.
```

internal/provider/resource_tfe_aws_oidc_configuration.go
```diff
@@ -0,0 +1,203 @@
...
+	var state modelTFEAWSOIDCConfiguration
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	oidcID := state.ID.ValueString()
+	tflog.Debug(ctx, fmt.Sprintf("Read AWS OIDC configuration: %s", oidcID))
+	oidc, err := r.config.Client.AWSOIDCConfigurations.Read(ctx, state.ID.ValueString())
```
dominic-retli-hashi: Nit: 
you can use `oidcID` in line 132 instead of re-reading `state.ID.ValueString()` since it was already assigned on line 130.

helenjw: Oh true! Thanks! I addressed in cbe25067332ee3759f771af3718f46076b80773a

internal/provider/resource_tfe_aws_oidc_configuration.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	options := tfe.AWSOIDCConfigurationUpdateOptions{
+		RoleARN: plan.RoleARN.ValueString(),
+	}
+
+	oidcID := state.ID.ValueString()
+	tflog.Debug(ctx, fmt.Sprintf("Update TFE AWS OIDC Configuration %s", oidcID))
+	oidc, err := r.config.Client.AWSOIDCConfigurations.Update(ctx, oidcID, options)
+	if err != nil {
```
iuri-slywitch-hashicorp, line 167: maybe we could have the same if-statement for the error here as in the read? adding a clause that we show to the user that the resource has not been found?

iuri-slywitch-hashicorp: I also wonder if we should be doing any tests for create and read. Other than that, looks very good, nice!

helenjw: > maybe we could have the same if-statement for the error here as in the read? adding a clause that we show to the user that the resource has not been found?

I think there's been a misunderstanding. In the `Update` implementation, when the error is handled by:
```go
if err != nil {
   resp.Diagnostics.AddError("Error updating TFE AWS OIDC Configuration", err.Error())
   return
}
```
We are actually exposing the error in the response back to the user. It does look different on the `Update` method compared to the `Read`. From what I've seen in other `Read` resource implementations (eg. `internal/provider/resource_tfe_variable.go`), if the resource is not found, it's not an error. A lot of other implementations will log this as a debug message and call `resp.State.RemoveResource(ctx)`. However, it passes along any other errors to the user via `resp.Diagnostics.AddError`.

For `Update`, it is commonly handled differently where all the errors (including `404`s) are passed along to the user. I suppose this may be because we want to let the user know they might have given us a bad `ID` that they should correct.

TLDR; not found errors on update will find its way to the user

helenjw: > I also wonder if we should be doing any tests for create and read. 

The basic test case will perform tests on create, read, update, delete, and also importing the resource

dominic-retli-hashi (approved):

internal/provider/resource_tfe_aws_oidc_configuration_test.go
```diff
@@ -13,6 +13,9 @@ func TestAccTFEAWSOIDCConfiguration_basic(t *testing.T) {
 	skipUnlessHYOKEnabled(t)
 
 	orgName := os.Getenv("HYOK_ORGANIZATION_NAME")
+	if orgName == "" {
```
iuri-slywitch-hashicorp: maybe put this function inside helper_test.go file as another if-statement?

helenjw: I thought about including this in the `skipUnlessHYOKEnabled` function there, but decided against it. Now that function only checks if `ENABLE_HYOK == 1`. The reason I opted to do this is because not every test requires the HYOK org to be set, or in other cases require other HYOK env variables to be set. I think it would be much more helpful to notify users explicitly which env variables are missing when running each test.

iuri-slywitch-hashicorp: makes sense! nit: maybe create another helper function for this case then? i'll leave it up to you.

iuri-slywitch-hashicorp: actually... we're not doing this in go-fe. so maybe we shouldn't do it here as well....

helenjw: thanks for the reminder! It's been removed

PR #1850, opened 2025-09-19
[IPL-6481] Ignore non-service account member ids for tfe_team_organization_memberships resource

smishra96:
## Description

This PR address a bug where service account users where visible while using team tokens. `tfe_team_organization_memberships` resource was modified to ignore non-service account member ids during operations. 

**Note:** If an user has a service-account organization membership id in the `tfe_team_organization_memberships` memberships list, they would keep on seeing drifts even after a successful apply as the resource would not be managing a sa membership id.

## Testing plan

1.  Use a team token to create organization memberships using `tfe_team_organization_memberships` resource
2. There should be no Drift detected while doing so

## Example

Bug: A drift appears when doing terraform plan using team token in the list of membership ids in `tfe_team_organization_memberships` resource because the service account members are not being ignored.

https://github.com/user-attachments/assets/e3007aad-458d-428e-875f-d8d1a8b99139

### DIFF
--- a/internal/provider/resource_tfe_team_organization_members.go
+++ b/internal/provider/resource_tfe_team_organization_members.go
@@ -86,18 +86,15 @@ func resourceTFETeamOrganizationMembersRead(d *schema.ResourceData, meta interfa
 	}
 
 	log.Printf("[DEBUG] Read users from team: %s", d.Id())
-	nonServiceAccountUsers, err := fetchNonServiceAccountUserIds(config.Client, d.Id())
+	nonServiceAccountOrganizationMemberships, err := filterNonServiceAccountOrganizationMembers(config.Client, organizationMemberships)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Maed223:
Hey @smishra96, I think we could better solve this issue through smarter use of the API rather than the added filtering for non-service accounts. 

Currently the resource uses `go-tfe`'s `TeamMembers.ListOrganizationMemberships` where we just return the result of the include of organization memberships on a Team. A better route here would be to use `OrganizationMemberships.List`, where the endpoint automatically filters out service accounts so we can avoid the overhead of doing this in the provider. The one catch is that `go-tfe` currently doesn't implement the `filter[team]=team_external_id` query param that would be needed and is otherwise present on the OrganizationMemberships controller. I think a better route for a solution here would be to update go-tfe's `OrganizationMembershipListOptions` to have a `Team` filter, and to then make use of it here in the provider.

PR #1855, opened 2025-09-23
Stacks GA update

Maed223:
## Description

Bringing in the upstream changes from go-tfe that align things with the state of what's being serialized in the Stacks API.

Many of the upstream changes that are relevant here are simple route changes for the GA endpoints.  Otherwise the only thing lost here are the `deployment_names` computed attribute on the stack. This was seemingly only being used within a guard against what is otherwise an API error internally. Externally this is probably better suited to be determined through Stacks config.

Stack Tests:
Screenshot 2025-09-23 at 3 30 40 PM
Screenshot 2025-09-23 at 3 31 18 PM
Screenshot 2025-09-23 at 3 31 31 PM

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -73,8 +73,6 @@ github.com/hashicorp/go-retryablehttp v0.7.8 h1:ylXZWnqa7Lhqpk0L1P1LzDtGcCR0rPVU
 github.com/hashicorp/go-retryablehttp v0.7.8/go.mod h1:rjiScheydd+CxvumBsIrFKlx3iS0jrZ7LvzFGFmuKbw=
 github.com/hashicorp/go-slug v0.16.7 h1:sBW8y1sX+JKOZKu9a+DQZuWDVaX+U9KFnk6+VDQvKcw=
 github.com/hashicorp/go-slug v0.16.7/go.mod h1:X5fm++dL59cDOX8j48CqHr4KARTQau7isGh0ZVxJB5I=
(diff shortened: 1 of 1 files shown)


go.sum
```diff
@@ -74,6 +74,8 @@ github.com/hashicorp/go-slug v0.16.7 h1:sBW8y1sX+JKOZKu9a+DQZuWDVaX+U9KFnk6+VDQv
 github.com/hashicorp/go-slug v0.16.7/go.mod h1:X5fm++dL59cDOX8j48CqHr4KARTQau7isGh0ZVxJB5I=
 github.com/hashicorp/go-tfe v1.92.0 h1:DxJ/ONlNgp4kP5xdglJBLYBeunDeii3CelpgA6PjebM=
 github.com/hashicorp/go-tfe v1.92.0/go.mod h1:QwqgCD5seztgp76CP7F0POJPflQNSqjIvBpVohg9X50=
+github.com/hashicorp/go-tfe v1.93.0 h1:hJubwn1xNCo1iBO66iQkjyC+skR61cK1AQUj4O9vvuI=
```
sebasslash, line 76: Don't forget to run `go mod tidy` after a dependency upgrade 😄

markdecrane: 🤦 https://github.com/hashicorp/terraform-provider-tfe/pull/1855/commits/76472b75416a5b665e237c456145ded154a2eec8

PR #1890, opened 2025-10-28
Warning for invalid authentication on Cloud

christian-doucette:
## Description

-Using CLI or Config file authentication for the TFE provider is valid on local but not cloud
-This causes a confusing error message due to the TFE provider trying to use the autogenerated single-use run token which is often not authorized for what its trying to do
-The new warning should make it more clear for users to resolve

_Remember to:_

- [x] _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)_ (no update to documentation needed, the documentation already says that you can't use credentials/config files when on Cloud so this warning is already in line with the current docs: https://registry.terraform.io/providers/hashicorp/tfe/latest/docs#authentication

## Testing plan

1. Create a workspace in Terraform cloud.
2. Ensure that TFE_TOKEN is not set in the workspace (not really a step but this is important for reproducing the result)
3. Run terraform code on that workspace to create a different workspace with the TFE provider. Sample: 
```
terraform { 
  cloud { 
    hostname = "my-hostname.com" 
    organization = "hashicorp" 

    workspaces { 
      name = "workspace-already-created-in-step-one" 
    } 
  } 

  required_providers {
    tfe = {
        version = "0.73.0"
    }
  }
}

provider "tfe" {
    hostname = "myhostname.com"
    //token must also not be set here
}

resource "tfe_workspace" "example" {
  name          = "new-workspace-to-be-created"
  organization  = "hashicorp"
}
```

4. Ensure that the warning about using config/credential authentication on cloud is received.  
 
```
% TESTARGS="-run Test_GetClient -run TestClient_sendAuthenticationWarning -run TestConfig_configureSetsTokenSource -run TestConfigureEnvOnCloudUsingConfigFiles" make testacc
TF_ACC=1 TF_LOG_SDK_PROTO=OFF go test $(go list ./... |grep -v 'vendor') -v -run Test_GetClient -run TestClient_sendAuthenticationWarning -run TestConfig_configureSetsTokenSource -run TestConfigureEnvOnCloudUsingConfigFiles -timeout 15m
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/client	0.203s [no tests to run]
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/logging	0.298s [no tests to run]
=== RUN   TestConfigureEnvOnCloudUsingConfigFiles
2025/12/19 16:47:22 [DEBUG] Configuring client for host "app.terraform.io"
2025/12/19 16:47:22 [DEBUG] Attempting to fetch token from Terraform CLI configuration for hostname "app.terraform.io"...
2025/12/19 16:47:22 [DEBUG] Service discovery for app.terraform.io at https://app.terraform.io/.well-known/terraform.json
    provider_test.go:314: Want to see if this shows up in Datadog flaky test
--- PASS: TestConfigureEnvOnCloudUsingConfigFiles (0.16s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/provider	0.527s
?   	github.com/hashicorp/terraform-provider-tfe/internal/provider/helpers	[no test files]
?   	github.com/hashicorp/terraform-provider-tfe/internal/provider/planmodifiers	[no test files]
?   	github.com/hashicorp/terraform-provider-tfe/internal/provider/validators	[no test files]

```

None

### DIFF
--- a/internal/provider/provider_test.go
+++ b/internal/provider/provider_test.go
@@ -271,15 +271,9 @@ func TestConfigureEnvOrganization(t *testing.T) {
 func TestConfigureEnvOnCloudUsingConfigFiles(t *testing.T) {
 	// tests that the provider sends a warning when running on cloud (checked using TFE_AGENT_VERSION)
 	// and using a token from configuration files
-	envToken := os.Getenv("TFE_TOKEN")
 
-	reset := func() {
-		os.Setenv("TFE_TOKEN", envToken)
-	}
-	defer reset()
-
-	// temporarily removes TFE_TOKEN so token will be from configuration files
-	os.Unsetenv("TFE_TOKEN")
+	// removes TFE_TOKEN for the test so token will be from configuration files
+	t.Setenv("TFE_TOKEN", "")
 	t.Setenv("TFC_AGENT_VERSION", "1.0")
 	t.Setenv("TFE_HOSTNAME", "app.terraform.io")
 	t.Setenv("TF_CLI_CONFIG_FILE", "test-fixtures/cli-config-files/terraformrc")


internal/client/config.go
```diff
@@ -165,9 +165,17 @@ func credentialsSource(credentials CredentialsMap) auth.CredentialsSource {
 	return creds
 }
 
+// Using presence of TFC_AGENT_VERSION to determine if this provider is running on HCP Terraform / enterprise
+func providerRunningInCloud() bool {
+	_, present := os.LookupEnv("TFC_AGENT_VERSION")
+	return present
+}
```
christian-doucette: Might be a stupid edge case, but if someone sets this variable themselves locally it will run as if on cloud. Shouldn't be a problem in this case because there are no security implications to having the additional warning

internal/client/config.go
```diff
@@ -165,9 +165,17 @@ func credentialsSource(credentials CredentialsMap) auth.CredentialsSource {
...
+
 // configure accepts the provider-level configuration values and creates a
 // clientConfiguration using fallback values from the environment or CLI configuration.
-func configure(tfeHost, token string, insecure bool) (*ClientConfiguration, error) {
+func configure(tfeHost, token string, insecure bool) (*ClientConfiguration, bool, error) {
```
christian-doucette: a lot of the lines of the PR are just for adding this new sendCredentialDeprecationWarning boolean that's passed from the client config to the client to the provider and then used to know whether to send the warning. if it would make more sense I could instead add an additional field to the `ClientConfiguration` and `tfe.Client` in go-tfe to accomplish the same thing

internal/provider/provider.go
```diff
@@ -153,19 +153,30 @@ func configure() schema.ConfigureContextFunc {
...
 			return nil, diag.FromErr(err)
 		}
 
+		var diagnosticWarnings diag.Diagnostics = nil
+		if sendCredentialDeprecationWarning {
+			diagnosticWarnings = diag.Diagnostics{
+				diag.Diagnostic{
+					Severity: diag.Warning,
+					Summary:  "Authentication method invalid for TFE Provider with HCP Terraform and Terraform Enterprise",
+					Detail:   "Use a TFE_TOKEN variable in the workspace or the token argument for the provider. This authentication method will be deprecated in a future version.",
```
christian-doucette: let me know if this warning message makes sense

internal/provider/provider_next.go
```diff
@@ -111,13 +111,17 @@ func (p *frameworkProvider) Configure(ctx context.Context, req provider.Configur
...
-	tfeClient, err := client.GetClient(data.Hostname.ValueString(), data.Token.ValueString(), data.SSLSkipVerify.ValueBool())
+	tfeClient, sendCredentialDeprecationWarning, err := client.GetClient(data.Hostname.ValueString(), data.Token.ValueString(), data.SSLSkipVerify.ValueBool())
 
 	if err != nil {
 		res.Diagnostics.AddError("Failed to initialize HTTP client", err.Error())
 		return
 	}
 
+	if sendCredentialDeprecationWarning {
+		res.Diagnostics.AddWarning("Authentication method invalid for TFE Provider with HCP Terraform and Terraform Enterprise", "Use a TFE_TOKEN variable in the workspace or the token argument for the provider. This authentication method will be deprecated in a future version.")
```
christian-doucette: I wasn't sure of the difference between `provider.go` and `provider_next.go`, should this warning be happening in both of them?

internal/client/client.go
```diff
@@ -81,10 +81,10 @@ func getTokenFromCreds(services *disco.Disco, hostname svchost.Hostname) string
 //
 // Internally, this function caches configured clients using the specified
 // parameters
-func GetClient(tfeHost, token string, insecure bool) (*tfe.Client, error) {
-	config, err := configure(tfeHost, token, insecure)
+func GetClient(tfeHost, token string, insecure bool) (*tfe.Client, bool, error) {
+	config, sendCredentialDeprecationWarning, err := configure(tfeHost, token, insecure)
```
brandonc: Instead of drilling a third return value through to `configure()`, I'd probably add a new field to `ClientConfiguration`... something like, `CredentialSource`... then combine that outcome with `providerRunningInCloud` in the provider code, closer to the diagnostic

internal/provider/provider.go
brandonc: Looks good though I think you should be explicit about which authentication method is invalid.

internal/client/client.go
brandonc: I think that would make it much more flexible/readable as well as moving the different concerns together

internal/provider/provider.go
ctrombley: It's called [Autogenerated API Tokens](https://developer.hashicorp.com/terraform/enterprise/run/run-environment#autogenerated-api-token) in the docs

internal/provider/provider_next.go
shwetamurali: I believe you should keep it in both

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ FEATURES:
 * `r/tfe_variable_set`: Adds `stack_ids` attribute for managing which stacks a variable set is applied to.
 * `d/tfe_variable_set`: Adds `stack_ids` computed attribute to retrieve which stacks a variable set is applied to.
 
+ENHANCEMENTS:
```
christian-doucette: lmk if this should be under Bug Fixes instead

internal/client/client.go
```diff
@@ -74,14 +74,30 @@ func getTokenFromCreds(services *disco.Disco, hostname svchost.Hostname) string
...
+	tokenSource tokenSource
+}
+
+// Using presence of TFC_AGENT_VERSION to determine if this provider is running on HCP Terraform / enterprise
+func providerRunningInCloud() bool {
+	_, envVariablePresent := os.LookupEnv("TFC_AGENT_VERSION")
+	return envVariablePresent
+}
+
+func (pc *ProviderClient) SendAuthenticationWarning() bool {
```
ctrombley, line 89: Consider updating the name of this method to `ValidateTokenSource`?

Sending the auth warning is something that provider.go is deciding to do based on the return value of this method.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ FEATURES:
...
+* Adds warning when using credentials/config file for authentication and running on cloud
```
ctrombley:
```suggestion
* Adds warning when using credentials/config file for authentication and running on cloud, by @christian-doucette [#1890](https://github.com/hashicorp/terraform-provider-tfe/pull/1890/changes#r2636468230)
```

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 ## Unreleased
+ENHANCEMENTS:
+* Adds warning when using credentials/config file for authentication and running on cloud
```
ctrombley: (nit)
```suggestion
* Adds warning when using credentials/config file for authentication and running on cloud, by @christian-doucette [#1890](https://github.com/hashicorp/terraform-provider-tfe/pull/1890/changes#r2636468230)
```

internal/client/client.go
```diff
@@ -74,14 +74,30 @@ func getTokenFromCreds(services *disco.Disco, hostname svchost.Hostname) string
...
 
+// TFE Client along with other necessary information for the provider to run it
+type ProviderClient struct {
+	TfeClient   *tfe.Client
+	tokenSource tokenSource
+}
+
+// Using presence of TFC_AGENT_VERSION to determine if this provider is running on HCP Terraform / enterprise
+func providerRunningInCloud() bool {
+	_, envVariablePresent := os.LookupEnv("TFC_AGENT_VERSION")
```
ctrombley, line 85: How much of a guarantee do we have that this means we are in the cloud? I worry about using an env var since TFE customers might put all sorts of stuff in there.

internal/client/config.go
```diff
@@ -232,11 +241,15 @@ func configure(tfeHost, token string, insecure bool) (*ClientConfiguration, erro
...
 	// If a token wasn't set in the provider configuration block, try and fetch it
 	// from the environment or from Terraform's CLI configuration or configured credential helper.
+
+	tokenSource := providerArgument
 	if token == "" {
 		if os.Getenv("TFE_TOKEN") != "" {
 			token = getTokenFromEnv()
+			tokenSource = environmentVariable
 		} else {
 			token = getTokenFromCreds(services, hostname)
```
ctrombley, line 251: I wonder if we should just check for TFC_AGENT_VERSION here, to avoid the need to wrap the tfe.Client in providerClient?

I think we are only able to dump out here rather than provide a warning diagnostic, but...
Why did we choose a warning diagnostic for this? Shouldn't we error the run?

christian-doucette: the reason for not erroring is to not have any breaking changes. if someone is doing an action with the TFE provider that the autogenerated run token does have permissions for, throwing an error here would cause that to fail when it would previously succeed

discussion on this here:
https://ibm-hashicorp.slack.com/archives/C09KJ217H1D/p1761663767812449?thread_ts=1761662030.666259&cid=C09KJ217H1D

it is unfortunate though since the handling of the warning is pretty annoying compared to an error

internal/client/client.go
christian-doucette: Brandon suggested that here: https://ibm-hashicorp.slack.com/archives/C09KJ217H1D/p1761662332413189?thread_ts=1761662030.666259&cid=C09KJ217H1D

like he mentioned it should always be set on cloud based off this: https://github.com/hashicorp/tfc-agent/blob/8e8f1942f3618a21bc49db29966c80c164ed9209/core/components/terraform/commands.go#L68
although potentially not the best to not have the TFE provider be reliant on the implementation of the TFC agent

it also is possible for anyone to set a TFC_AGENT_VERSION themselves and they will get this warning in cases where it might not be applicable.

christian-doucette: if you know of a better way to check if we're running on cloud let me know though. I dug around myself for a while and wasn't able to find anything good besides environment variables, as far as I can tell environment variables are really the only way the provider can tell about the environment its running on

christian-doucette: alternatively I could remove the cloud check and write a warning that will show up regardless of if the user is on cloud, but specify in the warning that it only applies to users running on cloud. that definitely isn't ideal though

internal/client/config_test.go
```diff
@@ -156,3 +156,52 @@ func TestConfig_cliConfig(t *testing.T) {
 		}
 	}
 }
+
+func TestConfig_configureSetsTokenSource(t *testing.T) {
+	// This tests that the tokenSource is set properly when calling the configure function
+	reset := func() {
+		os.Unsetenv("TFE_TOKEN")
+	}
+	defer reset()
```
brandonc: Use `t.Setenv(key, value)` to set an environment variable within a test function. The Go test runner automatically restores the original value of the variable (or unsets it) when the test, subtest, or the entire suite completes

PR #1957, opened 2026-01-29
Smishra96/fix-required-tfe-versions

smishra96:
## Description

A bug was introduced in the terraform provider tfe which began sending the [new URL param](https://github.com/hashicorp/atlas/pull/23641/changes#diff-02518d854c078934cdd61b905cff34cb6fad3fd480dacfbc7f37afbc40453348R366) unconditionally (causing TFE versions <= 1.0.0 to error on [this line](https://github.com/hashicorp/atlas/blob/27e8665fad9838c8fa3f1dfca790c8ca1204dfaf/app/helpers/api/v2/rendering_helper.rb#L192)) and return an error saying `Error retrieving variable set relations: invalid value for "include" field` .

This PR  addresses this and another similar bug, where a run task's stages attribute might get incorrectly marked as unsupported on newer TFE versions >= 1.0.0

Following resources/data sources were fixed:
- data_source_variable_set.go‎
- resource_stack_variable_set.go
- resource_tfe_variable_set.go
- resource_tfe_workspace_run_task.go‎

_Remember to:_

## Testing plan

1. Create a stacks variable set with TFE with version <= 1.0.0
2. Errors out gracefully with informative diag 

## External links

- [Jira](https://hashicorp.atlassian.net/browse/IPL-8981)

Screenshot 2026-01-30 at 3 56 58 PM

### DIFF
--- a/internal/provider/resource_tfe_variable_set.go
+++ b/internal/provider/resource_tfe_variable_set.go
@@ -16,6 +16,7 @@ import (
 
 	tfe "github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-tfe/internal/provider/helpers"
 )
 
 var variableSetIDRegexp = regexp.MustCompile("varset-[a-zA-Z0-9]{16}$")
@@ -27,7 +28,7 @@ func resourceTFEVariableSet() *schema.Resource {
 		Update: resourceTFEVariableSetUpdate,
 		Delete: resourceTFEVariableSetDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+			StateContext: schema.ImportStatePassthroughWithIdentity("id"),
 		},
 
 		CustomizeDiff: func(c context.Context, d *schema.ResourceDiff, meta interface{}) error {
@@ -39,6 +40,21 @@ func resourceTFEVariableSet() *schema.Resource {
 				return err
 			}
 			return nil
+		},
+
+		Identity: &schema.ResourceIdentity{
+			SchemaFunc: func() map[string]*schema.Schema {
+				return map[string]*schema.Schema{
+					"id": {
+						Type:              schema.TypeString,
+						RequiredForImport: true,
+					},
+					"hostname": {
+						Type:              schema.TypeString,
+						OptionalForImport: true,
+					},
+				}
+			},
 		},
 
 		Schema: map[string]*schema.Schema{
@@ -168,6 +184,11 @@ func resourceTFEVariableSetCreate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
+	err = helpers.WriteTFEIdentity(d, variableSet.ID, config.Client.BaseURL().Host)
+	if err != nil {
+		return err
+	}
+
 	return resourceTFEVariableSetRead(d, meta)
 }
 
@@ -218,6 +239,11 @@ func resourceTFEVariableSetRead(d *schema.ResourceData, meta interface{}) error 
 
 	if variableSet.Parent != nil && variableSet.Parent.Project != nil {
 		d.Set("parent_project_id", variableSet.Parent.Project.ID)
+	}
+
+	err = helpers.WriteTFEIdentity(d, variableSet.ID, config.Client.BaseURL().Host)
+	if err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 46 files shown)


internal/provider/data_source_variable_set.go
```diff
@@ -117,9 +118,14 @@ func dataSourceTFEVariableSetRead(d *schema.ResourceData, meta interface{}) erro
 					d.Set("parent_project_id", vs.Parent.Project.ID)
 				}
 
-				// Only now include vars and workspaces to cut down on request load.
+				includes := []tfe.VariableSetIncludeOpt{tfe.VariableSetWorkspaces, tfe.VariableSetVars}
+				if meetsMinVersionRequirement, err := config.MeetsMinRemoteTFEVersion(minTFEVersionVariableSetStacks); err != nil {
+					log.Printf("[DEBUG] could not determine if TFE version meets minimum required version %s: %v", minTFEVersionVariableSetStacks, err)
```
markdecrane, line 123: I don't like how we're silently swallowing the error here. I'd prefer that we either just error out, or at least provide a visible warning diagnostic to the user. Erroring out is preferable since an issue in determining the TFE version could create an unintended diff in the plan that'd be difficult to track down the reasoning for.

internal/provider/resource_tfe_variable_set.go
```diff
@@ -175,8 +175,17 @@ func resourceTFEVariableSetRead(d *schema.ResourceData, meta interface{}) error
 	config := meta.(ConfiguredClient)
 
 	log.Printf("[DEBUG] Read configuration of variable set: %s", d.Id())
+
+	includes := []tfe.VariableSetIncludeOpt{tfe.VariableSetWorkspaces}
+	meetsMinVersionRequirement, err := config.MeetsMinRemoteTFEVersion(minTFEVersionVariableSetStacks)
+	if err != nil {
+		log.Printf("[DEBUG] could not determine if TFE version meets minimum required version %s: %v", minTFEVersionVariableSetStacks, err)
```
markdecrane, line 203: Same point here as `data_source_variable_set`

Maed223 (commented):
The approach for the fix looks good, but I think we can better handle the error case when the version isn't able to be determined. It's likely best that we just error out in the case it can't be determined. Otherwise silently changing the behavior of resources/data-sources could have unintentional behavior in said error case. This will avoid making the newly introduced version check helper a source of brittleness in its extension. 

Additionally, we should have a changelog entry for the bug being fixed.

internal/provider/data_source_variable_set.go
smishra96: Ahh.. I understand now. I was unsure if I should be erroring out for any reason related to the tfe version validation, but if its part of the state files then I will error out.

Maed223 (commented):
~~For simplicity's sake I think we should just have a `panic(...)` in `MeetsMinRemoteTFEVersion` if an error is encountered. It seems justifiable since we can't reasonably continue without this piece of info (due to the potential for broken state diffs in the errored case), and it'll still allow deferred functions to execute. Otherwise it'll let us clean up some of the extra error handling we have here, and avoid the implicit assumption that we always need to exit from errors here in future use of this helper.~~

Disregard I changed my mind

internal/provider/resource_tfe_workspace_run_task.go
```diff
@@ -328,19 +328,16 @@ func (r *resourceWorkspaceRunTask) UpgradeState(ctx context.Context) map[int64]r
 
 func (r *resourceWorkspaceRunTask) supportsStagesProperty() bool {
 	// The Stages property is available in HCP Terraform and Terraform Enterprise v202404-1 onwards.
-	//
-	// The version comparison here can use plain string comparisons due to the nature of the naming scheme. If
-	// TFE every changes its scheme, the comparison will be problematic.
 	if r.supportsStages == nil {
-		value := r.capabilities.IsCloud() || r.capabilities.RemoteTFEVersion() > "v202404"
+		value, _ := r.config.MeetsMinRemoteTFEVersion(minTFEVersionWorkspaceRunTaskStages)
```
markdecrane: We should also handle the error here

smishra96: How should we do that, would it be ok to handle it with a warn log or should we bubble the error up, that would be a bit of a bigger change.. I do not have much context over this resource.

markdecrane: Since `supportsStagesProperty` is only used in `extractStageAndStages` I'd just place the version check within `extractStageAndStages`. Then in the `err != nil` case you would append an error diagnostic to the passed`*diag.Diagnostics`param and `return nil, nil`. All the invocations of `extractStageAndStages` check for error diags right after and exit if there are any.

PR #1967, opened 2026-02-11
Message fix

shaunakone:
## Description

_Describe why you're making this change._

_Remember to:_

## Testing plan

1.  _the conditions_

## External links

```

...
```

### DIFF
--- a/internal/provider/resource_tfe_workspace_run_test.go
+++ b/internal/provider/resource_tfe_workspace_run_test.go
@@ -443,4 +443,3 @@ func testAccTFEWorkspaceRun_WhenRunErrors(workspaceID string) string {
 	}
 `, workspaceID)
 }
-


internal/provider/workspace_run_helpers.go
```diff
@@ -49,7 +49,9 @@ func createWorkspaceRun(d *schema.ResourceData, meta interface{}, isDestroyRun b
 	waitForRun := runArgs["wait_for_run"].(bool)
 	manualConfirm := runArgs["manual_confirm"].(bool)
 
-	run, err := createRun(config.Client, waitForRun, manualConfirm, isDestroyRun, ws)
+	msg, _ := runArgs["message"].(string)
+	run, err := createRun(config.Client, waitForRun, manualConfirm, isDestroyRun, ws, msg)
```
markdecrane, line 53:
```suggestion
	waitForRun := runArgs["wait_for_run"].(bool)
	manualConfirm := runArgs["manual_confirm"].(bool)
	msg, _ := runArgs["message"].(string)

	run, err := createRun(config.Client, waitForRun, manualConfirm, isDestroyRun, ws, msg)
```

Tiny nit in formatting

internal/provider/resource_tfe_workspace_run_test.go
```diff
@@ -392,3 +392,52 @@ func testAccTFEWorkspaceRun_WhenRunErrors(workspaceID string) string {
 	}
 `, workspaceID)
 }
+func TestAccTFEWorkspaceRun_message(t *testing.T) {
```
markdecrane: We can get this integrated with the existing `TestAccTFEWorkspaceRun_withBothApplyAndDestroyBlocks` test. We'll just have to update the configuration it's using to specify the messages. Then in the `Steps` portion of that test case we'll add a check that the apply run message is what's expected. Unfortunately we won't be able to do that check for the destroy run message. We'd need to specify the `manual_confirm` attribute in the `destroy` block to be `true` in order to still have the run accessible to check (otherwise the workspace is destroyed along with the run to check), but that'll stop the destroy from being able to complete (and generally mess up the test). I still think that it's worthwhile to add the `message` to that `destroy` block even if we can't check it, it'll at least ensure it's presence doesn't blow anything up.

internal/provider/resource_tfe_workspace_run_test.go
```diff
@@ -392,3 +443,52 @@ func testAccTFEWorkspaceRun_WhenRunErrors(workspaceID string) string {
...
+					if run.Message != expected {
+						return fmt.Errorf("expected message %q got %q", expected, run.Message)
+					}
+
+					return nil
+				},
+			},
+		},
+	})
+}
```
markdecrane: We can omit this test now that the existing test covers the behavior.

internal/provider/workspace_run_helpers.go
```diff
@@ -156,26 +158,32 @@ func createRun(tfeClient *tfe.Client, waitForRun bool, manualConfirm bool, isDes
...
 	runConfig := tfe.RunCreateOptions{
 		Workspace: ws,
 		IsDestroy: tfe.Bool(isDestroyRun),
-		Message: tfe.String(fmt.Sprintf(
-			"Triggered by tfe_workspace_run resource via terraform-provider-tfe on %s",
-			time.Now().Format(time.UnixDate),
-		)),
+		Message:   tfe.String(message),
 		AutoApply: tfe.Bool(autoApply),
 	}
```
markdecrane: We can actually invert this logic to make it simpler. When the platform sees a nil message, it will automatically have the default run message like you specified above, so we don't need to supply it in this case. In fact it's better to defer to what the platform decides should be the message in this case so that the default message logic is in one place on the platform side (if it's to ever change).

What we can do instead is to check if `message != ""`, and if so we can update the `runConfig.Message` to hold that value, otherwise it'll just remain nil.

Issue #1974, opened 2026-02-18
tfe_variable value_wo sourced from ephemeral resource causes perpetual diff and value rotation every apply


jackivanov:
#### Description

When tfe_variable.value_wo is set from an ephemeral resource (for example ephemeral.random_password), the variable is replaced on every subsequent terraform apply, even if the configuration has not changed.

Because the replacement happens each time, the variable content is also rotated on every apply (the stored value in Terraform Cloud/Enterprise changes every run). This occurs regardless of whether sensitive is set to true or false.

#### Terraform Enterprise version

Terraform Cloud (HCP)

#### Terraform version

```plaintext
% terraform version
Terraform v1.14.5
on darwin_arm64
+ provider registry.terraform.io/hashicorp/random v3.8.1
+ provider registry.terraform.io/hashicorp/tfe v0.74.0
```

#### Terraform Configuration Files

```terraform
ephemeral "random_password" "test" {
  length           = 16
  override_special = "!#$%&*()-_=+[]{}<>:?"
}

resource "tfe_variable" "test" {
  key          = "random_password"
  value_wo     = ephemeral.random_password.test.result
  category     = "terraform"
  workspace_id = "ws-qVvMy5oVoicxLXfU"
  description  = "Random password for the workspace"
  sensitive    = false
}

terraform {
  required_providers {
    tfe = {
      source  = "hashicorp/tfe"
      version = "~> 0.74"
    }
  }
}

provider "tfe" {
  organization = "upwork-inc"
  token        = var.tfc_user_token
}

variable "tfc_user_token" {}

```

#### Debug Output

```plaintext
% terraform apply
ephemeral.random_password.test: Opening...
ephemeral.random_password.test: Opening complete after 0s
ephemeral.random_password.test: Closing...
ephemeral.random_password.test: Closing complete after 0s

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following
symbols:
  + create

Terraform will perform the following actions:

  # tfe_variable.test will be created
  + resource "tfe_variable" "test" {
      + category       = "terraform"
      + description    = "Random password for the workspace"
      + hcl            = false
      + id             = (known after apply)
      + key            = "random_password"
      + readable_value = (known after apply)
      + sensitive      = false
      + value_wo       = (write-only attribute)
      + workspace_id   = "ws-qVvMy5oVoicxLXfU"
        # (1 unchanged attribute hidden)
    }

[...]

PR #1984, opened 2026-03-09
TF applies now converge when the value_wo write only value is used in policy_set_parameter

Uk1288:
## Description

When the `value_wo` is used in `policy_set_parameter` resource, `tf plan` always result in new changes resulting in non-convergent `tf apply`.

```
  length           = 16
}

  secret_string_wo_version = 1
}

}

  username            = "admin"
}
```

```
  length           = 16
}

}
```

_Remember to:_

## Old Vs New Config

1.  Old config with `value_wo`, ensure to use the public tfe version
```
  length           = 16
}

resource "tfe_policy_set_parameter" "test_policy_set_parameter" {
  key          = "my_key_name"
  value_wo    = ephemeral.random_password.test.result
  policy_set_id = "test_policy_set_id"
}
```

2.  New config with `value_wo_version` to use with this PR,
```
resource "tfe_policy_set_parameter" "test_policy_set_parameter" {
  key          = "my_key_name"
  value_wo    = ephemeral.random_password.test.result
  value_wo_version    = 1
  policy_set_id = "test_policy_set_id"
}
```

## Testing plan
1.  run `terraform apply` using the old config & public tfe version, it should succeed with changes applied
1.  run another `terraform apply`, it should show new changes to apply for every subsequent apply
1.  with this PR, a second attribute called `value_wo_version` is added in line with [terraform ephemeral values](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/ephemeral)
1.  run `terraform apply`, changing the version now triggers update in the policy resource

## External links

- [Issue](https://github.com/hashicorp/terraform-provider-tfe/issues/1974)

```
$ TESTARGS="-run TestAccTFEPolicySetParameter" make testacc

...
```

## Rollback Plan
1. Revert PR
3. Remove the new attribute `value_wo_attribute` from affected config

### DIFF
--- a/internal/provider/resource_tfe_policy_set_parameter.go
+++ b/internal/provider/resource_tfe_policy_set_parameter.go
@@ -297,6 +297,7 @@ func (r *resourceTFEPolicySetParameter) Update(ctx context.Context, req resource
 	p, err := r.config.Client.PolicySetParameters.Update(ctx, plan.PolicySetID.ValueString(), plan.ID.ValueString(), options)
 	if err != nil {
 		resp.Diagnostics.AddError(fmt.Sprintf("Error updating parameter %s", plan.ID), err.Error())
+		return
 	}
 
 	// Update state


internal/provider/resource_tfe_policy_set_parameter.go
```diff
@@ -305,15 +299,8 @@ func (r *resourceTFEPolicySetParameter) Update(ctx context.Context, req resource
 		resp.Diagnostics.AddError(fmt.Sprintf("Error updating parameter %s", plan.ID), err.Error())
```
markdecrane, line 299: Do we need an early return here now? Otherwise if we error won't we hit a panic with a nil pointer dereference in `modelFromTFEPolicySetParameter`?

Uk1288: that's right!

Uk1288: done ✅

PR #1992, opened 2026-03-10
Write-only value is not safed when used in tfe_organization_run_task resource

Uk1288:
## Description

When the `hmac_key_wo_version` is used in `tfe_organization_run_task` resource

```
  length           = 16
}

  secret_string_wo_version = 1
}

}

  username            = "admin"
}
```

```
  length           = 16
}

}
```

_Remember to:_

## Old Vs New Config

1.  Old config with `hmac_key_wo`, ensure to use the public tfe version
```
resource "tfe_organization_run_task" "example" {
  organization = "hashicorp"
  url          = "http://url"
  name         = "task-test"
  enabled      = true
  description  = "An example task"
  hmac_key_wo = "hmac_key_test"
}
```
2.  New config with `hmac_key_wo_version` to use with this PR,
```
resource "tfe_organization_run_task" "example" {
  organization = "hashicorp"
  url                  = "http://url"
  name         = "task-test"
  enabled      = true
  description  = "An example task"
  hmac_key_wo = "hmac_key_test"
  hmac_key_wo_version = 2
}
```

## Testing plan

1.  with this PR, a second attribute called `hmac_key_wo_version` is added in line with [terraform ephemeral values](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/ephemeral)
1.  run `terraform apply`, changing the version now triggers update in the `hmac_key_wo`

## External links

## Rollback Plan
1. Revert PR
4. Remove the new attribute `hmac_key_wo_version` from affected config

### DIFF
--- a/internal/provider/resource_tfe_organization_run_task_test.go
+++ b/internal/provider/resource_tfe_organization_run_task_test.go
@@ -272,12 +272,6 @@ func TestAccTFEOrganizationRunTask_HMACWriteOnly(t *testing.T) {
 					resource.TestCheckNoResourceAttr("tfe_organization_run_task.foobar", "hmac_key_wo"),
 					resource.TestCheckNoResourceAttr("tfe_organization_run_task.foobar", "hmac_key_wo_version"),
 				),
-				// Ensure that the resource has been replaced
-				ConfigStateChecks: []statecheck.StateCheck{
-					compareValuesDiffer.AddStateValue(
-						"tfe_organization_run_task.foobar", tfjsonpath.New("id"),
-					),
-				},
 			},
 		},
 	})


internal/provider/resource_tfe_organization_run_task.go
```diff
@@ -42,34 +41,37 @@ func NewOrganizationRunTaskResource() resource.Resource {
...
+		Enabled:          types.BoolValue(v.Enabled),
+		HMACKey:          types.StringValue(""), // This value is never emitted by the API so we inject it later
+		ID:               types.StringValue(v.ID),
+		Name:             types.StringValue(v.Name),
+		Organization:     types.StringValue(v.Organization.Name),
+		URL:              types.StringValue(v.URL),
+		HMACKeyWOVersion: hmacKeyWOVersion,
 	}
 
 	if len(hmacKey.String()) > 0 {
```
markdecrane, line 69: nit since this is previous behavior:

`string.String()` will return a quoted value (like `"\"value"\"`) or otherwise `"null"` if the value were null, so this check is a bit misleading. Something like `!hmacKey.IsNull() && hmacKey.ValueString() != ""` would be more appropriate.

Uk1288: I tested this with `hmac_key = null` and it seems like the zero value of the type `schema.StringAttribute` is always set to `""`, here is the output when it is explicitly set to null or not set at all: `hmacKey=basetypes.StringValue{state:0x2, value:\"\"}"`

PR #1995, opened 2026-03-10
Add Write-only token version to tfe_team_notification_configuration resource

Uk1288:
## Description

When the `token_wo` is used in `tfe_team_notification_configuration` resource, the `token_wo_version` is now used to control its update.

```
  length           = 16
}

  secret_string_wo_version = 1
}

}

  username            = "admin"
}
```

```
  length           = 16
}

}
```

_Remember to:_

## Old Vs New Config

1.  Old config with `token_wo`, ensure to use the public tfe version
```
resource "tfe_team_notification_configuration" "test" {
  name             = "my-test-email-notification-configuration"
  destination_type = "generic"
  url = "http://testing-urll"
  token_wo = "TOKEN"
  triggers         = ["change_request:created"]
  team_id          = tfe_team.foobar.id
}

```
2.  New config with `token_wo_version` to use with this PR,
```
resource "tfe_team_notification_configuration" "test" {
  name             = "my-test-email-notification-configuration"
  destination_type = "generic"
  url = "http://testing-urll"
  token_wo = "TOKEN"
  token_wo_version = 2
  triggers         = ["change_request:created"]
  team_id          = tfe_team.foobar.id
}

```

## Testing plan

1.  with this PR, a second attribute called `token_wo_version` is added in line with [terraform ephemeral values](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/ephemeral)
1.  run `terraform apply`, changing the version now triggers update in the `token_wo`

## External links

## Rollback Plan
1. Revert PR
3. Remove the new attribute `token_wo_version` from affected config

### DIFF
--- a/internal/provider/resource_tfe_team_notification_configuration.go
+++ b/internal/provider/resource_tfe_team_notification_configuration.go
@@ -18,7 +18,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
-	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
@@ -227,9 +226,6 @@ func (r *resourceTFETeamNotificationConfiguration) Schema(ctx context.Context, r
 				Validators: []validator.Int64{
 					int64validator.ConflictsWith(path.MatchRoot("token")),
 					int64validator.AlsoRequires(path.MatchRoot("token_wo")),
-				},
-				PlanModifiers: []planmodifier.Int64{
-					int64planmodifier.RequiresReplace(),
 				},
 			},
 
@@ -536,8 +532,9 @@ func (r *resourceTFETeamNotificationConfiguration) ImportState(ctx context.Conte
 }
 
 // determineTokenForUpdate is invoked only after terraform determines that an attribute update is needed.
-// note that the update can be triggered by other attributes outside of the token attributes.
-// this function compares the write-only token version vs token to ensure that token is not mistakenly unset when its not intended.
+// note that the update can be triggered by other attributes outside of the token/token_wo attributes.
+// this function compares the TokenWOVersion vs Token to ensure that during api update call, token is not mistakenly unset.
+// Returns nil if no token update is needed.
 func (r *resourceTFETeamNotificationConfiguration) determineTokenForUpdate(plan, state, config modelTFETeamNotificationConfiguration) (updateToken *string, isWOVal bool) {
 	// Determine if we're using write-only token in plan vs state
 	usingWriteOnlyInPlan := !plan.TokenWOVersion.IsNull()
--- a/internal/provider/resource_tfe_team_notification_configuration_test.go
+++ b/internal/provider/resource_tfe_team_notification_configuration_test.go
@@ -1003,13 +1003,6 @@ func TestAccTFETeamNotificationConfiguration_tokenWO(t *testing.T) {
 					resource.TestCheckNoResourceAttr("tfe_team_notification_configuration.foobar", "token_wo"),
 					resource.TestCheckResourceAttr("tfe_team_notification_configuration.foobar", "token_wo_version", "2"),
 				),
-				// Register the id with the value comparer so we can assert that the
-				// resource has been replaced in the next step.
-				ConfigStateChecks: []statecheck.StateCheck{
-					compareValuesDiffer.AddStateValue(
-						"tfe_team_notification_configuration.foobar", tfjsonpath.New("id"),
-					),
-				},
 			},
 			{
 				Config: testAccTFETeamNotificationConfiguration_token(org.Name),
@@ -1017,12 +1010,6 @@ func TestAccTFETeamNotificationConfiguration_tokenWO(t *testing.T) {
 					resource.TestCheckResourceAttr("tfe_team_notification_configuration.foobar", "token", "1234567890"),
 					resource.TestCheckNoResourceAttr("tfe_team_notification_configuration.foobar", "token_wo"),
 				),
-				// Ensure that the resource has been replaced
-				ConfigStateChecks: []statecheck.StateCheck{
-					compareValuesDiffer.AddStateValue(
-						"tfe_team_notification_configuration.foobar", tfjsonpath.New("id"),
-					),
-				},
 			},
 		},
 	})


internal/provider/resource_tfe_team_notification_configuration.go
```diff
@@ -212,9 +216,19 @@ func (r *resourceTFETeamNotificationConfiguration) Schema(ctx context.Context, r
...
+
+			"token_wo_version": schema.Int64Attribute{
+				Optional:    true,
+				Description: "Version of the write-only token to trigger updates",
+				Validators: []validator.Int64{
+					int64validator.ConflictsWith(path.MatchRoot("token")),
+					int64validator.AlsoRequires(path.MatchRoot("token_wo")),
+				},
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.RequiresReplace(),
```
jillirami: To my understanding, if a user increments the version, Terraform will evaluate `RequiresReplace` and recreate the resource as opposed to calling the Update method. Should we remove this require?

```suggestion
					int64planmodifier,
```

internal/provider/resource_tfe_team_notification_configuration.go
```diff
@@ -428,7 +425,11 @@ func (r *resourceTFETeamNotificationConfiguration) Update(ctx context.Context, r
 
 	// Read Terraform prior state data into the model
 	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
-
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// Read configuration data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
```
jillirami, line 427: 🙌

internal/provider/resource_tfe_team_notification_configuration.go
```diff
@@ -484,16 +486,12 @@ func (r *resourceTFETeamNotificationConfiguration) Update(ctx context.Context, r
...
-	result, diags := modelFromTFETeamNotificationConfiguration(tnc, !config.TokenWO.IsNull(), plan.Token)
+	result, diags := modelFromTFETeamNotificationConfiguration(tnc, config.TokenWOVersion)
 	if diags.HasError() {
 		resp.Diagnostics.Append((diags)...)
 		return
 	}
 
-	// Write the hashed private key to the state if it was provided
-	store := r.writeOnlyValueStore(resp.Private)
-	resp.Diagnostics.Append(store.SetPriorValue(ctx, config.TokenWO)...)
```
jillirami, line 495: In both create and update notification methods, `store := r.writeOnlyValueStore(resp.Private)` persists the write-only value to to provider-private storage. 

As you have identified, this contradicts Terraform Plugin Framework [docs](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) expectation that write-only arguments should be completely ephemeral.

Of note regarding `resp.Diagnostics.Append(store.SetPriorValue(ctx, config.TokenWO)...)` - it seems that storing a secure hash in provider private state is an explicitly recommended pattern for write-only arguments, as long as the raw secret is not persisted. Reference: https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments#best-practices 

> Use the resource's [private state](https://developer.hashicorp.com/terraform/plugin/framework/resources/private-state) to store secure hashes of write-only argument values, the provider will then use the hash to determine if a write-only argument value has changed in later Terraform runs.

internal/provider/resource_tfe_team_notification_configuration.go
```diff
@@ -104,11 +106,12 @@ func modelFromTFETeamNotificationConfiguration(v *tfe.NotificationConfiguration,
 		result.EmailUserIDs = types.SetValueMust(types.StringType, emailUserIDs)
 	}
 
-	if lastValue.String() != "" {
-		result.Token = lastValue
+	if v.Token != "" {
+		result.Token = types.StringValue(v.Token)
 	}
```
jillirami: Can we expect that the API returns a plain-text token value? To my current understanding this condition would not be met, and we would need to preserve last-known-token

internal/provider/resource_tfe_team_notification_configuration.go
Uk1288: This is a good call out!

**Existing issue with the private state route:**

```
  length           = 16
}

  secret_string_wo_version = 1
}

}

  username            = "admin"
}
```

In contrast, because the tfe-provider currently relies on the WO value for update detection, the resources get indefinitely updated and the config never converges. [See reported issue here](https://github.com/hashicorp/terraform-provider-tfe/issues/1974).

Example for the notification resource. Running an apply always shows an update that needs to be applied.

```
  length           = 16
}

resource "tfe_team_notification_configuration" "foobar" {
  name             = "team notification test"
  destination_type = "generic"
  url = "http://notification"
  triggers         = ["change_request:created"]
  team_id          = tfe_team.foobar.id
}
```

internal/provider/resource_tfe_team_notification_configuration.go
Uk1288: yes you are right, token will not be sent back. This change will be updated

internal/provider/resource_tfe_team_notification_configuration.go
Uk1288: while testing, I noticed that the provider was updating the resource if the `RequiresReplace` was not explicitly stated.

ctrombley: I think this API supports updates for the token, though! I agree that we could remove this RequiresReplace and make better use of updates here.

internal/provider/resource_tfe_team_notification_configuration.go
```diff
@@ -520,6 +515,30 @@ func (r *resourceTFETeamNotificationConfiguration) ImportState(ctx context.Conte
...
+	if usingWriteOnlyInPlan && plan.TokenWOVersion.ValueInt64() != state.TokenWOVersion.ValueInt64() && !config.TokenWO.IsNull() {
+		return config.TokenWO.ValueStringPointer()
+	}
+	// Case 4: Regular token changed. Only set Token if our planned value would be a CHANGE from
+	// the prior state. This prevents accidentally resetting the token on unrelated changes.
+	if state.Token.ValueString() != plan.Token.ValueString() {
+		return plan.Token.ValueStringPointer()
+	}
+	return nil
 }
```
ctrombley, line 561: Can this be simplified?  I don't think we need to check "did it switch from X to Y?" or "did the version change?". Terraform has already decided an update is necessary. We just need to pick the correct current value, something like:

```suggestion
   func (r *resourceTFENotificationConfiguration) determineTokenForUpdate(config, plan
  modelTFENotificationConfiguration) *string {
       // Priority 1: Write-only token is being set in this specific apply
       if !config.TokenWO.IsNull() {
           return config.TokenWO.ValueStringPointer()
       }

       // Priority 2: Standard token is present in the plan
       // Note: We don't need to check if it changed. If Update() is called,
       // and this value is in the plan, it's the intended value.
       if !plan.Token.IsNull() {
           return plan.Token.ValueStringPointer()
       }

       // If neither is present, we send nil (no update to token)
       return nil
   }
```

I don't think we need to involve state comparisons at all here.

ctrombley (commented):
This is looking really good Ukeme! I had a few questions and comments.

internal/provider/resource_tfe_team_notification_configuration.go
Uk1288: The [previous plan modifier forced a replace](https://github.com/hashicorp/terraform-provider-tfe/blob/087c54bfca62cd6a2b929518168fa4dd193e5dee/internal/provider/resource_tfe_team_notification_configuration.go#L217) whenever the token_wo changed. There is a test to [track that the notification config gets replaced](https://github.com/hashicorp/terraform-provider-tfe/blob/main/internal/provider/resource_tfe_team_notification_configuration_test.go#L955). Removing the replace logic breaks this test, it makes me wonder if there was a reason why the re-create was intentionally tested for.

For this reason, I'd like to keep the existing behaviour to keep this PR's scope to `moving from private state mode` to `version mode`. Behavioural changes can be in a followup PR.

internal/provider/resource_tfe_team_notification_configuration.go
Uk1288: Notice that I have kept the behaviour where token used to create/update is manually persisted across state. 
I attempted to remove those lines too but it causes some state inconsistency errors during applies. I have left this existing behaviour to be targeted in another PR.

internal/provider/resource_tfe_team_notification_configuration.go
Uk1288: The `determineTokenForUpdate` takes into consideration the `token`, `token_wo` and `token_wo_version` attributes. Notice that the update hook can be triggered by changes in other attributes apart from these 3. 

When an update is triggered by other attributes, we don’t want to unset the token value in the backend request. Remember that the tokens are not saved in the terraform state, so if we go with the above logic, token will be set to null or incorrectly set to a new value of `token_wo` (the above logic does not take into account the token_version). 

Mostly the `determineTokenForUpdate` will decide if an updated token value should be sent when a change is triggered by the other attributes. It must be able to decide without having all the token values, hence version numbers are compared. The code above re-introduces the reported issue.

internal/provider/resource_tfe_team_notification_configuration.go
ctrombley: @Uk1288, after reviewing, I agree that the original design using private hashed values was incorrect:
- write only values require data from ephemeral contexts, and those change every time.
- to support ephemeral values properly, we must move away from implicit hash-based detection on `token_wo` and rely explicitly on `token_wo_version` for updates.

I think that means the existing test is wrong and can be updated! I still think that RequiresReplace is not necessary, and can be removed (and the test can be updated) to make better use of the API's ability to update existing resources. Thoughts?

Uk1288: I would say that the private hashed values was another way of implementing it. It was not incorrect per say.

Yes I am good to remove the `requiresreplace` and remove the accompanying tests. I will push that out to the other resources as well.

PR #2040, opened 2026-04-21
fix: treat omitted remote_state_consumer_ids as an empty set during updates

### LINKED ISSUE #2000: Removing remote_state_consumer_ids from a tfe_workspace_settings resource does not actually remove them
#### Terraform version
```plaintext
1.13.4
```

I had this in my TF configuration:
```terraform
resource "tfe_workspace_settings" "example" {
  workspace_id   = tfe_workspace.example.id
  agent_pool_id  = tfe_agent_pool.example.id
  ...
  remote_state_consumer_ids = [tfe_workspace.consumer.id]
}
```

Then I wanted to stop sharing the state of the `example` workspace with the `consumer` workspace, so I removed the `remote_state_consumer_ids` parameter completely:

```terraform
resource "tfe_workspace_settings" "example" {
  workspace_id   = tfe_workspace.example.id
  agent_pool_id  = tfe_agent_pool.example.id
  ...
}
```

#### Expected Behavior
All consumers should have been removed, the state of the `example` workspace should not be shared with any workspace anymore. That's what my Terraform configuration says declaratively.

#### Actual Behavior
The previous consumers are still present, and my TF plans do not find any changes to apply. 

#### Additional Context
The only workaround I found was to explicitly set `remote_state_consumer_ids = []`, which removes all consumers. Example:
```terraform
resource "tfe_workspace_settings" "example" {
  workspace_id   = tfe_workspace.example.id
  agent_pool_id  = tfe_agent_pool.example.id
  ...
  remote_state_consumer_ids = []
}
```

ctrombley:
## Description

This fix PR resolves #2000.

## Testing plan

1.  Reproduce with a config similar to that provided in #2000
2. Ensure that with the fix in place, omitting `remote_state_consumer_ids` from a config that previously used it clears out the relationships successfully.

### DIFF
--- a/internal/provider/resource_tfe_workspace_settings.go
+++ b/internal/provider/resource_tfe_workspace_settings.go
@@ -93,6 +93,8 @@ type validateRemoteStateConsumerIDs struct{}
 // explicit empty set when it is removed from configuration after being set.
 type unsetRemoteStateConsumerIDsIfOmitted struct{}
 
+const remoteStateConsumerIDsConfiguredKey = "remote_state_consumer_ids_configured"
+
 // validate global and remote state mutual exclusion
 type validateRemoteStateExclusion struct{}
 
@@ -107,12 +109,23 @@ var _ planmodifier.Set = (*unsetRemoteStateConsumerIDsIfOmitted)(nil)
 var _ planmodifier.Set = (*validateRemoteStateConsumerIDs)(nil)
 var _ planmodifier.Bool = (*validateRemoteStateExclusion)(nil)
 
-func (m unsetRemoteStateConsumerIDsIfOmitted) PlanModifySet(_ context.Context, req planmodifier.SetRequest, resp *planmodifier.SetResponse) {
-	if req.Plan.Raw.IsNull() || req.State.Raw.IsNull() {
+func (m unsetRemoteStateConsumerIDsIfOmitted) PlanModifySet(ctx context.Context, req planmodifier.SetRequest, resp *planmodifier.SetResponse) {
+	if req.Plan.Raw.IsNull() {
 		return
 	}
 
-	if !req.ConfigValue.IsNull() && !req.ConfigValue.IsUnknown() {
+	if !req.ConfigValue.IsNull() {
+		resp.Diagnostics.Append(resp.Private.SetKey(ctx, remoteStateConsumerIDsConfiguredKey, []byte("true"))...)
+		return
+	}
+
+	if req.State.Raw.IsNull() {
+		return
+	}
+
+	wasConfigured, diags := req.Private.GetKey(ctx, remoteStateConsumerIDsConfiguredKey)
+	resp.Diagnostics.Append(diags...)
+	if diags.HasError() || len(wasConfigured) == 0 {
 		return
 	}
 
@@ -558,7 +571,7 @@ func (r *workspaceSettings) readSettings(ctx context.Context, workspaceID string
 	return r.workspaceSettingsModelFromTFEWorkspace(ws), nil
 }
 
-func (r *workspaceSettings) updateSettings(ctx context.Context, data *modelWorkspaceSettings, state *tfsdk.State) error {
+func (r *workspaceSettings) updateSettings(ctx context.Context, data *modelWorkspaceSettings, priorState, targetState *tfsdk.State) error {
 	workspaceID := data.WorkspaceID.ValueString()
 
 	updateOptions := tfe.WorkspaceUpdateOptions{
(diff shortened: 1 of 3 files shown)


internal/provider/resource_tfe_workspace_settings.go
```diff
@@ -99,9 +105,41 @@ type validateSelfReference struct{}
...
 var _ planmodifier.Set = (*validateRemoteStateConsumerIDs)(nil)
 var _ planmodifier.Bool = (*validateRemoteStateExclusion)(nil)
 
+func (m unsetRemoteStateConsumerIDsIfOmitted) PlanModifySet(ctx context.Context, req planmodifier.SetRequest, resp *planmodifier.SetResponse) {
+	if req.Plan.Raw.IsNull() {
+		return
+	}
+
+	if !req.ConfigValue.IsNull() {
+		resp.Diagnostics.Append(resp.Private.SetKey(ctx, remoteStateConsumerIDsConfiguredKey, []byte("true"))...)
```
Uk1288, line 118: After the `remoteStateConsumerIDsConfiguredKey` is set, it looks like it never gets unset. If a user removes remote_state_consumer_ids from their config, the modifier will correctly clear the consumers because the key is present.
However, on subsequent runs (even if remote_state_consumer_ids was never re-added), the key remains set, so the modifier will continue to treat the attribute as "was previously configured and now omitted,".

internal/provider/resource_tfe_workspace_settings.go
```diff
@@ -736,16 +780,28 @@ func (r *workspaceSettings) Create(ctx context.Context, req resource.CreateReque
...
 func (r *workspaceSettings) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
-	var data modelWorkspaceSettings
+	var data, configured modelWorkspaceSettings
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	resp.Diagnostics.Append(req.Config.Get(ctx, &configured)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if configured.ExecutionMode.IsNull() {
```
Uk1288: Is the update to the `ExecutionMode` and `AgentPoolID` part of the fix for remote consumer IDs unsetting or its fixing a different issue?

Uk1288 (commented):
Good progress!

I came across one behaviour worth noting. Importing a workspace setting with existing consumer ids can sometime incorrectly result in no-op instead of a destroy plan.

### Steps to reproduce:
- Create a workspace manually.
- Add one or more remote state consumers out of band in the UI or API.
- create this config:
```
resource "tfe_workspace_settings" "example" {
  workspace_id         = "ws-123"
  global_remote_state  = false
  project_remote_state = false
}
```
- Import the settings resource:
`terraform import tfe_workspace_settings.example ws-123`
- Run plan with the config.

**Result:**
Even though the resource has existing explicit consumers, because the private key was never set before import, the plan modifier does nothing.
The plan results in a no-op indicating the real world infra matches state which is incorrect. Expected behaviour would be to plan removal diff.

internal/provider/resource_tfe_workspace_settings.go
ctrombley: I'm not sure what this is doing here, it looks like a mistake. Addressed in 02c7b2a8.

internal/provider/resource_tfe_workspace_settings.go
ctrombley: Addressed in 02c7b2a8.

internal/provider/resource_tfe_workspace_settings.go
```diff
@@ -562,8 +613,22 @@ func (r *workspaceSettings) updateSettings(ctx context.Context, data *modelWorks
...
 		updateOptions.ExecutionMode = tfe.String(executionMode)
-		updateOptions.SettingOverwrites.ExecutionMode = tfe.Bool(true)
-		updateOptions.SettingOverwrites.AgentPool = tfe.Bool(true)
+
+		// Determine whether to mark execution_mode as overwritten. During
+		// Create the overwrites field is null/unknown (computed, not in config)
+		// so we always send the override. During Update, only send the
+		// override when the planned overwrites confirm it — this avoids
+		// inconsistencies when execution_mode is merely inherited from state.
+		sendOverwrite := data.Overwrites.IsNull() || data.Overwrites.IsUnknown()
```
Uk1288: 💅🏾 this code block seems a bit confusing because of how `sendOverwrite` is set and updated within the `if` block. Would it be clearer to write it this way?


```suggestion
		if data.Overwrites.IsNull() || data.Overwrites.IsUnknown() {
			updateOptions.SettingOverwrites.ExecutionMode = tfe.Bool(true)
			updateOptions.SettingOverwrites.AgentPool = tfe.Bool(true)
		} else {
			var overwritesPlanned []modelOverwrites
			data.Overwrites.ElementsAs(ctx, &overwritesPlanned, true)
			if len(overwritesPlanned) > 0 && overwritesPlanned[0].ExecutionMode.ValueBool() {
				updateOptions.SettingOverwrites.ExecutionMode = tfe.Bool(true)
				updateOptions.SettingOverwrites.AgentPool = tfe.Bool(true)
			}
		}
```

ctrombley: Addressed in [eb62c3f](https://github.com/hashicorp/terraform-provider-tfe/pull/2040/commits/eb62c3f2ee46676a9f31bdc4e3d22ab56ef3eda4).

Uk1288 (approved):

PR #2046, opened 2026-04-30
Adding Post Apply Status Constants in Run

jose-kunnel:
## Description

This PR adds support for new run status values introduced in the TFE API.

Currently, the provider does not recognize the following run statuses:

post_apply_running
post_apply_completed

These statuses are not included in the applyPendingStatuses list within resource_tfe_workspace_run.go, which causes runs with post-apply run tasks to fail. When such a run enters the post_apply_running state, it is treated as an invalid status and results in an error.

_Remember to:_

## Testing plan

Before this change

Creating a run with a post-apply run task would fail when transitioning to post_apply_running, as the provider does not recognize this state.

After this change

Runs with post-apply run tasks are handled correctly, and transitions through post_apply_running and post_apply_completed no longer cause errors.

## External links
- [JIRA] (https://hashicorp.atlassian.net/browse/IPL-9456)
- [Related PR](https://github.com/hashicorp/go-tfe/pull/1323)


ctrombley (approved):
Looks good, thanks!

PR #2072, opened 2026-05-19
Adds `r/tfe_scim_settings` and `d/tfe_scim_settings` to manage SCIM settings in TFE

skj-skj:
## Description

Adds `r/tfe_scim_settings` and `d/tfe_scim_settings` to manage SCIM provisioning on Terraform Enterprise. SCIM depends on SAML, so this resource is meant to be used alongside `tfe_saml_settings`.

While the resource exists, SCIM is always enabled, `terraform destroy` disables it. You can `pause` provisioning, and optionally link a SCIM group to site admin via `site_admin_group_scim_id` (setting it to `""` unlinks). If SCIM is disabled out-of-band, the next plan will propose re-creating the resource.

Also bumps `go-tfe` to `v1.106.0` and updates the CI skip regex to exclude `TestAccTFESCIMSettings_omnibus` — like SAML, SCIM is a singleton, so all cases live in one omnibus test with non-parallel subtests to avoid flakes across CI partitions.

_Remember to:_

## Testing plan

1. Spin up a TFE v2.0.1 instance and ran the acceptance tests against it.

## External links

[JIRA](https://hashicorp.atlassian.net/browse/TF-35679)
[RFC](https://hermes-sharepoint.hashicorp.services/document/01XOO7K4IWFH44KZAPDJC3QXNKIRWEZMBA)

1. `TESTARGS="-run TestAccTFESCIMSettings" ENABLE_TFE=1 envchain scim make testacc`
    image

2. `TESTARGS="-run TestAccTFESCIMSettingsDataSource" ENABLE_TFE=1 envchain scim make testacc`
    image

## Rollback Plan

we can revert this pr

NA


.github/workflows/ci.yml
```diff
@@ -56,8 +56,8 @@ jobs:
           # Run terminal cmd 'go help testflag' to learn more about -list flag
           # action.yml uses https://github.com/hashicorp-forge/go-test-split-action/blob/main/action.yml to split acceptance tests
           # which runs against all tests using the list arg
-          # lists_tests regex is used to skip the TestAccTFESAMLSettings_omnibus test suite for CI tests only
-          list_tests: "[^(TestAccTFESAMLSettings_omnibus)]"
+          # lists_tests regex is used to skip the TestAccTFESAMLSettings_omnibus and TestAccTFESCIMSettings_omnibus test suite for CI tests only
+          list_tests: "[^(TestAccTFE(SAML|SCIM)Settings_omnibus)]"
```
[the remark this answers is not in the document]
skj-skj: Good catch. `[^...]` is a negated character class in RE2 (no lookaheads available), by listing all characters from both test names in the set, those tests contain no character outside it and get skipped. Every other test has at least one character not in the set and matches normally. I've updated the regex to spell out both full test names for clarity. Verified locally.

internal/provider/resource_tfe_scim_settings_test.go
```diff
@@ -0,0 +1,420 @@
...
+						token, err := testAccConfiguredClient.Client.Admin.Settings.SCIM.Tokens.Create(
+							context.Background(), "tf-acc-test-scim-token",
+						)
+						if err != nil {
+							t.Fatalf("create SCIM token: %v", err)
+						}
+
+						siteAdminGroupID = createSCIMGroup(t, "tf-acc-site-admins", token.Token)
+						t.Setenv("TF_VAR_site_admin_group_scim_id", siteAdminGroupID)
+					},
```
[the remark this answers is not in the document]
skj-skj: fixed it. deleting scim group is not required as it will be deleted by backend when scim gets disabled.

internal/provider/resource_tfe_scim_settings_test.go
```diff
@@ -0,0 +1,420 @@
...
+				},
+				// Create a SCIM group out-of-band and link it via TF_VAR.
+				{
+					PreConfig: func() {
+						token, err := testAccConfiguredClient.Client.Admin.Settings.SCIM.Tokens.Create(
+							context.Background(), "tf-acc-test-scim-token",
+						)
+						if err != nil {
+							t.Fatalf("create SCIM token: %v", err)
+						}
```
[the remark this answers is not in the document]
skj-skj: fixed it.

internal/provider/resource_tfe_scim_settings.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// updateSCIMSettings applies the SCIM settings for Create and Update. The plan
+// is the source of truth: every field is sent on every call (schema defaults
+// populate fields the user omits). site_admin_group_scim_id sends JSON null
+// when empty (unlinks the group) and the raw value otherwise (links it).
+func (r *resourceTFESCIMSettings) updateSCIMSettings(ctx context.Context, m modelTFESCIMSettings) (*tfe.AdminSCIMSetting, error) {
+	var siteAdminGroupSCIMID jsonapi.NullableAttr[string]
+	if v := m.SiteAdminGroupSCIMID.ValueString(); v == "" {
+		siteAdminGroupSCIMID = tfe.NullString()
```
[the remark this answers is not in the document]
skj-skj: fixed it, handling `SiteAdminGroupSCIMID` unknown value. if value is not known before apply it will fail

internal/provider/resource_tfe_scim_settings.go
```diff
@@ -0,0 +1,220 @@
...
+	scimSettings, err := r.client.Admin.Settings.SCIM.Read(ctx)
+	if err != nil {
+		resp.Diagnostics.AddError("Error importing SCIM Settings", "Could not retrieve SCIM Settings, unexpected error: "+err.Error())
+		return
+	}
+
+	result := modelFromTFEAdminSCIMSettings(*scimSettings)
+	diags := resp.State.Set(ctx, &result)
+	resp.Diagnostics.Append(diags...)
+}
```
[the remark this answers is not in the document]
skj-skj: fixed it, import will fail if scim is disabled.

internal/provider/helper_test.go
```diff
@@ -358,3 +362,52 @@ func retryFn(maxRetries, secondsBetween int, f retryableFn) (any, error) {
...
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodPost,
+		fmt.Sprintf("https://%s/scim/v2/Groups", hostname), bytes.NewReader(body))
+	if err != nil {
+		t.Fatalf("build SCIM group request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/scim+json")
+	req.Header.Set("Authorization", "Bearer "+scimToken)
+
+	httpClient := &http.Client{Timeout: 30 * time.Second}
+	resp, err := httpClient.Do(req)
```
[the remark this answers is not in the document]
skj-skj: Not required, we use ngrok in CI so TLS is always valid, and the test suite never sets `TFE_SSL_SKIP_VERIFY`.

nikolasrieble (commented):
Review comments left inline.

internal/provider/resource_tfe_scim_settings.go
```diff
@@ -0,0 +1,233 @@
...
+	tflog.Debug(ctx, "Updating SCIM Settings")
+	scimSettings, err := r.updateSCIMSettings(ctx, m)
+	if err != nil {
+		resp.Diagnostics.AddError("Error creating SCIM Settings", "Could not set SCIM Settings, unexpected error: "+err.Error())
+		return
+	}
+
+	result := modelFromTFEAdminSCIMSettings(*scimSettings)
+	diags = resp.State.Set(ctx, &result)
+	resp.Diagnostics.Append(diags...)
```
nikolasrieble, line 156: If SCIM is disabled out-of-band and a user runs `terraform destroy`, the API call will likely error, failing the destroy. The Terraform framework expects `Delete` to succeed when the resource is already gone. Could we check if SCIM is already disabled (either read first or catch the specific error) and return successfully in that case?

nikolasrieble (commented):
A few findings — one functional, one question, one nit.

.github/workflows/ci.yml
```diff
@@ -56,8 +56,8 @@ jobs:
           # Run terminal cmd 'go help testflag' to learn more about -list flag
           # action.yml uses https://github.com/hashicorp-forge/go-test-split-action/blob/main/action.yml to split acceptance tests
           # which runs against all tests using the list arg
-          # lists_tests regex is used to skip the TestAccTFESAMLSettings_omnibus test suite for CI tests only
```
nikolasrieble, line 59: Question: is this regex working as intended? `[^(TestAccTFESAMLSettings_omnibus|TestAccTFESCIMSettings_omnibus)]` is a character class negation — it matches strings whose first character is NOT in the set of characters `(`, `T`, `e`, `s`, etc. It's not a negative lookahead. Since all `TestAcc*` tests start with `T` (which is in the negated set), this would seemingly exclude everything. The same pattern existed before for SAML, so it's presumably either interpreted non-standardly by go-test-split-action, or it's been silently broken. Could you confirm how the action consumes this value?

internal/provider/data_source_scim_settings.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Metadata returns the data source type name.
+func (d *dataSourceTFESCIMSettings) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_scim_settings"
+}
+
+// Schema defines the schema for the data source.
+func (d *dataSourceTFESCIMSettings) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
```
nikolasrieble, line 50: Nit: both the resource and data source schemas are missing a top-level `Description` on `schema.Schema{}`, which leaves the registry docs description blank. Something like `"Reads the current SCIM provisioning settings for Terraform Enterprise."` would help.

internal/provider/resource_tfe_scim_settings_test.go
```diff
@@ -0,0 +1,377 @@
...
+				},
+				// Omitting `paused` reverts to the default (false).
+				{
+					Config: testAccTFESCIMSettings_enable(),
+					Check: resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttr("tfe_scim_settings.enable_scim", "enabled", "true"),
+						resource.TestCheckResourceAttr("tfe_scim_settings.enable_scim", "paused", "false"),
+					),
+				},
+				// Explicitly unpause.
```
nikolasrieble: Every test is code that needs to be maintained. This step (explicit `paused = false`) produces the same outcome as step 3 (omit `paused`, default kicks in). Both verify `paused = "false"` in state. The only difference is whether the framework's default mechanism or an explicit value produces the result — but that's testing the Terraform plugin framework's guarantee, not your provider code.

**Proposal:** Remove this step. Step 3 already covers the meaningful case (config change triggers revert to default).

**Argument:** No unique provider code path is exercised. If the framework's default mechanism broke, step 3 would catch it. This step adds runtime and maintenance cost without additional regression protection.

internal/provider/resource_tfe_scim_settings_test.go
```diff
@@ -0,0 +1,377 @@
...
+				},
+				// Re-apply same config: should be a no-op (no perpetual diff).
+				{
+					Config:   testAccTFESCIMSettings_withSiteAdminGroup(),
+					PlanOnly: true,
+				},
+				// Import round-trips the linked group through state.
+				{
+					ResourceName:      "tfe_scim_settings.enable_scim",
+					ImportState:       true,
```
nikolasrieble, line 144: This step re-links the same group that was already linked in step 2 (after unlinking in step 5). The provider code path exercised here (Update with a non-empty `site_admin_group_scim_id`) is identical to step 2.

**Proposal:** Remove this step.

**Argument:** Step 2 already proves linking works. Step 7 (switching from group A to group B) covers the more interesting case of updating to a *different* non-empty value. Re-linking the same group after unlinking doesn't exercise any new provider logic — it's the same Update call with the same payload shape.

internal/provider/resource_tfe_scim_settings_test.go
```diff
@@ -0,0 +1,377 @@
...
+		resource.Test(t, resource.TestCase{
+			PreCheck:                 func() { testAccPreCheck(t) },
+			ProtoV6ProviderFactories: testAccMuxedProviders,
+			CheckDestroy:             testAccTFESCIMSettingsDestroy,
+			Steps: []resource.TestStep{
+				// Enable SCIM.
+				{
+					Config: testAccTFESCIMSettings_enable(),
+				},
+				// Import by the fixed "scim" ID.
```
nikolasrieble, line 213: This standalone import sub-test is debatable. The import step in "SCIM settings site admin group" (step 4) already tests import with a group linked. This tests import *without* a group linked — a simpler state shape.

**Argument for removing:** Same provider code path (`ImportState` → Read → set state). The presence or absence of a linked group doesn't change the import logic.

**Argument for keeping:** It's a simpler base case that's easier to debug in isolation if import breaks, and it runs independently of the more complex site-admin-group test.

Leaving this to your judgement — if you keep it, the cost is minor (one extra enable/disable cycle against the instance).

.github/workflows/ci.yml
skj-skj: yep, the regex working.

internal/provider/resource_tfe_scim_settings.go
skj-skj: Tested this -> calling Delete on an already-disabled SCIM returns no error, the API is idempotent. Added a test case `destroy when SCIM already disabled out-of-band` to cover this scenario going forward.

internal/provider/data_source_scim_settings.go
skj-skj: fixed, added the Description in `Schema`

internal/provider/resource_tfe_scim_settings_test.go
skj-skj: fixed, removed the explicit `unpaused` step

internal/provider/resource_tfe_scim_settings_test.go
skj-skj: fixed, reordered the steps so `switch to group B` now follows directly after `link group A`, restoring the `non-null` -> `non-null` transition. The clear and omit steps follow after.

internal/provider/resource_tfe_scim_settings_test.go
skj-skj: I would like to keep it. to make import failure easy to debug

nikolasrieble (approved):
Thank you for your work! I do not see any issue.

internal/provider/data_source_scim_settings.go
```diff
@@ -0,0 +1,110 @@
...
+func (d *dataSourceTFESCIMSettings) Read(ctx context.Context, _ datasource.ReadRequest, resp *datasource.ReadResponse) {
+	s, err := d.client.Admin.Settings.SCIM.Read(ctx)
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to read SCIM settings", err.Error())
+		return
+	}
+
+	// Set state
+	diags := resp.State.Set(ctx, &modelDataTFESCIMSettings{
+		ID:                        types.StringValue("scim"),
```
Uk1288: 💅🏾 should probably read from `s.ID`

internal/provider/resource_tfe_scim_settings.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	s, err := r.client.Admin.Settings.SCIM.Update(ctx, tfe.AdminSCIMSettingUpdateOptions{
+		Enabled:              tfe.Bool(true),
+		Paused:               m.Paused.ValueBoolPointer(),
+		SiteAdminGroupSCIMID: siteAdminGroupSCIMID,
+	})
+
+	if err != nil {
+		return s, fmt.Errorf("failed to update SCIM Settings: %w", err)
```
Uk1288: should probably say:
```suggestion
		return s, fmt.Errorf("failed to set SCIM Settings: %w", err)
```
since it is used by both create & update

internal/provider/resource_tfe_scim_settings.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	result := modelFromTFEAdminSCIMSettings(*scimSettings)
+	diags = resp.State.Set(ctx, &result)
+	resp.Diagnostics.Append(diags...)
+}
+
+func (r *resourceTFESCIMSettings) Delete(ctx context.Context, _ resource.DeleteRequest, resp *resource.DeleteResponse) {
+	tflog.Debug(ctx, "Delete SCIM Settings")
+	err := r.client.Admin.Settings.SCIM.Delete(ctx)
+	if err != nil {
```
Uk1288, line 181: if delete results in not-found, we can make this more graceful by removing from state instead of erroring out. See examples in the repo

internal/provider/resource_tfe_scim_settings.go
```diff
@@ -0,0 +1,234 @@
...
+
+// resourceTFESCIMSettings implements the tfe_scim_settings resource type
+type resourceTFESCIMSettings struct {
+	client *tfe.Client
+}
+
+// modelFromTFEAdminSCIMSettings builds a modelTFESCIMSettings struct from a tfe.AdminSCIMSetting value
+func modelFromTFEAdminSCIMSettings(v tfe.AdminSCIMSetting) modelTFESCIMSettings {
+	m := modelTFESCIMSettings{
+		ID:                        types.StringValue("scim"),
```
Uk1288: 💅🏾 set from `v.ID`

Uk1288 (commented):
Looking good!

Minor cleanup comments.

internal/provider/resource_tfe_scim_settings.go
skj-skj: thanks for the suggestion! I looked into this but I think we should keep the current behavior. The API returns [404 for unauthorized requests](https://developer.hashicorp.com/terraform/enterprise/api-docs/admin/scim-settings#disable-scim), so 404 here likely means expired or invalid creds, silently removing from state would hide the read issue. Also the Delete API is idempotent, calling it again after disabling the SCIM will return 200 so there isn't true "not found" scenario.

internal/provider/data_source_scim_settings.go
skj-skj: fixed it, thx

internal/provider/resource_tfe_scim_settings.go
skj-skj: fixed it, thx

Uk1288 (approved):

PR #2088, opened 2026-06-05
[TF-35683]Adds SCIM resources to Team Data Source for TFE.

kkavish:
## Description

Team data source must output SCIM attributes when SCIM is enabled in TFE.

_Remember to:_

## Testing plan

1.  You can just run the test case `TestAccTFESCIMTeamDataSource_omnibus`
2. Or you can deploy TFE with SAML and SCIM enabled.
3. Create a TFE Team, create a SCIM group using Admin APIs.
4. Configure TFE provider and try to fetch the TFE team with external ID and access the SCIM attributes.

TFE tests - https://github.com/hashicorp/terraform-provider-tfe/actions/runs/27018220194/job/79741292194

## External links

[TF-35683](https://hashicorp.atlassian.net/browse/TF-35683)

```
$ TFE_HOSTNAME=<REDACTED> \
TFE_TOKEN=<REDACTED> \
TFE_ORGANIZATION=db-check-test \
ENABLE_TFE=1 \
TF_ACC=1 \
go test ./internal/provider/ -v -run TestAccTFETeamDataSource_scim -timeout 30m 2>&1
=== RUN   TestAccTFETeamDataSource_scim
=== RUN   TestAccTFETeamDataSource_scim/SCIM_attributes_across_full_lifecycle
--- PASS: TestAccTFETeamDataSource_scim (11.83s)
    --- PASS: TestAccTFETeamDataSource_scim/SCIM_attributes_across_full_lifecycle (11.83s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/provider	12.669s
```

## Rollback Plan

Revert the PR.

image

[TF-35683]: https://hashicorp.atlassian.net/browse/TF-35683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/data_source_team_test.go
+++ b/internal/provider/data_source_team_test.go
@@ -145,6 +145,10 @@ func TestAccTFESCIMTeamDataSource_omnibus(t *testing.T) {
 					Check: resource.ComposeAggregateTestCheckFunc(
 						resource.TestCheckResourceAttrSet("data.tfe_team.test", "id"),
 						resource.TestCheckResourceAttr("data.tfe_team.test", "name", teamName),
+						resource.TestCheckNoResourceAttr("data.tfe_team.test", "scim_linked"),
+						resource.TestCheckNoResourceAttr("data.tfe_team.test", "scim_group_name"),
+						resource.TestCheckNoResourceAttr("data.tfe_team.test", "scim_sync_paused"),
+						resource.TestCheckNoResourceAttr("data.tfe_team.test", "scim_updated_at"),
 						// Capture the team external ID for use in step 3.
 						func(s *terraform.State) error {
 							rs, ok := s.RootModule().Resources["tfe_team.test"]
@@ -162,6 +166,8 @@ func TestAccTFESCIMTeamDataSource_omnibus(t *testing.T) {
 				{
 					Config: testAccTFETeamDataSourceConfig_scimEnabled(teamName, org.Name),
 					Check: resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttrSet("data.tfe_team.test", "id"),
+						resource.TestCheckResourceAttr("data.tfe_team.test", "name", teamName),
 						resource.TestCheckResourceAttr("data.tfe_team.test", "scim_linked", "false"),
 						resource.TestCheckNoResourceAttr("data.tfe_team.test", "scim_group_name"),
 						resource.TestCheckResourceAttr("data.tfe_team.test", "scim_sync_paused", "false"),
@@ -194,6 +200,8 @@ func TestAccTFESCIMTeamDataSource_omnibus(t *testing.T) {
 					},
 					Config: testAccTFETeamDataSourceConfig_scimEnabled(teamName, org.Name),
 					Check: resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttrSet("data.tfe_team.test", "id"),
+						resource.TestCheckResourceAttr("data.tfe_team.test", "name", teamName),
 						resource.TestCheckResourceAttr("data.tfe_team.test", "scim_linked", "true"),
 						resource.TestCheckResourceAttr("data.tfe_team.test", "scim_group_name", teamName),
 						resource.TestCheckResourceAttr("data.tfe_team.test", "scim_sync_paused", "false"),
--- a/internal/provider/data_source_team_test.go
+++ b/internal/provider/data_source_team_test.go
@@ -4,12 +4,15 @@
 package provider
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
+	"os"
 	"testing"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
 func TestAccTFETeamDataSource_basic(t *testing.T) {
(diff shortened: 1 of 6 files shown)


internal/provider/data_source_team_test.go
```diff
@@ -99,3 +102,132 @@ data "tfe_team" "sso_team" {
...
+							t.Fatalf("create SCIM token: %v", err)
+						}
+						t.Cleanup(func() {
+							_ = testAccConfiguredClient.Client.Admin.Settings.SCIM.Tokens.Delete(
+								context.Background(), token.ID)
+						})
+
+						// No explicit group cleanup: disabling SCIM (CheckDestroy) removes all groups.
+						scimGroupID := createSCIMGroup(t, teamName, token.Token)
+						linkSCIMGroupToTeam(t, teamID, scimGroupID)
```
skj-skj: we can use `go-tfe` to link scim group to tfe teams.
```go
import (
  ...
  ...
)
...
err := testAccConfiguredClient.Client.Admin.Settings.SCIM.SCIMGroupMappings.Create(ctx, teamID, &tfe.AdminSCIMGroupMappingCreateOptions{SCIMGroupID: scimGroupID})
```

internal/provider/data_source_team_test.go
```diff
@@ -99,3 +102,132 @@ data "tfe_team" "sso_team" {
...
+// same singleton state.
+//
+// Should this test name ever change, you will also need to update the regex in ci.yml.
+func TestAccTFESCIMTeamDataSource_omnibus(t *testing.T) {
+	skipIfCloud(t)
+
+	t.Run("SCIM attributes across full lifecycle", func(t *testing.T) {
+		teamName := "tf-acc-scim-team-" + randomString(t)
+		org := os.Getenv("TFE_ORGANIZATION")
+
```
skj-skj, line 131: the acc test case [gets failed](https://github.com/hashicorp/terraform-provider-tfe/actions/runs/27012102657/job/79719253752) as org was not found, you can use `createOrganization`.

I tried `createBusinessOrganization` but it failed in tfe as I got this error
```sh
Cannot upgrade an organization's subscription when enterprise is enabled. Set ENABLE_TFE=0 to run.
```


```suggestion
		teamName := "tf-acc-scim-team-" + randomString(t)
		org, orgCleanup := createOrganization(t, client, tfe.OrganizationCreateOptions{
		    Name:  tfe.String("tst-" + randomString(t)),
		    Email: tfe.String(fmt.Sprintf("%s@tfe.local", randomString(t))),
	    })
		t.Cleanup(orgCleanup)
```

internal/provider/data_source_team_test.go
```diff
@@ -99,3 +102,132 @@ data "tfe_team" "sso_team" {
...
+		// teamID is captured from Terraform state after step 1 so that step 3
+		// PreConfig can call linkSCIMGroupToTeam with the correct external ID.
+		var teamID string
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:                 func() { testAccPreCheck(t) },
+			ProtoV6ProviderFactories: testAccMuxedProviders,
+			Steps: []resource.TestStep{
```
skj-skj, line 140: `CheckDestroy` is missing, you can use `testAccTFESCIMSettingsDestroy` from scim settings test


```suggestion
		resource.Test(t, resource.TestCase{
			PreCheck:                 func() { testAccPreCheck(t) },
			ProtoV6ProviderFactories: testAccMuxedProviders,
			CheckDestroy:             testAccTFESCIMSettingsDestroy,
			Steps: []resource.TestStep{
```

skj-skj (changes requested):
looks good, requesting few changes in test

skj-skj (dismissed):
🚀

internal/provider/data_source_team_test.go
kkavish: https://github.com/hashicorp/terraform-provider-tfe/actions/runs/27018220194/job/79741292194 thanks for showing me how to run TFE tests! Here is a successful run.

internal/provider/helper_test.go
```diff
@@ -24,6 +24,8 @@ const RunTasksURLEnvName = "RUN_TASKS_URL"
 const RunTasksHMACKeyEnvName = "RUN_TASKS_HMAC"
 const EnableHYOKEnvName = "ENABLE_HYOK"
 
+const TFEScimGroupAPI = "https://%s/scim/v2/Groups"
```
shwetamurali: `TFESCIMGroupAPI` for consistency with SCIM used elsewhere

website/docs/d/team.html.markdown
```diff
@@ -30,4 +30,8 @@ The following arguments are supported:
 In addition to all arguments above, the following attributes are exported:
 
 * `id` - The ID of the team.
-* `sso_team_id` - (Optional) The [SSO Team ID](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/single-sign-on#team-names-and-sso-team-ids) of the team, if it has been defined
+* `sso_team_id` - (Optional) The [SSO Team ID](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/single-sign-on#team-names-and-sso-team-ids) of the team, if it has been defined.
+* `scim_linked` - Whether the team is linked to a SCIM group. Only populated when SCIM is enabled on the TFE instance.
```
shwetamurali: There's no way for a user to distinguish between:
- Older TFE instance without SCIM support → scim_linked = false (nil pointer, d.Set never called, zero value)
- Newer TFE with SCIM enabled but team not linked → scim_linked = false (explicit false from API)
From the user's perspective in Terraform state, both cases produce false. If this distinction matters, a `scim_enabled` attribute or making the booleans nullable (which requires Plugin Framework) would be needed. This may be acceptable as-is since the PR targets TFE users who know their SCIM status, so it is up to you.

internal/provider/data_source_team_test.go
```diff
@@ -99,3 +102,146 @@ data "tfe_team" "sso_team" {
...
+		// teamID is captured from Terraform state after step 1 so that step 3
+		// PreConfig can call SCIMGroupMappings.Create with the correct external ID.
+		var teamID string
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:                 func() { testAccPreCheck(t) },
+			ProtoV6ProviderFactories: testAccMuxedProviders,
+			CheckDestroy:             testAccTFESCIMSettingsDestroy,
+			Steps: []resource.TestStep{
+				// Step 1: No SCIM at all. Verify the data source reads successfully
```
shwetamurali, line 141: The test only checks id and name. Adding assertions like `TestCheckResourceAttr(..., "scim_linked", "false")` or `TestCheckNoResourceAttr(..., "scim_group_name")` would strengthen confidence that the nil guards work when the API returns no SCIM fields at all

website/docs/d/team.html.markdown
kkavish: @shwetamurali that's a very nice point. Thanks for bringing it up. If SCIM is not enabled (disabled or not supported) SCIM fields are `nil` and we do not set them in the state.

```
if team.SCIMLinked != nil {
		d.Set("scim_linked", *team.SCIMLinked)
	}
	if team.SCIMGroupName != nil {
		d.Set("scim_group_name", *team.SCIMGroupName)
	}
	if team.SCIMSyncPaused != nil {
		d.Set("scim_sync_paused", *team.SCIMSyncPaused)
	}
	if team.SCIMUpdatedAt != nil {
		d.Set("scim_updated_at", team.SCIMUpdatedAt.Format(time.RFC3339))
	}
```

So, they can't have zero-values if SCIM is not turned on or not supported.

If a user sees a zero-value, that means SCIM is enabled. I think we need to document it better here.

skj-skj (dismissed):

website/docs/d/team.html.markdown
shwetamurali: Thanks for clarifying, makes sense. The nil guard approach makes a clean separation. Agreed that a note making this explicit for users would be helpful.

website/docs/d/team.html.markdown
```diff
@@ -30,4 +30,8 @@ The following arguments are supported:
 In addition to all arguments above, the following attributes are exported:
 
 * `id` - The ID of the team.
-* `sso_team_id` - (Optional) The [SSO Team ID](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/single-sign-on#team-names-and-sso-team-ids) of the team, if it has been defined
+* `sso_team_id` - (Optional) The [SSO Team ID](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/single-sign-on#team-names-and-sso-team-ids) of the team, if it has been defined.
+* `scim_linked` - Whether the team is linked to a SCIM group. Only populated when SCIM is enabled on the TFE instance. If not present, SCIM is either not supported or enabled in the TFE instance.
```
shwetamurali: NitL This reads like "not supported OR enabled" which is the opposite of the intent. It should probably say "not supported or not enabled". (Or rephrase to: "If not present, SCIM is not enabled or not supported on the TFE instance.")

website/docs/d/team.html.markdown
```diff
@@ -30,4 +30,8 @@ The following arguments are supported:
...
+* `scim_group_name` - The display name of the SCIM group linked to this team. Only populated when SCIM is enabled and the team is linked to a SCIM group. If not present, SCIM is either not supported or enabled in the TFE instance.
```
shwetamurali: The Step 2 test asserts `scim_sync_paused = "false"` when SCIM is enabled but the team is NOT linked. So `scim_sync_paused` behaves like `scim_linked` it's populated whenever SCIM is enabled, regardless of linking. Its description should say "Only populated when SCIM is enabled on the TFE instance" (same as `scim_linked`)

shwetamurali (approved):

PR #2090, opened 2026-06-08
Adds `r/tfe_scim_group_mapping`

skj-skj:
## Description

This PR adds `r/tfe_scim_group_mapping` to map a SCIM group to a team in Terraform Enterprise. A team can be mapped to at most one SCIM group, and its membership is then managed by your IdP through SCIM provisioning.

Changing `team_id` or `scim_group_id` forces a new mapping, since the underlying API only supports updating the `paused` state in place.

Because the Teams API only returns the linked group's *name* (not its ID), `Read` resolves `scim_group_id` by looking the group up by name. This is handled by a new `findSCIMGroupByName` helper, which pages through groups and matches exactly (case-insensitive).

This PR also adds acceptance tests. Since SCIM settings are a shared singleton, the coverage is grouped into a single omnibus test to avoid race conditions.

Docs for `r/tfe_scim_group_mapping` are included as well.

_Remember to:_

## Testing plan

1.  Spin up a TFE instance with SAML + SCIM enabled and run acceptance tests against it.

## External links

[JIRA](https://hashicorp.atlassian.net/browse/TF-35682)
[RFC](https://hermes-sharepoint.hashicorp.services/document/01XOO7K4IWFH44KZAPDJC3QXNKIRWEZMBA)

1. `TESTARGS="-run TestAccTFESCIMGroupMapping" ENABLE_TFE=1 envchain scim make testacc`
    Untitled

2. [Nightly TFE Tests #1188](https://github.com/hashicorp/terraform-provider-tfe/actions/runs/27203237365)
    Untitled 2

## Rollback Plan

we can revert this pr

NA

### DIFF
--- a/internal/provider/data_source_scim_group.go
+++ b/internal/provider/data_source_scim_group.go
@@ -113,7 +113,7 @@ func (d *dataSourceTFESCIMGroup) Read(ctx context.Context, req datasource.ReadRe
 	// matching behavior stays consistent with the SCIM group mapping resource.
 	match, err := findSCIMGroupByName(ctx, d.client, name)
 	if err != nil {
-		resp.Diagnostics.AddError("Unable to list SCIM groups", err.Error())
+		resp.Diagnostics.AddError("Error reading SCIM group", err.Error())
 		return
 	}
 
--- a/internal/provider/data_source_scim_group.go
+++ b/internal/provider/data_source_scim_group.go
@@ -105,34 +105,16 @@ func (d *dataSourceTFESCIMGroup) Read(ctx context.Context, req datasource.ReadRe
 	}
 	name := data.Name.ValueString()
 
-	options := &tfe.AdminSCIMGroupListOptions{
-		// ?q= is a fuzzy substring match used here only as a server-side
-		// prefilter; we still narrow to an exact, case-insensitive match below.
-		Query: name,
-	}
-
 	tflog.Debug(ctx, "Listing SCIM groups", map[string]any{
 		"name": name,
 	})
 
-	// Keep the case-insensitive exact match (at most one) and stop
-	// paginating as soon as we find it.
-	var match *tfe.AdminSCIMGroup
-	for {
-		list, err := d.client.Admin.Settings.SCIM.Groups.List(ctx, options)
-		if err != nil {
-			resp.Diagnostics.AddError("Unable to list SCIM groups", err.Error())
-			return
-		}
-
-		if matched := filterExactSCIMGroups(list.Items, name); len(matched) > 0 {
-			match = matched[0]
-			break
-		}
-		if list.Pagination == nil || list.CurrentPage >= list.TotalPages {
-			break
-		}
-		options.PageNumber = list.NextPage
+	// Reuse the shared helper so the pagination and exact, case-insensitive
+	// matching behavior stays consistent with the SCIM group mapping resource.
+	match, err := findSCIMGroupByName(ctx, d.client, name)
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to list SCIM groups", err.Error())
+		return
 	}
 
 	if match == nil {
--- a/internal/provider/data_source_scim_group.go
+++ b/internal/provider/data_source_scim_group.go
@@ -113,7 +113,7 @@ func (d *dataSourceTFESCIMGroup) Read(ctx context.Context, req datasource.ReadRe
 	// matching behavior stays consistent with the SCIM group mapping resource.
 	match, err := findSCIMGroupByName(ctx, d.client, name)
 	if err != nil {
-		resp.Diagnostics.AddError("Unable to list SCIM groups", err.Error())
+		resp.Diagnostics.AddError("Unable to find SCIM group", err.Error())
 		return
 	}
 
--- a/internal/provider/data_source_scim_group.go
+++ b/internal/provider/data_source_scim_group.go
@@ -113,7 +113,7 @@ func (d *dataSourceTFESCIMGroup) Read(ctx context.Context, req datasource.ReadRe
 	// matching behavior stays consistent with the SCIM group mapping resource.
 	match, err := findSCIMGroupByName(ctx, d.client, name)
 	if err != nil {
-		resp.Diagnostics.AddError("Unable to find SCIM group", err.Error())
+		resp.Diagnostics.AddError("Unable to list SCIM groups", err.Error())
 		return
 	}
 
--- a/internal/provider/resource_tfe_scim_group_mapping.go
+++ b/internal/provider/resource_tfe_scim_group_mapping.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
@@ -19,6 +18,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
+
+	tfe "github.com/hashicorp/go-tfe"
 )
 
 type modelTFESCIMGroupMapping struct {
--- a/internal/provider/resource_tfe_scim_group_mapping.go
+++ b/internal/provider/resource_tfe_scim_group_mapping.go
@@ -1,0 +1,313 @@
+// Copyright IBM Corp. 2018, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package provider
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+type modelTFESCIMGroupMapping struct {
+	ID          types.String `tfsdk:"id"`
+	TeamID      types.String `tfsdk:"team_id"`
+	SCIMGroupID types.String `tfsdk:"scim_group_id"`
+	Paused      types.Bool   `tfsdk:"paused"`
+}
+
+// resourceTFESCIMGroupMapping implements the tfe_scim_group_mapping resource type.
+type resourceTFESCIMGroupMapping struct {
+	client *tfe.Client
+}
+
+var (
+	_ resource.Resource                = &resourceTFESCIMGroupMapping{}
+	_ resource.ResourceWithConfigure   = &resourceTFESCIMGroupMapping{}
+	_ resource.ResourceWithImportState = &resourceTFESCIMGroupMapping{}
+)
+
+// NewSCIMGroupMappingResource is a resource function for the framework provider.
+func NewSCIMGroupMappingResource() resource.Resource {
+	return &resourceTFESCIMGroupMapping{}
+}
+
+// Metadata implements resource.Resource
+func (r *resourceTFESCIMGroupMapping) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_scim_group_mapping"
+}
+
+// Configure implements resource.ResourceWithConfigure
+func (r *resourceTFESCIMGroupMapping) Configure(_ context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	// Early exit if provider is not properly configured (i.e. we're only validating config or something)
+	if req.ProviderData == nil {
+		return
+	}
+
+	client, ok := req.ProviderData.(ConfiguredClient)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected resource Configure type",
+			fmt.Sprintf("Expected tfe.ConfiguredClient, got %T. This is a bug in the tfe provider, so please report it on GitHub.", req.ProviderData),
+		)
+		return
+	}
+	r.client = client.Client
+}
+
+// Schema implements resource.Resource
+func (r *resourceTFESCIMGroupMapping) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Maps a SCIM group to a team in Terraform Enterprise. A team can be mapped to at most one SCIM group.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "The ID of the SCIM group mapping. Since a team can only be mapped to one SCIM group, this is the same as `team_id`.",
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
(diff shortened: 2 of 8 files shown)


internal/provider/scim_helpers.go
```diff
@@ -24,3 +26,31 @@ func filterExactSCIMGroups(groups []*tfe.AdminSCIMGroup, name string) []*tfe.Adm
...
 }
+
+// findSCIMGroupByName returns the SCIM group whose name matches exactly
+// (case-insensitive), paging as needed. Returns (nil, nil) if none match, so
+// the caller can craft its own "not found" message. ?q= only prefilters on the
+// server; filterExactSCIMGroups does the real matching.
+func findSCIMGroupByName(ctx context.Context, client *tfe.Client, name string) (*tfe.AdminSCIMGroup, error) {
+	options := &tfe.AdminSCIMGroupListOptions{
+		Query: name,
+	}
```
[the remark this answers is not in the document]
skj-skj: fixed, reusing `findSCIMGroupByName` func

internal/provider/scim_helpers.go
```diff
@@ -24,3 +26,31 @@ func filterExactSCIMGroups(groups []*tfe.AdminSCIMGroup, name string) []*tfe.Adm
...
+		}
+
+		if list.Pagination == nil || list.CurrentPage >= list.TotalPages {
+			break
+		}
+		options.PageNumber = list.NextPage
+	}
+
+	return nil, nil
+}
```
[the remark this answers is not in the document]
skj-skj: good catch, added the test cases

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// Save state now that the mapping exists, so a failed pause below doesn't
+	// orphan it.
+	paused := plan.Paused.ValueBool()
+	result := modelTFESCIMGroupMapping{
+		ID:          types.StringValue(teamID),
+		TeamID:      types.StringValue(teamID),
+		SCIMGroupID: plan.SCIMGroupID,
+		Paused:      types.BoolValue(false),
+	}
```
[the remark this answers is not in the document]
skj-skj: Create and Update only run during the apply phase, and by that point Terraform has already resolved every unknown value from the plan into a concrete value.

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,313 @@
...
+	var plan modelTFESCIMGroupMapping
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	teamID := plan.TeamID.ValueString()
+	paused := plan.Paused.ValueBool()
+
```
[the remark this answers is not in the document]
skj-skj: Create and Update only run during the apply phase, and by that point Terraform has already resolved every unknown value from the plan into a concrete value.

internal/provider/resource_tfe_scim_group_mapping_test.go
```diff
@@ -0,0 +1,552 @@
...
+							"tfe_scim_group_mapping.test", "id",
+							"tfe_team.test", "id",
+						),
+					),
+				},
+				// Re-apply must be a no-op.
+				{
+					Config:   testAccTFESCIMGroupMapping_basic(org.Name, teamName, tokenDescription, groupName),
+					PlanOnly: true,
+				},
```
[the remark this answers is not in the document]
skj-skj: as per the godoc of [terraform-plugin-testing](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep.PlanOnly) 

> // PlanOnly can be set to only run `plan` with this configuration, and not
	// actually apply it. This is useful for ensuring config changes result in
	// no-op plans

since a PlanOnly step already fails on a non-empty plan. I'll leave it as-is

internal/provider/data_source_scim_group.go
```diff
@@ -105,34 +105,16 @@ func (d *dataSourceTFESCIMGroup) Read(ctx context.Context, req datasource.ReadRe
...
-		if list.Pagination == nil || list.CurrentPage >= list.TotalPages {
-			break
-		}
-		options.PageNumber = list.NextPage
+	// Reuse the shared helper so the pagination and exact, case-insensitive
+	// matching behavior stays consistent with the SCIM group mapping resource.
+	match, err := findSCIMGroupByName(ctx, d.client, name)
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to find SCIM group", err.Error())
+		return
```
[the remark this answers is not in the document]
skj-skj: fixed updated the err msg

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,313 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package provider
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	tfe "github.com/hashicorp/go-tfe"
```
kkavish: this should be in a separate section at the bottom.

1. Go internal deps
2. external deps
3. Hashi deps

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,313 @@
...
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
+	groupName := ""
+	if team.SCIMGroupName != nil {
+		groupName = *team.SCIMGroupName
+	}
+
+	scimGroupID, err := r.resolveSCIMGroupID(ctx, groupName)
```
kkavish: If group name is nil it is set to empty, and then we try to resolve it, should we not fail instead of trying to resolve?

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,313 @@
...
+	var plan modelTFESCIMGroupMapping
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	teamID := plan.TeamID.ValueString()
+
+	tflog.Debug(ctx, fmt.Sprintf("Creating SCIM group mapping for team %s", teamID))
```
kkavish: scim group missing from the error log

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,313 @@
...
+	err := r.client.Admin.Settings.SCIM.SCIMGroupMappings.Create(ctx, teamID, &tfe.AdminSCIMGroupMappingCreateOptions{
+		SCIMGroupID: plan.SCIMGroupID.ValueString(),
+	})
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error creating SCIM group mapping",
+			fmt.Sprintf("Could not create SCIM group mapping for team %s, unexpected error: %s", teamID, err.Error()),
```
kkavish: missing scim group in the error.

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,313 @@
...
+		SCIMGroupID: plan.SCIMGroupID,
+		Paused:      types.BoolValue(false),
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, &result)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Create always starts unpaused, so pause it now if requested.
+	if paused {
```
kkavish, line 211: So, the problem is - create will only succeed once the group -> team sync is successful. Also, if there are updates to group while this executes, customers will see newer updates in the team. I don't want a logic change here, just a callout in the docs to set the expectations.

internal/provider/scim_helpers_test.go
```diff
@@ -4,10 +4,13 @@
 package provider
 
 import (
+	"context"
+	"errors"
 	"testing"
 
 	tfe "github.com/hashicorp/go-tfe"
```
kkavish: fix imports please

internal/provider/resource_tfe_scim_group_mapping_test.go
```diff
@@ -0,0 +1,552 @@
...
+					Config:      testAccTFESCIMGroupMappingEmptyTeamID(),
+					ExpectError: lengthErr,
+					PlanOnly:    true,
+				},
+				{
+					Config:      testAccTFESCIMGroupMappingEmptySCIMGroupID(),
+					ExpectError: lengthErr,
+					PlanOnly:    true,
+				},
+			},
```
kkavish, line 437: Does it make sense to add initial pause `true` scenario here?

website/docs/r/scim_group_mapping.html.markdown
```diff
@@ -0,0 +1,100 @@
+---
+layout: "tfe"
+page_title: "Terraform Enterprise: tfe_scim_group_mapping"
+description: |-
+  Maps a SCIM group to a team.
+---
+
+# tfe_scim_group_mapping
+
+Use this resource to map a SCIM group to a team in Terraform Enterprise. When a
```
kkavish:
```suggestion
Use this resource to map a team in Terraform Enterprise to a SCIM group. When a
```

internal/provider/scim_helpers_test.go
```diff
@@ -53,3 +56,246 @@ func TestFilterExactSCIMGroups(t *testing.T) {
...
+		}
+
+		group, err := findSCIMGroupByName(ctx, newSCIMGroupsTestClient(fake), testSCIMGroupName)
+		require.Error(t, err)
+		assert.Nil(t, group)
+		assert.Contains(t, err.Error(), "pagination did not advance")
+		// Only one request was made; the loop bailed instead of repeating.
+		assert.Len(t, fake.calls, 1)
+	})
+}
```
kkavish, line 321: Missing test: Empty group name.

website/docs/r/scim_group_mapping.html.markdown
```diff
@@ -0,0 +1,100 @@
...
+must already be provisioned into Terraform Enterprise by your IdP before the
+mapping can be created.
+
+## Example Usage
+
+The SCIM group you map to must already exist in Terraform Enterprise. Groups are
+created by your IdP after SCIM provisioning is enabled, so this typically follows
+a workflow where SCIM is enabled first, the IdP pushes the group, and then the
+mapping is created. Use the [`tfe_scim_group`](../d/scim_group.html) data source
+to look up the group's ID by name:
```
kkavish, line 27: 👍

kkavish (commented):
@skj-skj Thanks for this contribution, I have some comments (most of them minor). After we have addressed them, this should be good to go.

internal/provider/resource_tfe_scim_group_mapping.go
skj-skj: yeah, the fail logic is actually handled inside `resolveSCIMGroupID`, it returns an error right away if the name is empty, so we never resolve it.

internal/provider/resource_tfe_scim_group_mapping.go
skj-skj: good catch, fixed it

internal/provider/resource_tfe_scim_group_mapping.go
skj-skj: good catch, fixed it

internal/provider/resource_tfe_scim_group_mapping.go
skj-skj: thx for the callout, I have added a [note](https://github.com/hashicorp/terraform-provider-tfe/pull/2090/changes/57a2164e845a70810a95fdeb3e0ea31e395e0f64#diff-7bf1edb83f41ca2cae25000ec997a2e231fd0672094ac774e113b3efc8ab1637R77-R84) in the docs

internal/provider/scim_helpers_test.go
skj-skj: fixed, added the [test case](https://github.com/hashicorp/terraform-provider-tfe/pull/2090/changes/3375ac39914acf86e1398fd46059d0708a6ff131#diff-e67767ee3e12a2dcd63633d1457fe21f934f80a96c76b27527b42aabbc46876dR229-R232)

internal/provider/resource_tfe_scim_group_mapping_test.go
skj-skj: yep, it makes sense, added the [test case](https://github.com/hashicorp/terraform-provider-tfe/pull/2090/changes/3375ac39914acf86e1398fd46059d0708a6ff131#diff-df09d059ec60dd10f59f2148a61b5946f3c3c8a4a05dc9425b862ac5752816f1R127-R131)

internal/provider/resource_tfe_scim_group_mapping.go
kkavish: then either don't check group name for nil or empty in here and do it in `resolveSCIMGroupID` or do it here and remove from `resolveSCIMGroupID`. The 2 checks make code harder to maintain and read, all validations must be in one function.

website/docs/r/scim_group_mapping.html.markdown
```diff
@@ -0,0 +1,108 @@
...
+  paused        = true
+}
+```
+
+~> **Note:** Creating a mapping with `paused = true` does not skip the initial
+sync. The mapping API does not support creating a mapping in a paused state, so
+the mapping is created active and then paused in a follow-up request. This means
+an initial sync of the SCIM group's members onto the team runs before the pause
+takes effect, and any changes made to the SCIM group during that brief window may
+be reflected on the team.
```
kkavish, line 82: 👍

internal/provider/scim_helpers_test.go
skj-skj: fixed the import

internal/provider/resource_tfe_scim_group_mapping.go
skj-skj: got it, I have fixed the import.

internal/provider/resource_tfe_scim_group_mapping.go
skj-skj: makes sense, I have removed this redundant SCIM group name check

kkavish (dismissed):

internal/provider/resource_tfe_scim_group_mapping.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	// If the team is no longer SCIM-linked, the mapping was removed out-of-band.
+	if team.SCIMLinked == nil || !*team.SCIMLinked {
+		tflog.Debug(ctx, fmt.Sprintf("Team %s is no longer SCIM-linked; removing mapping from state", teamID))
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
+	scimGroupID, err := r.resolveSCIMGroupID(ctx, *team.SCIMGroupName)
```
shwetamurali: `team.SCIMGroupName` could be nil if the API returns a SCIM-linked team without the group name populated. Maybe add a nil check before dereferencing, similar to how `SCIMSyncPaused` is handled below on line 158

internal/provider/data_source_scim_group.go
```diff
@@ -105,34 +105,16 @@ func (d *dataSourceTFESCIMGroup) Read(ctx context.Context, req datasource.ReadRe
...
-		}
-		if list.Pagination == nil || list.CurrentPage >= list.TotalPages {
-			break
-		}
-		options.PageNumber = list.NextPage
+	// Reuse the shared helper so the pagination and exact, case-insensitive
+	// matching behavior stays consistent with the SCIM group mapping resource.
+	match, err := findSCIMGroupByName(ctx, d.client, name)
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to list SCIM groups", err.Error())
```
shwetamurali: `findSCIMGroupByName` already wraps errors with "unable to list SCIM groups: ...", so the diagnostic ends up with the same phrase in both the summary and detail. Consider changing the summary to something like "Error reading SCIM group" or stripping the prefix from the helper's error wrapping

internal/provider/resource_tfe_scim_group_mapping.go
skj-skj: good catch, I have [added](https://github.com/hashicorp/terraform-provider-tfe/pull/2090/changes/52d38fa3f10f3c84ca4ab9d1d51545f0f72c3659#diff-025cef9a36e31bdb77e19b06324f5daddc4a86ea76454107c748bc02f940eab3R296-R298) the nil check in `resolveSCIMGroupID` func to handle both nil and empty check in once place

internal/provider/data_source_scim_group.go
skj-skj: done, fixed the err msg

kkavish (approved):

shwetamurali (approved):

PR #2092, opened 2026-06-09
TF-38110-migrate-data-source-ip-ranges-to-plugin-framework

gbaker-ibm:
## Description

Migrate data_source_ip_ranges to tf plugin framework

_Remember to:_

## Testing plan

1.  Should have no changes to behaviour; read ip ranges using the data source. 

```
$ TF_ACC=1 go test ./internal/provider/... --run="TFEIPRanges" -v
=== RUN   TestAccTFEIPRangesDataSource_basic
--- PASS: TestAccTFEIPRangesDataSource_basic (1.69s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider   2.190s
```

## Rollback Plan

Revert PR and make a release. 

N/A.

### DIFF
--- a/internal/provider/data_source_ip_ranges.go
+++ b/internal/provider/data_source_ip_ranges.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
 var (
@@ -77,7 +78,7 @@ func (d *dataSourceTFEIPRanges) Schema(ctx context.Context, req datasource.Schem
 		Description: "This data source can be used to retrieve a list of HCP Terraform's IP ranges.",
 		Attributes: map[string]schema.Attribute{
 			"id": schema.StringAttribute{
-				Description: "Static identifier for HCP IP ranges.",
+				Description: "Static identifier for HCP Terraform's IP ranges data source.",
 				Computed:    true,
 			},
 			"api": schema.ListAttribute{
@@ -112,6 +113,7 @@ func (d *dataSourceTFEIPRanges) Read(ctx context.Context, req datasource.ReadReq
 		return
 	}
 
+	tflog.Debug(ctx, "Requesting IP Ranges")
 	ipRanges, err := d.config.Client.Meta.IPRanges.Read(ctx, "")
 	if err != nil {
 		resp.Diagnostics.AddError("Error retrieving IP Ranges", err.Error())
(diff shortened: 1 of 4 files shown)


brandonc (dismissed):
Neat

internal/provider/data_source_ip_ranges.go
```diff
@@ -1,63 +1,146 @@
...
+type modelTFEIPRanges struct {
+	ID            types.String `tfsdk:"id"`
+	API           types.List   `tfsdk:"api"`
+	Notifications types.List   `tfsdk:"notifications"`
+	Sentinel      types.List   `tfsdk:"sentinel"`
+	VCS           types.List   `tfsdk:"vcs"`
+}
+
+// modelFromTFEIPRanges builds a modelTFEIPRanges struct from a tfe.IPRanges value.
+func modelFromTFEIPRanges(i *tfe.IPRange) (modelTFEIPRanges, diag.Diagnostics) {
```
ctrombley: Let's pass our existing context in here so we can use it in the call to `types.ListValueFrom`, instead of using the global `ctx`:

```suggestion
func modelFromTFEIPRanges(ctx context.Context, i *tfe.IPRange) (modelTFEIPRanges, diag.Diagnostics) {
```

This will require corresponding changes wherever `modelFromTFEIPRanges` is used - `ctx` should already be in scope in those locations.

internal/provider/data_source_ip_ranges.go
```diff
@@ -1,63 +1,146 @@
...
+func (d *dataSourceTFEIPRanges) Configure(ctx context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	// Early exit if provider is unconfigured (i.e. we're only validating config or something)
+	if req.ProviderData == nil {
+		return
+	}
+	client, ok := req.ProviderData.(ConfiguredClient)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected Data Source Configure type",
+			fmt.Sprintf("Expected tfe.ConfiguredClient, got %T. This is a bug in the tfe provider, so please report it on GitHub.", req.ProviderData),
```
ctrombley, line 142: I'm not 100% sure about this, but I think we want:

```suggestion
			fmt.Sprintf("Expected provider.ConfiguredClient, got %T. This is a bug in the tfe provider, so please report it on GitHub.", req.ProviderData),
```

ctrombley (changes requested):
Looking really good, thanks @gbaker-ibm! I found a couple small things.

internal/provider/data_source_ip_ranges.go
gbaker-ibm: At the moment all other Configure blocks use the tfe version of message. Would you like this applied regardless?

internal/provider/data_source_ip_ranges.go
gbaker-ibm: Wholly agree, implemented

ctrombley (approved):
Looks good, thanks @gbaker-ibm !

PR #2103, opened 2026-06-17
r/tfe_project_notification_configuration: fix token consistency + add url_wo write-only support

### LINKED ISSUE #2102: tfe_project_notification_configuration: "inconsistent values for sensitive attribute" on Create when destination_type is slack/microsoft-teams/email
### Provider Version

`hashicorp/tfe` v0.78.0 (the release that introduced this resource).

### Terraform Version

Terraform 1.15.6 (TFC remote execution).

### Affected Resource(s)

`tfe_project_notification_configuration`

### Terraform Configuration Files

Minimal config that reproduces:

```hcl
terraform {
  required_providers {
    tfe = {
      source  = "hashicorp/tfe"
      version = "~> 0.78"
    }
  }
}

resource "tfe_project" "example" {
  organization = "my-org"
  name         = "example"
}

resource "tfe_project_notification_configuration" "example" {
  project_id       = tfe_project.example.id
  name             = "example"
  destination_type = "slack"
  enabled          = true
  url              = "https://hooks.slack.com/services/AAA/BBB/CCC"

  triggers = [
    "run:needs_attention",
    "run:errored",
  ]
}
```

Same failure mode for `destination_type = "microsoft-teams"` or `"email"` — any destination type whose schema validator forbids setting the `token` attribute.

### Expected Behavior

`terraform apply` creates the project-level notification configuration successfully and commits it to state.

### Actual Behavior

The underlying API resource is created successfully, but the apply errors with:

```
Error: Provider produced inconsistent result after apply

When applying changes to tfe_project_notification_configuration.example,
provider "provider[\"registry.terraform.io/hashicorp/tfe\"]" produced an
unexpected new value: .token: inconsistent values for sensitive attribute.

This is a bug in the provider, which should be reported in the provider's
own issue tracker.
```

Terraform's post-apply consistency check fails. The resource lands in state as **tainted**, and subsequent `apply` runs plan to destroy and recreate it — hitting the same error in a loop. The API-side resource accumulates duplicates if the user keeps retrying (or stays orphaned if state was rolled back).

### Steps to Reproduce

1. Set up a fresh project and apply the minimal config above.

### Root cause

In [`internal/provider/resource_tfe_project_notification_configuration.go`](https://github.com/hashicorp/terraform-provider-tfe/blob/v0.78.0/internal/provider/resource_tfe_project_notification_configuration.go), `modelFromTFEProjectNotificationConfiguration` initializes `Token` to an empty string:

```go
result := modelTFEProjectNotificationConfiguration{
    ID:              types.StringValue(v.ID),
    Name:            types.StringValue(v.Name),
    DestinationType: types.StringValue(string(v.DestinationType)),
    Enabled:         types.BoolValue(v.Enabled),
    ProjectID:       types.StringValue(v.SubscribableChoice.Project.ID),
    TokenWOVersion:  tokenWOVersion,
    Token:           types.StringValue(""),  // <-- always initialized to "" rather than null
}

// ...

if lastValue.ValueString() != "" {
    result.Token = lastValue
}

if !tokenWOVersion.IsNull() {
    result.Token = types.StringNull()
}
```

When `destination_type` is anything other than `generic`, the schema validator forbids the user from setting `token` in config:

```go
"token": schema.StringAttribute{
    Optional:    true,
    Sensitive:   true,
    Validators: []validator.String{
        validators.AttributeValueConflictValidator(
            "destination_type",
            []string{"email", "microsoft-teams", "slack"},
        ),
        // ...
    },
},
```

So at plan time, `token` is null. After Create, `lastValue` is also null (since the user couldn't set `token`), and `tokenWOVersion` is null (no write-only variant). Neither of the conditional branches fires, leaving `result.Token = types.StringValue("")`. The resulting state has `token = ""`, which mismatches the planned `null`, and Terraform core rejects the apply because `token` is marked `Sensitive`.

### Comparison with the working sibling resource

The workspace-scoped sibling `tfe_notification_configuration` does this correctly in [`internal/provider/resource_tfe_notification_configuration.go`](https://github.com/hashicorp/terraform-provider-tfe/blob/v0.78.0/internal/provider/resource_tfe_notification_configuration.go):

```go
result := modelTFENotificationConfiguration{
    ID:              types.StringValue(v.ID),
    // ... no explicit Token initialization; defaults to types.StringNull()
}

if v.Token != "" {
    result.Token = types.StringValue(v.Token)
}
isWriteOnlyValue := !tokenWOVersion.IsNull()
if isWriteOnlyValue {
    result.Token = types.StringNull()
}
```

When the API returns `v.Token == ""`, `result.Token` stays as the zero value `types.StringNull()`. No inconsistency. This is the pattern the project-scoped resource should follow.

### Suggested Fix

In `modelFromTFEProjectNotificationConfiguration`:

```diff
 result := modelTFEProjectNotificationConfiguration{
     ID:              types.StringValue(v.ID),
     Name:            types.StringValue(v.Name),
     DestinationType: types.StringValue(string(v.DestinationType)),
     Enabled:         types.BoolValue(v.Enabled),
     ProjectID:       types.StringValue(v.SubscribableChoice.Project.ID),
     TokenWOVersion:  tokenWOVersion,
-    Token:           types.StringValue(""),
 }
 
-if lastValue.ValueString() != "" {
-    result.Token = lastValue
+if v.Token != "" {
+    result.Token = types.StringValue(v.Token)
+} else if lastValue.ValueString() != "" {
+    result.Token = lastValue
 }
 
 if !tokenWOVersion.IsNull() {
     result.Token = types.StringNull()
 }
```

Or, more minimally, simply replace the initializer:

```diff
-Token: types.StringValue(""),
+Token: types.StringNull(),
```

— the existing conditional `if lastValue.ValueString() != ""` then correctly sets `Token` to the planned value when the user provided one (i.e., for `destination_type = "generic"`).

### References

- The same code path runs on `Read` and `Update`, which is why `terraform untaint` plus `lifecycle { ignore_changes = [token] }` is currently required to keep affected configs stable across plans.

### LINKED ISSUE #2104: tfe_project_notification_configuration: add url_wo / url_wo_version write-only support (parity with tfe_notification_configuration)
### Feature request

Add `url_wo` and `url_wo_version` write-only attributes to `tfe_project_notification_configuration`, mirroring the support that the workspace-scoped sibling `tfe_notification_configuration` gained in v0.76.2.

### Motivation

For `destination_type = "slack"` (and `generic`), the `url` attribute is **effectively a secret**. Slack and generic webhook URLs carry the bearer token directly in the path, e.g.

```
https://hooks.slack.com/services/T<workspace>/B<channel>/<token>
```

Anyone who reads the URL can post to the channel / endpoint. With the current schema:

```go
"url": schema.StringAttribute{
    Optional: true,
    // not Sensitive, no write-only variant
    ...
}
```

the URL ends up in:

- the terraform state file,
- every TFC state-version snapshot,
- `terraform show` and the JSON plan output,
- any CI log line that prints plan summaries.

The workspace-scoped sibling `r/tfe_notification_configuration` solved this in v0.76.2 by adding `url_wo` (write-only) plus `url_wo_version` (change-trigger). The project-scoped resource, added in v0.78.0, didn't pick up the pattern.

### Existing precedent in the same provider

`r/tfe_notification_configuration` schema after v0.76.2:

```go
"url": schema.StringAttribute{
    Optional: true,
    Validators: []validator.String{
        stringvalidator.ConflictsWith(path.MatchRoot("url_wo")),
        stringvalidator.PreferWriteOnlyAttribute(path.MatchRoot("url_wo")),
        // ... existing required-if-destination-type validators
    },
},
"url_wo": schema.StringAttribute{
    Optional:  true,
    WriteOnly: true,
    Sensitive: true,
    Validators: []validator.String{
        stringvalidator.ConflictsWith(path.MatchRoot("url")),
        stringvalidator.AlsoRequires(path.MatchRoot("url_wo_version")),
    },
},
"url_wo_version": schema.Int64Attribute{
    Optional: true,
    Validators: []validator.Int64{
        int64validator.ConflictsWith(path.MatchRoot("url")),
        int64validator.AlsoRequires(path.MatchRoot("url_wo_version")),
    },
},
```

Read path:

```go
isURLWriteOnly := !urlWOVersion.IsNull()
if isURLWriteOnly {
    result.URL = types.StringNull()  // never persist when using write-only
} else if v.URL != "" {
    result.URL = types.StringValue(v.URL)
}
```

Update path uses a small `determineURLForUpdate` helper symmetric to `determineTokenForUpdate`.

### Proposed implementation for the project-scoped resource

The change is the same shape as the existing `token` / `token_wo` / `token_wo_version` triplet already present in `r/tfe_project_notification_configuration`:

1. Add three new attributes to the schema (`url_wo`, `url_wo_version`) following the validator conventions above.
2. Add `URLWO types.String` and `URLWOVersion types.Int64` to `modelTFEProjectNotificationConfiguration`.
3. In `modelFromTFEProjectNotificationConfiguration`, accept `urlWOVersion types.Int64` and force `result.URL = types.StringNull()` when it's set.
4. In `Create`, plumb the write-only URL into `options.URL` from `config.URLWO` when `config.URLWO.IsNull()` is false (otherwise keep using `plan.URL`).
5. In `Update`, add a `determineURLForUpdate` helper following the existing `determineTokenForUpdate` pattern (handles switch-from-normal-to-write-only, switch-back, version-bumps, and unrelated-attribute updates that should preserve the existing URL).
6. Update the resource documentation page accordingly.

The user-facing config swap then looks like:

```hcl
resource "tfe_project_notification_configuration" "example" {
  destination_type = "slack"
  url              = "https://hooks.slack.com/services/AAA/BBB/CCC"   # ends up in state
  # ...
}
```

becomes

```hcl
resource "tfe_project_notification_configuration" "example" {
  destination_type = "slack"
  url_wo           = "https://hooks.slack.com/services/AAA/BBB/CCC"   # never in state
  url_wo_version   = 1                                                  # bump to rotate
  # ...
}
```

### Related

- Workspace-scoped sibling's `url_wo` PR / release: v0.76.2 changelog entry "Enhanced with `url_wo` write-only option".
- This feature gap was surfaced during the investigation of #2102 (the related `token` initializer bug that was blocking project-scoped notifications for non-generic destination types).

I'd be happy to send a PR for this if it's something the maintainers would accept.

c4po:
Closes #2102. Closes #2104.

This PR bundles two related improvements to `tfe_project_notification_configuration`. They share a code path (`modelFromTFEProjectNotificationConfiguration`) and the same test file, and are scoped tightly to this single resource.

---

## Part 1 — Fix token consistency for non-generic destinations (#2102)

For destination types whose schema validator forbids setting `token` (`slack`, `microsoft-teams`, `email`), `Create` returned state with `token = ""` while the plan had `token = null`. Terraform core's post-apply consistency check on the sensitive attribute then failed:

```
Error: Provider produced inconsistent result after apply

When applying changes to tfe_project_notification_configuration.<x>,
provider produced an unexpected new value: .token: inconsistent values
for sensitive attribute.
```

The underlying API resource was created successfully, but the state landed as tainted and every subsequent apply destroyed and re-created the resource, hitting the same error in a loop.

### Root cause

`modelFromTFEProjectNotificationConfiguration` initialized `Token: types.StringValue("")`. For validator-blocked destination types, neither the user-provided-token branch nor the write-only branch fires, so Token stays at `""`. The equivalent workspace-scoped sibling `r/tfe_notification_configuration` already does the right thing — initializes Token at the zero value (null) and only assigns when the API returns a non-empty token.

### Fix

Drop the explicit `Token: types.StringValue("")` initializer so it defaults to `types.StringNull()`. The existing conditional logic still correctly assigns the user-supplied token (or write-only null) for the `generic` destination type.

---

## Part 2 — Add `url_wo` / `url_wo_version` write-only support (#2104)

For `destination_type = "slack"` (or `generic`), the `url` attribute is **effectively a secret** — Slack and generic webhook URLs embed the bearer token directly in the path:

```
https://hooks.slack.com/services/T<workspace>/B<channel>/<token>
```

Without a write-only variant, the URL ends up in:

- the terraform state file,
- every TFC state-version snapshot,
- `terraform show` and JSON plan output,
- any CI log line that prints plan summaries.

The workspace-scoped sibling `r/tfe_notification_configuration` gained `url_wo` + `url_wo_version` in v0.76.2; the project-scoped resource (added in v0.78.0) didn't pick the pattern up.

### Implementation

Mirrors the workspace-scoped sibling exactly:

- New schema attributes:
  - `url_wo` — `Optional`, `WriteOnly`, `Sensitive`. Never persisted to state.
  - `url_wo_version` — `Optional`, `Computed` `Int64`. Auto-incremented when the hash of `url_wo` changes between applies; can also be set manually to force a URL update without changing the value.
- `url` schema gets `Sensitive: true` and a `ConflictsWith url_wo` validator. The `AttributeRequiredIfValueString` validator is upgraded to `AttributeRequiredIfValueStringUnlessOtherSet` so either `url` or `url_wo` satisfies the requirement for non-`email` destinations.
- `modelTFEProjectNotificationConfiguration` gains `URLWO` and `URLWOVersion` fields.
- `modelFromTFEProjectNotificationConfiguration` now accepts `urlWOVersion types.Int64` and forces state's `url` to null when write-only is active.
- `Create` and `Update` plumb the URL through `config.URLWO` when set, and call `storeWOHash` to persist a SHA-256 hash in private state for change detection.
- New `ModifyPlan` hook (`resource.ResourceWithModifyPlan`):
  - Blocks write-only → plaintext transitions (prevents accidental exposure of a previously-secret URL).
  - Auto-increments `url_wo_version` when the SHA-256 hash of `url_wo` changes between applies (or sets it to 1 on first create).
  - Honors manual mode: skips auto-detection when `url_wo_version` is explicitly set in config.
- New `determineURLForUpdate` helper, symmetric to the existing `determineTokenForUpdate`, prevents accidental clearing of the URL when an unrelated attribute change triggers an Update.

All shared helpers (`computeWOHash`, `storeWOHash`, `blockWOToPlaintextTransition`) already exist as package-level functions in the workspace-scoped sibling file and are reused as-is. `modifyPlanWOVersion` is a method on the resource type in the sibling; it's duplicated as a method on this resource so the two can evolve independently if the project-scoped API diverges.

---

## Tests

Two new acceptance tests in `internal/provider/resource_tfe_project_notification_configuration_test.go`:

### `TestAccTFEProjectNotificationConfiguration_slack` (regression test for #2102)

Creates a slack-destination notification with no token in config and verifies:

1. Create succeeds without the consistency error.
2. The `token` attribute is absent from state (null, not `""`).
3. A second plan with the same config produces no drift (`PlanOnly: true` step).

With the bug present, step 1 fails with the consistency-check error; with the fix, both steps pass.

### `TestAccTFEProjectNotificationConfiguration_urlWO` (coverage for #2104)

Exercises the new write-only URL pair:

1. Create with `url_wo` — state's `url` is absent, `url_wo_version` is 1.
2. PlanOnly re-apply — no drift detected.
3. Change `url_wo` — `url_wo_version` auto-increments to 2 (hash-change detection).
4. Attempt to switch back to plaintext `url` — blocked with the new write-only-to-plaintext guard error.

Both tests run under the existing `skipUnlessBeta` gate, matching the rest of the file.

---

## Verification

```
$ go build ./...
$ go vet ./internal/provider/...
$ gofmt -l internal/provider/resource_tfe_project_notification_configuration.go internal/provider/resource_tfe_project_notification_configuration_test.go website/docs/r/project_notification_configuration.html.markdown
(no output)
```

---

## Diff stat

```
 CHANGELOG.md                                                          |   4 +
 internal/provider/resource_tfe_project_notification_configuration.go  | 213 ++++++++++++++++++++--
 internal/provider/resource_tfe_project_notification_configuration_test.go |  168 ++++++++++++++++
 website/docs/r/project_notification_configuration.html.markdown       |   8 +-
 4 files changed, 381 insertions(+), 12 deletions(-)
```

Existing tests (`TestAccTFEProjectNotificationConfiguration_basic`, `..._update`, `..._validateSchemaAttributesSlack`, `..._emailUserIDs`) are unchanged and continue to pass — they exercise `destination_type = "generic"` (which sets `token` in config and so didn't surface the consistency bug) and schema-validator failure modes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ FEATURES:
 
 BUG FIXES:
 * Removed two documentation files for resources that do not exist and that are not planned. By @brandonc [#2100](https://github.com/hashicorp/terraform-provider-tfe/pull/2100)
+* `r/tfe_project_notification_configuration`: Fix "Provider produced inconsistent result after apply" error on Create for `destination_type` values that forbid setting `token` (`slack`, `microsoft-teams`, `email`). The plan-time `token` is null but the resource was initializing the state value to `""`, tripping Terraform core's sensitive-attribute consistency check. The initializer now defaults to null, matching the equivalent workspace-scoped resource. [#2102](https://github.com/hashicorp/terraform-provider-tfe/issues/2102)
 
 
 ## v0.78.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ BUG FIXES:
 * Removed two documentation files for resources that do not exist and that are not planned. By @brandonc [#2100](https://github.com/hashicorp/terraform-provider-tfe/pull/2100)
 * `r/tfe_project_notification_configuration`: Fix "Provider produced inconsistent result after apply" error on Create for `destination_type` values that forbid setting `token` (`slack`, `microsoft-teams`, `email`). The plan-time `token` is null but the resource was initializing the state value to `""`, tripping Terraform core's sensitive-attribute consistency check. The initializer now defaults to null, matching the equivalent workspace-scoped resource. [#2102](https://github.com/hashicorp/terraform-provider-tfe/issues/2102)
 
+ENHANCEMENTS:
+* `r/tfe_project_notification_configuration`: Adds `url_wo` (write-only) and `url_wo_version` attributes, plus auto change-detection via SHA-256 hashing in private state and a write-only → plaintext transition guard. Mirrors the existing `token_wo` mechanism and brings the resource to parity with `r/tfe_notification_configuration`, letting users keep Slack / Generic webhook URLs out of state. [#2104](https://github.com/hashicorp/terraform-provider-tfe/issues/2104)
+
 
 ## v0.78.0
 
--- a/internal/provider/resource_tfe_project_notification_configuration.go
+++ b/internal/provider/resource_tfe_project_notification_configuration.go
@@ -4,7 +4,9 @@
 package provider
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 
@@ -18,6 +20,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
@@ -30,6 +33,7 @@ import (
 var _ resource.Resource = &resourceTFEProjectNotificationConfiguration{}
 var _ resource.ResourceWithConfigure = &resourceTFEProjectNotificationConfiguration{}
 var _ resource.ResourceWithImportState = &resourceTFEProjectNotificationConfiguration{}
+var _ resource.ResourceWithModifyPlan = &resourceTFEProjectNotificationConfiguration{}
 
 func NewProjectNotificationConfigurationResource() resource.Resource {
 	return &resourceTFEProjectNotificationConfiguration{}
@@ -56,12 +60,14 @@ type modelTFEProjectNotificationConfiguration struct {
 	TokenWOVersion  types.Int64  `tfsdk:"token_wo_version"`
 	Triggers        types.Set    `tfsdk:"triggers"`
 	URL             types.String `tfsdk:"url"`
+	URLWO           types.String `tfsdk:"url_wo"`
+	URLWOVersion    types.Int64  `tfsdk:"url_wo_version"`
 	ProjectID       types.String `tfsdk:"project_id"`
 }
 
 // modelFromTFEProjectNotificationConfiguration builds a modelTFEProjectNotificationConfiguration
 // struct from a tfe.NotificationConfiguration value.
-func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfiguration, tokenWOVersion types.Int64, lastValue types.String) (*modelTFEProjectNotificationConfiguration, diag.Diagnostics) {
+func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfiguration, tokenWOVersion, urlWOVersion types.Int64, lastValue types.String) (*modelTFEProjectNotificationConfiguration, diag.Diagnostics) {
 	var diags diag.Diagnostics
 	result := modelTFEProjectNotificationConfiguration{
 		ID:              types.StringValue(v.ID),
@@ -70,6 +76,7 @@ func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfigurati
 		Enabled:         types.BoolValue(v.Enabled),
 		ProjectID:       types.StringValue(v.SubscribableChoice.Project.ID),
 		TokenWOVersion:  tokenWOVersion,
+		URLWOVersion:    urlWOVersion,
 		// Token defaults to types.StringNull(); only populated below when
 		// the user actually provided one (lastValue). Initializing to "" here
 		// causes a post-apply "inconsistent values for sensitive attribute"
@@ -116,7 +123,11 @@ func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfigurati
 		result.Token = types.StringNull()
 	}
 
-	if v.URL != "" {
+	// URL is null in state when using url_wo so the secret never persists.
+	isURLWriteOnly := !urlWOVersion.IsNull()
+	if isURLWriteOnly {
+		result.URL = types.StringNull()
+	} else if v.URL != "" {
 		result.URL = types.StringValue(v.URL)
 	}
 
@@ -268,12 +279,14 @@ func (r *resourceTFEProjectNotificationConfiguration) Schema(ctx context.Context
 			},
 
 			"url": schema.StringAttribute{
-				Description: "The HTTP or HTTPS URL where notification requests will be made. This value must not be provided if `email_addresses` or `email_user_ids` is present, or if `destination_type` is `email`.",
+				Description: "The HTTP or HTTPS URL where notification requests will be made. This value must not be provided if `email_addresses` or `email_user_ids` is present, or if `destination_type` is `email`. Use `url_wo` instead to prevent the URL from being stored in state.",
 				Optional:    true,
+				Sensitive:   true,
 				Validators: []validator.String{
-					validators.AttributeRequiredIfValueString(
+					validators.AttributeRequiredIfValueStringUnlessOtherSet(
 						"destination_type",
 						[]string{"generic", "microsoft-teams", "slack"},
+						"url_wo",
 					),
 					validators.AttributeValueConflictValidator(
 						"destination_type",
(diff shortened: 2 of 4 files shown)


markdecrane:
Hey @c4po! The underlying issues were closed out by #2150 and #2140, apologies for not catching this PR when we were looking to address those issues 😅. Otherwise, these changes are slated to go out this Thursday in the `v0.80.0` release.

PR #2105, opened 2026-06-18
TF-38378-reformatting-of-tfplugindocs-template-outputs-to-match-prior-doc-format

gbaker-ibm:
## Description

Setting up some test examples and templates for use in a run of tfplugindocs. No behaviours should be modified. 

## External links

- [tfplugindocs reference](https://github.com/hashicorp/terraform-plugin-docs)

## Example Outputs

As a note, this format is not the exact same as the prior docs, but does convey the same information in a similar format. Making a closer result likely requires changes to or distance from tfplugindocs. 

Screenshot on the first half of tfe_notification_configuration in the output

Screenshot on the second half of tfe_notification_configuration in the output

## Rollback Plan

Rollback of the PR is sufficient. 

N/A

### DIFF
--- a/examples/actions/tfe_query_run/direct.sh
+++ b/examples/actions/tfe_query_run/direct.sh
@@ -1,1 +1,0 @@
-terraform apply -invoke=action.tfe_query_run.wait_for_latest
--- a/examples/actions/tfe_query_run/direct.sh
+++ b/examples/actions/tfe_query_run/direct.sh
@@ -1,0 +1,1 @@
+terraform apply -invoke=action.tfe_query_run.wait_for_latest
--- a/examples/resources/tfe_team/resource.tf
+++ b/examples/resources/tfe_team/resource.tf
@@ -1,0 +1,8 @@
+import {
+  to = tfe_team.test
+  identity = {
+    id           = "team-uomQZysH9ou42ZYY"
+    organization = "my-org-name"
+    hostname     = "app.terraform.io"
+  }
+}
--- a/templates/resources.md.tmpl
+++ b/templates/resources.md.tmpl
@@ -65,6 +65,8 @@ Additional example(s):
 
 {{- if .HasImport }}
 
+{{.Name}} can be imported in the following format: 
+
 {{ codefile "shell" .ImportFile }}
 {{- end }}
 {{- end }}
--- a/templates/resources.md.tmpl
+++ b/templates/resources.md.tmpl
@@ -1,0 +1,70 @@
+---
+layout: "tfe"
+page_title: "Terraform Enterprise: {{.Name}}"
+description: |-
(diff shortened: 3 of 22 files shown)


templates/resources.md.tmpl
ctrombley, line 1: I see that this template file is the only one that is different from the others, and has sections for import blocks and some other things used by resources. 

Since these extras are also gated with conditionals (e.g. `{{- if .HasImportIdentityConfig }}`) I wonder if we can get away with a single template across all components, regardless of type. `provider.md.tmpl`? or `component.md.tmpl`? ("Component" isn't a common term in this domain, but it's useful for the purpose of this discussion and also totally fair ground for naming internal components.)

Does tfplugindocs allow for this or does it require a top-level template for each component type?

examples/resources/tfe_team/resource.tf
```diff
@@ -0,0 +1,8 @@
+import {
+  to = tfe_team.test
+  identity = {
+    id           = "team-uomQZysH9ou42ZYY"
+    organization = "my-org-name"
+    hostname     = "app.terraform.io"
+  }
+}
```
ctrombley: This doesn't seem quite right - I would expect no import block here and at least one invocation of `tfe_team`.

examples/actions/tfe_query_run/direct.sh
ctrombley, line 1: While this is perfectly valid, I think it might cause us a minor headache when we start bulk validating everything in the `examples` folder since it's the only new artifacts that doesn't have hcl. 

My inclination is to pull it for now and focus on the HCL examples. We can address it later separately.

ctrombley (commented):
This is looking great, thanks @gbaker-ibm!

Can we run `terraform fmt` against each of the examples here before merging this? 

(Running `terraform fmt` in CI and checking for the existence of changes might be another good validation check for the validation automation story - CI can fail early if the formatting has not been applied before checking in the examples.)

templates/resources.md.tmpl
gbaker-ibm: This in particular does not appear possible with current tfplugindocs, it only applies fallback templates at the component-type level.

examples/actions/tfe_query_run/direct.sh
gbaker-ibm: I can remove it. It is worth noting that import examples in both .sh and .tf form must reside in the same examples/ directory as usage .tf examples in the current form of tfplugindocs.

examples/resources/tfe_team/resource.tf
gbaker-ibm: Agree, fixed + updates from `terraform fmt` also added

ctrombley (approved):
Looks good, thanks @gbaker-ibm !

PR #2110, opened 2026-06-23
Migrate resource_tfe_ssh_key to go-tfe v2

brandonc:
## Description

- Introduces ClientV2 within Meta ConfiguredClient
- Utilizes new client for `tfe_ssh_key` resource CRUD operations

_Remember to:_


internal/provider/resource_tfe_ssh_key.go
```diff
@@ -44,10 +45,24 @@ type modelTFESSHKey struct {
...
+	attributes := &models.SshKeys_attributes{}
+	attributes.SetName(name)
+	attributes.SetValue(value)
+
+	optionsSSHKey := &models.SshKeys{}
+	options := &models.SshKeysEnvelope{}
+	options.SetData(optionsSSHKey)
+	optionsSSHKey.SetAttributes(attributes)
+	return options
+}
```
markdecrane, line 58: Non-blocking, and otherwise more of a go-tfe v2 related thought.

It saddens me slightly that we're losing the options struct(s) that marshaled to jsonapi, if nothing else since we'll have to push the responsibility to anything consuming v2 creating a potential duplication of efforts. Any idea if there's a path where we could code gen helper fixtures in a _somewhat_ similar way as a possible future improvement?

Maed223 (dismissed):

ctrombley:
Thanks, this looks good! I think I spotted a potential regression:

`internal/client/GetClient` currently applies provider transport/TLS settings only to the v1 client (`HTTPClient: config.HTTPClient`). The new v2 client is created without reusing that configured transport.

When using `tfe_ssh_key` in environments that rely on `ssl_skip_verify` (self-signed/private CA) this might fail because SSH key CRUD now goes through v2 while most other resources still use v1.

I also don’t see a test that would catch this today:
- `internal/client/client_test.go::Test_GetClient` only exercises `providerClient.TfeClient` (v1) via `Organizations.List`.
- It does not make any request through `providerClient.TFEClientV2`.

Can we add a test that asserts v2 requests succeed against the TLS test server when `GetClient(..., insecure=true)` is used (and fail when false)?

internal/provider/resource_tfe_ssh_key.go
brandonc: This method eliminates all the mapping ambiguity but does sacrifice usability. I've been trying to think of an alternative, like code-generating helpers like this or implementing a mapper from an annotated struct to a `models` package struct. One major obstacle is that the `models` types use Setter functions, not fields.

brandonc:
@ctrombley Fixed this by introducing a new HTTPTransport config option for v2 clients, and using the config client transport that was constructed when configuring the provider

ctrombley (approved):
Thank you @brandonc!

PR #2115, opened 2026-06-24
Add targetable docgen to terraform provider tfe

marzuqmazid-ibm:
## Description

This PR adds a targetable doc generation workflow for terraform-provider-tfe using tfplugindocs.

Previously, docs required manually installing tfplugindocs outside the repo. This change integrates it into the Go module system and exposes it via `make generate`, making doc generation reproducible and consistent with other dev workflows.

Key points:
- tfplugindocs is tracked as a Go tool dependency (`tools/tools.go`)
- `make generate` builds full docs into `docs/`
- `make generate RESOURCE=<name>` targets a single resource/data source
- Existing handwritten docs in `website/docs/` remain unchanged

Changes:
- `tools/tools.go` — add tfplugindocs tool dependency  
- `go.mod` / `go.sum` — pin tfplugindocs version  
- `main.go` — add `//go:generate` directive  
- `GNUmakefile` — replace old docs target with `generate`

- [x] Update the Change Log
- [x] Update the Documentation

## Testing plan

1. Clone the repo and confirm it builds:
2. Run full generation
    - Expect all docs generated under `docs/`, existing `website/docs/` untouched.
    
## External links

- tfplugindocs

Not applicable — this change only introduces tooling and does not affect provider behavior.

## Rollback Plan

Revert the makefile change, remove `tools/tools.go`, and run `go mod tidy`. No impact to provider functionality.

None. This only affects doc generation tooling and does not change access controls, encryption, or logging.


main.go
```diff
@@ -3,6 +3,8 @@
 
 package main
 
+//go:generate go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs generate --provider-name tfe
```
ctrombley: I would prefer we not wire doc generation through `main.go` via //go:generate; can we invoke `tfplugindocs` directly from the make target instead, so docs tooling is explicit and decoupled from provider runtime code?

GNUmakefile
```diff
@@ -63,5 +63,10 @@ test-compile:
 	fi
 	go test -c $(TEST) $(TESTARGS)
 
-.PHONY: build test testacc vet fmt fmtcheck errcheck test-compile sweep
-
+generate:
+	go generate ./...
```
ctrombley: Can we make this the entry point for tfplugindocs instead of main.go?  

```suggestion
    go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs@v0.25.0 generate --provider-name tfe
```

GNUmakefile
```diff
@@ -63,5 +63,10 @@ test-compile:
...
+ifdef RESOURCE
+	find docs -type f -name '*.md' ! -name '$(RESOURCE).md' -delete
+	find docs -type d -empty -delete
```
ctrombley: I'd like to request some change to the way the glob matching works. Right now all docs are being deleted that do not match the pattern. We don't want to be that destructive. We can use a strategy of generating docs into a temporary folder, and then copying anything that matches the passed glob into the repository, overwriting just the files that match the glob.

The following suggestion has this structure:
1. create a temp folder (and cleanup with `trap` before exiting)
2. generate **all** docs using `tfplugindocs` to the temp folder
3. set `matches` to the generated file names that match the glob passed in RESOURCE
4. check existence of `matches`, exit if none
5. if there are matches, for each one, copy the generated docs from the temp folder into the expected location in the repository. The use of `rel` is present so that we properly traverse each of the nested directories (`docs/resources/`, `docs/ephemeral-resources/`, `docs/actions/`, `docs/data-sources/`) when copying docs over from the temp dir.

```suggestion
    @tmp="$$(mktemp -d)"; trap 'rm -rf "$$tmp"' EXIT; \
    go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs@v0.25.0 generate --provider-name tfe --rendered-website-dir "$$tmp"; \
    matches="$$(find "$$tmp" -type f -path "$$tmp/$(RESOURCE)")"; \
    test -n "$$matches" || { echo "No generated docs matched RESOURCE=$(RESOURCE)"; exit 1; }; \
    for f in $$matches; do \
       rel="$${f#$$tmp/}"; \
       mkdir -p "docs/$$(dirname "$$rel")"; \
       cp "$$f" "docs/$$rel"; \
       echo "updated docs/$$rel"; \
    done
```

ctrombley (changes requested):
Hey @marzuqmazid-ibm , this is pretty close! I have requested a few changes in comments.

I would also like to ask if we could avoid checking in all of the generated docs files.

You should be able to follow these steps:
1. `git checkout origin/main -- docs`: this should reset everything in the `docs/` folder to the main origin branch.
1. Address the PR comments I have left
1. Once the PR comments are addressed, run generation for a single file of your choosing (example: `make generate RESOURCE='resources/organization.md'`)
1. check in and push your changes to this branch

Right now this PR touches 138 files, but we are looking for a PR with only 3 files changed:
```
 M GNUmakefile
 M docs/resources/organization.md  (or whichever doc file you chose to generate)
 M main.go
```

Please let me know if you have questions!

GNUmakefile
marzuqmazid-ibm: Updated!

GNUmakefile
marzuqmazid-ibm: Updated!

main.go
marzuqmazid-ibm: Yes that makes sense! Would you suggest that I remove this line completely?

tools/tools.go
ctrombley, line 1: I belive this can go away too now that tfplugindocs is referenced from GNUMakefile.

ctrombley (approved):
Nicely done!

PR #2117, opened 2026-06-25
Compile the TFE provider for linux s390x

JenGoldstrich:
## Description

Adds support running the Terraform TFE provider from an s390x runner.

Nearly identical to https://github.com/hashicorp/terraform-provider-aws/pull/48272

_Remember to:_

## Testing plan

It's just a new compiled platform!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -125,7 +125,7 @@ jobs:
       # Verify expected Artifacts list for a workflow run.
       matrix:
         goos: [freebsd, windows, linux, darwin]
-        goarch: ["386", "amd64", "arm", "arm64"]
+        goarch: ["386", "amd64", "arm", "arm64", "s390x"]
         exclude:
           - goos: freebsd
             goarch: arm64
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -137,6 +137,15 @@ jobs:
             goarch: 386
           - goos: darwin
             goarch: arm
+          # Only compile on s390x for Linux
+          - goos: openbsd
```
christian-doucette: openbsd isn't needed in the exclusions since we aren't building to that in the first place

.github/workflows/build.yml
```diff
@@ -137,6 +137,15 @@ jobs:
...
+            goarch: s390x
+          - goos: darwin
+            goarch: s390x
+          - goos: freebsd
+            goarch: s390x
+          - goos: windows
```
christian-doucette: also very minor, but the rest of the exclusions list is sorted by goos not goarch 

so it would be nice to keep that for readability

ex:
```
        exclude:
          - goos: freebsd
            goarch: arm64
          - goos: freebsd
            goarch: s390x
          - goos: windows
            goarch: arm64
          - goos: windows
            goarch: arm
          - goos: windows
            goarch: s390x
          - goos: darwin
            goarch: 386
          - goos: darwin
            goarch: arm
          - goos: darwin
            goarch: s390x
```

christian-doucette (commented):
couple small comments

christian-doucette (approved):

PR #2125, opened 2026-07-02
Tf 38917 backfill of descriptions into component schemas for data sources ephemerals and actions

gbaker-ibm:
## Description

Both sdk/v2 and plugin-framework versions of data-sources, ephemeral-resources, and actions have the capacity to hold description information which might be used in documentation generation or in other uses. Although copies of the information usually exist in `website/docs`, they have often not been backfilled into the source files themselves. This PR performs that backfill. 

We should not consider the backfill process complete for three categories of work: 

1. Items with nested attributes - these do have the information backfilled, but require migration to provider-framework for tools like [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) to capture them. 
2. Items with mandatory, unused "id" fields - these are discussed later in this PR. Performing the backfill here could be misleading. 
3. Items with mapped attributes - these require migration to plugin-framework in order to provide the opportunity to associate sub-attribute descriptions with the sub-attribute. Instead, this information has been backfilled into their parent attribute, but this solution ought to be considered temporary. 

## Testing plan

Regarding the beneficial output

1. Generate a copy of the provider schema. One method can be seen [here](https://gist.github.com/gbaker-ibm/f74fc2a9d48491893b4abfc71a7ae3d3)
5. Identify that all non-resources now have descriptions associated with their components and the attributes under those components. For resources which have nested subcomponents, identify instead that the information is simply available in the source for later use. 
A functional tool for doing so exists [here](https://gist.github.com/gbaker-ibm/9b3532894df61ab1cd72027f4ba0553f). 
For the failing items, note that their "id" is the only attribute missing a description, that they are using sdk/v2 instead of plugin-framework, and that their id is not used in the go source code. We note that sdk/v2 will generate an "id" field even if none is intended to exist. (`oauth_client, organization, organization_tags, organizations, policy_set, slug, workspace_ids`). 

Regarding non-negative changes

1.  Run existing acceptance tests against DataSource(s)
2. Run existing acceptance test against the action `query_run`

### For the data sources: 

Note that some tests failed. We submit that most of these do not result from changes made by this branch, considering that the testing states match. 

Attached as a file is the test output for the main branch without this repo with the same arguments: 
[test_results_base.txt](https://github.com/user-attachments/files/29707654/test_results_base.txt)

Below is the run for this repo. 

```
TFE_ADMIN_PROVISION_LICENSES_TOKEN=$TFE_TOKEN TF_ACC=1 go test ./internal/provider/... --run="DataSource" -v

[...]

PR #2128, opened 2026-07-07
Backfill descriptions into resource component schemas

marzuqmazid-ibm:
## Description

Both the SDK/v2 and Plugin Framework versions of resources have the capacity to hold description information at both the component (resource) level and on their individual schema attributes, which may be used in documentation generation or other tooling. Although copies of this information usually exist in `website/docs`, it has often not been backfilled into the source files themselves. This PR performs that backfill for resource component schemas.

Missing descriptions were added at both the component level and the attribute level across resource files, sourced from the corresponding `website/docs/r/*.html.markdown` files. Deprecated attributes retain their existing `Deprecated`/`DeprecationMessage` fields unchanged, with a short deprecation note appended to their new descriptions.

As with the sibling data-source backfill, the backfill should not be considered complete for the following categories of work:
- **Items with nested attributes** — the information is backfilled into the source, but these require migration to the Plugin Framework for tools like `tfplugindocs` to capture them.
- **Mandatory, unused `id` fields** — SDK/v2 auto-generates an `id` field even where none is intended. Describing it could be misleading, so these were intentionally left undescribed.
- **Items with mapped attributes** — associating descriptions with sub-attributes requires migration to the Plugin Framework. In the interim, the information has been backfilled into the parent attribute; this should be considered a temporary solution.

Additionally, state-migration schemas (V0 upgrade schemas and the `*_migrate.go` files) and identity schemas (SDK/v2 `Identity:` blocks and Plugin Framework `identityschema` blocks) were intentionally left untouched, as they are not user-facing.

Remember to:
- [x] Update the Change Log
- [x] Update the Documentation — no doc changes required; descriptions are sourced from the existing website docs, which remain the source of truth.

## Testing plan

Regarding the beneficial output:
1. Generate a copy of the provider schema (e.g. `terraform providers schema -json`).
2. Confirm that the affected resources now have a component-level description, and that their attributes have descriptions associated with them. For resources with nested subcomponents, confirm instead that the information is present in the source for later use.

Regarding non-negative changes:
1. Run the existing acceptance tests against the affected resources.
2. This is a schema-metadata-only change with no behavioral impact, so no new tests were added and no runtime code paths were altered.

## External links

- Related PR (data source descriptions backfill): [Backfill of descriptions into component schemas for data sources, ephemerals, and actions](https://github.com/hashicorp/terraform-provider-tfe/pull/2125)

[Resource Backfill Acceptance Test Output for PR.rtf](https://github.com/user-attachments/files/29763143/Resource.Backfill.Acceptance.Test.Output.for.PR.rtf)

## Rollback Plan

If these changes need to be reverted, the commit can be rolled back directly. This is a documentation/metadata-only change with no runtime, state, or API impact, so a revert carries no migration or data risk.

None. This change only adds descriptive metadata to schema attributes and does not affect access controls, encryption, logging, or any other security control.


ctrombley (commented):
@marzuqmazid-ibm excellent job, this is looking really good. 

One thing I wanted to ask about - it seems like we didn't backfill resource-level descriptions, only attribute-level ones. Is that intentional? Or should we tackle that as well in this PR?

internal/provider/resource_tfe_workspace_run.go
```diff
@@ -155,37 +160,44 @@ func resourceTFEWorkspaceRunSchema() *schema.Resource {
 	return &schema.Resource{
 		Schema: map[string]*schema.Schema{
 			"manual_confirm": {
-				Type:     schema.TypeBool,
-				Required: true,
+				Description: "If set to true a human will have to manually confirm a plan in HCP Terraform's UI to start an apply. If set to false, this resource will be automatically applied. Defaults to false.",
```
ctrombley: If `manual_confirm` is `Required: true`, I think that means we can remove the "Defaults to false":

```suggestion
				Description: "If set to true a human will have to manually confirm a plan in HCP Terraform's UI to start an apply. If set to false, this resource will be automatically applied.",
```

ctrombley (dismissed):
Nice work @marzuqmazid-ibm! Approved with one minor suggestion.

ctrombley (approved):
Looks good, thanks @marzuqmazid-ibm!

PR #2134, opened 2026-07-10
hyok_enabled should be a computed attr on workspaces

JarrettSpiker:
## Description

Fixes a bug where enabling HYOK on a provider-managed workspace causes drift to show up in runs. The intended flow for enabling HYOK (for the time being) is to create your workspace and then manually enable HYOK either though the UI or API. However, this would result in the next terraform run attempting to set hyok_enabled back to false...which is cannot do, because the provider does not actually try to modify that attribute.

It was meant to be computed (read-only), not optional. 

Currently it is not possible to enable HYOK using the provider. This is something we intend to tackle separately, using a different resource for HYOK enablement.

_Remember to:_

## Testing plan

1. Test that `hyok_enabled` can no longer be set on the workspace resource
2. Test that drift doesn't occur when hyok is enabled on a workspace outside of the provider
3. Test that other resources can read the `tfe_workspace.whatever.hyok_enabled` attribute properly

## External links
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#request-body)

```
❯ go test -test.fullpath=true -timeout 3000s -run ^TestAccTFEWorkspace_HYOKEnabled$ github.com/hashicorp/terraform-provider-tfe/internal/provider
ok  	github.com/hashicorp/terraform-provider-tfe/internal/provider	2.003s
...
```

## Rollback Plan
We can revert the PR. The attribute already doesnt do anything

None
-->

### DIFF
--- a/internal/provider/resource_tfe_workspace.go
+++ b/internal/provider/resource_tfe_workspace.go
@@ -415,7 +415,7 @@ func resourceTFEWorkspace() *schema.Resource {
 			},
 			"hyok_enabled": {
 				Type:        schema.TypeBool,
-				Optional:    true,
+				Computed:    true,
 				Description: "Whether HYOK (Hold Your Own Key) is enabled for the workspace. Available only in HCP Terraform.",
 			},
 		},
--- a/website/docs/cdktf/python/d/workspace.html.markdown
+++ b/website/docs/cdktf/python/d/workspace.html.markdown
@@ -53,7 +53,7 @@ In addition to all arguments above, the following attributes are exported:
 * `environment` - The environment of the workspace.
 * `file_triggers_enabled` - Indicates whether runs are triggered based on the changed files in a VCS push (if `true`) or always triggered on every push (if `false`).
 * `global_remote_state` - (Optional) Whether the workspace should allow all workspaces in the organization to access its state data during runs. If false, then only specifically approved workspaces can access its state (determined by the `remote_state_consumer_ids` argument). Cannot be true if `project_remote_state` is true.
-* `hyok_enabled` - (Optional) Whether HYOK is enabled for the workspace.
+* `hyok_enabled` - (Available only in HCP Terraform) Whether HYOK is enabled for the workspace.
 * `inherits_project_auto_destroy` - Indicates whether this workspace inherits project auto destroy settings.
 * `locked` - Indicates whether the workspace is locked.
 * `remote_state_consumer_ids` - (Optional) A set of workspace IDs that will be set as the remote state consumers for the given workspace. Cannot be used if `global_remote_state` or `project_remote_state` is set to `true`.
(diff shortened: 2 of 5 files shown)


helenjw (dismissed):
👌

website/docs/cdktf/python/d/workspace.html.markdown
```diff
@@ -53,7 +53,7 @@ In addition to all arguments above, the following attributes are exported:
 * `environment` - The environment of the workspace.
 * `file_triggers_enabled` - Indicates whether runs are triggered based on the changed files in a VCS push (if `true`) or always triggered on every push (if `false`).
 * `global_remote_state` - (Optional) Whether the workspace should allow all workspaces in the organization to access its state data during runs. If false, then only specifically approved workspaces can access its state (determined by the `remote_state_consumer_ids` argument). Cannot be true if `project_remote_state` is true.
-* `hyok_enabled` - (Optional) Whether HYOK is enabled for the workspace.
+* `hyok_enabled` - (Available only in HCP Terraform) Whether HYOK is enabled for the workspace.
```
shwetamurali: Heads up that the equivalent cdktf docs for the other languages weren't updated and still say Optional specifically `website/docs/cdktf/typescript/d/workspace.html.markdown` (line ~59) and `website/docs/cdktf/java/d/workspace.html.markdown` (line ~58). Since these` cdktf/*` docs are auto-generated, it'd be better to regenerate them so all languages stay in sync rather than hand-editing only the python copy otherwise they'll drift

internal/provider/resource_tfe_workspace.go
```diff
@@ -415,7 +415,7 @@ func resourceTFEWorkspace() *schema.Resource {
 			},
 			"hyok_enabled": {
 				Type:        schema.TypeBool,
-				Optional:    true,
+				Computed:    true,
```
shwetamurali, line 418: Changing this will error any existing config that explicitly set `hyok_enabled`. I understand that we don't want people to set it themselves so just confirming this is desired behavior

JarrettSpiker: yes, unfortunately that is the desired behaviour. I expect that this will have very little customer impact based on:
1. setting this attr in the config not actually _doing_ anything right now
2. the slow rollout of HYOK across adopting orgs, and the need to enable it in the UI. This bug was actually found by a customer trying to start utilizing the provider to roll out more broadly

I will update the Changelog to make it clear though that an error could occur after upgrade, and how to fix it.

website/docs/cdktf/python/d/workspace.html.markdown
JarrettSpiker: Thanks for catching this...TIL about autogenerating these docs!

shwetamurali (approved):

shwetamurali (approved):

PR #2140, opened 2026-07-14
Fix inconsistent sensitive token result for project notification config

Maed223:
## Description

`tfe_project_notification_configuration` without the `token` attribute provided fails during update:

```
Error: Provider produced inconsistent result after apply
When applying changes to tfe_project_notification_configuration.slack["prj-xuA7Y1sogoBLKHto"], provider "provider[\"registry.terraform.io/hashicorp/tfe\"]" produced an unexpected new value: .token: inconsistent values for sensitive attribute.

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

`token` is optional and non-computed, so the post apply state value is required to equal the planned value (`null` when the user omits it). Previously the result defaulted to`types.StringValue("")`, so without a given token attribute the provider returns `""` while the plan shows `null`. The fix here is to remove the empty string default so the optional token defaults to null.

A set of fixes in the same vain went in for `triggers`, `email_addresses`, and `email_user_ids` attributes to properly handle empty/null cases which were surfaced by running the test suite locally. Previously they produced errors on apply.

## Testing

Added an integration test, also made some testing changes to get the existing tests passing (failures were hidden under `skipUnlessBeta`). The separate changes consist of:
1. Removing assertions of the token value against the read call, the API doesn't return the token value
2. Use a valid project trigger, previously used `change_request:created` which isn't valid for the project level.

```
=== RUN   TestAccTFEProjectNotificationConfiguration_basic
--- PASS: TestAccTFEProjectNotificationConfiguration_basic (3.69s)
=== RUN   TestAccTFEProjectNotificationConfiguration_emailUserIDs
--- PASS: TestAccTFEProjectNotificationConfiguration_emailUserIDs (2.25s)
=== RUN   TestAccTFEProjectNotificationConfiguration_update
--- PASS: TestAccTFEProjectNotificationConfiguration_update (3.24s)
=== RUN   TestAccTFEProjectNotificationConfiguration_validateSchemaAttributesSlack
--- PASS: TestAccTFEProjectNotificationConfiguration_validateSchemaAttributesSlack (1.15s)
=== RUN   TestAccTFEProjectNotificationConfiguration_slack
--- PASS: TestAccTFEProjectNotificationConfiguration_slack (2.22s)
...
```

### DIFF
--- a/internal/provider/resource_tfe_project_notification_configuration.go
+++ b/internal/provider/resource_tfe_project_notification_configuration.go
@@ -73,6 +73,10 @@ func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfigurati
 	}
 
 	if len(v.EmailAddresses) == 0 {
+		// email_addresses is optional and computed, so returning an empty set
+		// (rather than null) is accepted post-apply for both an explicit empty
+		// set and an omitted value. This differs from triggers, which is not
+		// computed and therefore must echo the exact planned value.
 		result.EmailAddresses = types.SetValueMust(types.StringType, []attr.Value{})
 	} else {
 		emailAddresses, diags := types.SetValueFrom(ctx, types.StringType, v.EmailAddresses)
@@ -96,6 +100,9 @@ func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfigurati
 	}
 
 	if len(v.EmailUsers) == 0 {
+		// email_user_ids is optional and computed, so an empty set is accepted
+		// post-apply for both an explicit empty set and an omitted value (see
+		// the email_addresses note above).
 		result.EmailUserIDs = types.SetValueMust(types.StringType, []attr.Value{})
 	} else {
 		emailUserIDs := make([]attr.Value, len(v.EmailUsers))
--- a/internal/provider/resource_tfe_project_notification_configuration.go
+++ b/internal/provider/resource_tfe_project_notification_configuration.go
@@ -61,7 +61,7 @@ type modelTFEProjectNotificationConfiguration struct {
 
 // modelFromTFEProjectNotificationConfiguration builds a modelTFEProjectNotificationConfiguration
 // struct from a tfe.NotificationConfiguration value.
-func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfiguration, tokenWOVersion types.Int64, lastValue types.String) (*modelTFEProjectNotificationConfiguration, diag.Diagnostics) {
+func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfiguration, tokenWOVersion types.Int64, lastValue types.String, priorTriggers types.Set) (*modelTFEProjectNotificationConfiguration, diag.Diagnostics) {
 	var diags diag.Diagnostics
 	result := modelTFEProjectNotificationConfiguration{
 		ID:              types.StringValue(v.ID),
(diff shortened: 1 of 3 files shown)


internal/provider/resource_tfe_project_notification_configuration.go
```diff
@@ -402,7 +404,7 @@ func (r *resourceTFEProjectNotificationConfiguration) Read(ctx context.Context,
 		return
 	}
 
-	result, diags := modelFromTFEProjectNotificationConfiguration(pnc, state.TokenWOVersion, state.Token)
+	result, diags := modelFromTFEProjectNotificationConfiguration(pnc, state.TokenWOVersion, state.Token, state.Triggers)
```
shwetamurali, line 414: When the API returns zero triggers, Read sets `result.Triggers = state.Triggers` (the previous value) instead of reflecting what the API actually returned. This means if someone deletes the triggers outside of Terraform, a terraform plan won't detect the drift: and refresh copies the old value back into state, so config and state still match and nothing gets corrected.
I think this is intentional, because the API returns "0 triggers" for both an explicit empty set and an unset/null value. If Read returned null in the zero case instead, a user who configured triggers = [] would get a permanent diff. Echoing the prior state value avoids that permanent-diff problem, at the cost of not detecting genuine out-of-band deletions.
Is this the intended behavior?

shwetamurali:
The identical `Token: types.StringValue("")` default is present in `resource_tfe_team_notification_configuration.go:73`, which will hit the same error when a token isn't provided. Can we do these fixes here too, or open a ticket for that

internal/provider/resource_tfe_project_notification_configuration.go
```diff
@@ -70,11 +70,10 @@ func modelFromTFEProjectNotificationConfiguration(v *tfe.NotificationConfigurati
 		Enabled:         types.BoolValue(v.Enabled),
 		ProjectID:       types.StringValue(v.SubscribableChoice.Project.ID),
 		TokenWOVersion:  tokenWOVersion,
-		Token:           types.StringValue(""),
 	}
 
 	if len(v.EmailAddresses) == 0 {
-		result.EmailAddresses = types.SetNull(types.StringType)
+		result.EmailAddresses = types.SetValueMust(types.StringType, []attr.Value{})
```
shwetamurali, line 80: triggers now has a helpful comment explaining why it preserves the prior value, but email_addresses/email_user_ids use a hardcoded empty set with no explanation. A similar note saying that these differ because they're computed so any value is accepted post-apply whereas triggers is not would prevent someone trying to fix the inconsistency later

markdecrane: Updated to add explanatory comments

markdecrane:
@shwetamurali I'll make a ticket to port the Token fix to the other applicable resources since the scope of this PR already started to creep in finding other issues.

internal/provider/resource_tfe_project_notification_configuration.go
markdecrane: Yeah this is an intentional tradeoff I made since this was a fly by fix. I figured that an external removal of all the triggers would be unlikely, and in any case would sort itself out with another apply. We could go a couple of different routes like making it computed, but figured this got us around the immediate issue without making a semantic change here.

shwetamurali (dismissed):

shwetamurali (approved):

PR #2145, opened 2026-07-16
tfe_workspace_run: Configuration version missing

### LINKED ISSUE #1939: `tfe_workspace_run` fail for empty workspace
#### Terraform version

```plaintext
v1.14.3
```

```terraform
resource "tfe_workspace" "this" {
  name              = "..."
  organization      = data.tfe_organization.this.name
  project_id        = data.tfe_project.this.id
  terraform_version = ">= 1.12, < 2.0"
}

resource "tfe_workspace_run" "this" {
  workspace_id = tfe_workspace.this.id

  destroy {
    manual_confirm    = false
    retry             = true
    retry_attempts    = 3
    retry_backoff_min = 10
    retry_backoff_max = 30
    wait_for_run      = true
  }
}

```

#### Debug Output

```plaintext
│ Error: error creating run for workspace ws-7xTk3aNBB......: unprocessable entity
│ 
│ Configuration version is missing
```

#### Expected Behavior

The destroy workspace run should execute successfully and must not fail even when the workspace contains no resources.
Use case:

1. We create a workspace along with a tfe_workspace_run and several other resources.
2. Terraform code that uses this workspace is applied from a separate CI/CD pipeline.
3. In some edge cases (unhappy paths), the apply step in (2) fails, leaving the workspace empty (no Terraform resources in the state).
4. As part of the workspace lifecycle, we then remove the workspace created in (1). This triggers a destroy run, as expected. However, the destroy run fails with the error `Configuration version is missing.`
While this behavior is technically expected when no resources exist, there should be an option to suppress this error and allow the destroy run to succeed.

#### Actual Behavior

`destroy` run of `tfe_workspace_run` fails with `Configuration version is missing`.

shwetamurali:
## Description
Fixes https://github.com/hashicorp/terraform-provider-tfe/issues/1939

`tfe_workspace_run` destroy runs fail with `Configuration version is missing` when the target workspace is empty (i.e. it never had a configuration version uploaded). This blocks the workspace lifecycle in scenarios where an upstream apply failed and left the workspace with no resources and no configuration version, making it impossible to cleanly destroy the workspace.

This PR adds a new optional boolean argument `allow_config_version_missing` to both the `apply` and `destroy` blocks of `tfe_workspace_run`. When set to `true`, if the run cannot be created because the configuration version is missing, the operation is treated as a no-op success instead of returning an error.

The default behavior is unchanged (`false`), so this is fully opt-in and backward compatible.

## Before / After

**Before**: destroying an empty workspace using `tfe_workspace_run` fails:

~~~hcl
resource "tfe_workspace_run" "this" {
  workspace_id = tfe_workspace.this.id

  destroy {
    manual_confirm = false
    wait_for_run   = true
  }
}
~~~

~~~
│ Error: error creating run for workspace ws-7xTk3aNBB......: unprocessable entity
│
│ Configuration version is missing
~~~

**After**: with `allow_config_version_missing = true`, the destroy run is treated as a no-op success and the workspace is destroyed cleanly:

~~~hcl
resource "tfe_workspace_run" "this" {
  workspace_id = tfe_workspace.this.id

  destroy {
    manual_confirm               = false
    wait_for_run                 = true
    allow_config_version_missing = true
  }
}
~~~

~~~
tfe_workspace_run.this: Destroying...
tfe_workspace_run.this: Destruction complete after 1s
~~~

## Testing plan

1. Create a workspace with a `tfe_workspace_run` resource that has a `destroy` block with `allow_config_version_missing = true` (config below).
2. Ensure the workspace never receives a configuration version. Do **not** run an apply against it (this reproduces the "empty workspace / failed upstream apply" edge case). Create the workspace with `queue_all_runs = false` and no configuration uploaded.
3. Run `terraform apply` to create the resources. The `tfe_workspace_run` resource is created (the `destroy` block only takes effect on destruction).
4. Run `terraform destroy` (or remove the `tfe_workspace_run` from config and apply). This triggers a destroy run.
5. **Expected:** the destroy run does not error with `Configuration version is missing`; instead the provider logs that the configuration version is missing and treats it as a no-op success, allowing the destroy to complete.
6. Repeat the same steps **without** `allow_config_version_missing` (or with it set to `false`) and confirm the original `Configuration version is missing` error is still returned — verifying the default behavior is unchanged.

Example Terraform config:

~~~hcl
resource "tfe_organization" "this" {
  name  = "my-org-name"
  email = "admin@company.com"
}

resource "tfe_workspace" "this" {
  name           = "empty-ws"
  organization   = tfe_organization.this.name
  queue_all_runs = false
}

resource "tfe_workspace_run" "this" {
  workspace_id = tfe_workspace.this.id

  destroy {
    manual_confirm               = false
    wait_for_run                 = true
    allow_config_version_missing = true
  }
}
~~~

## Changes

- `internal/provider/resource_tfe_workspace_run.go`: Added the `allow_config_version_missing` schema field (default `false`).
- `internal/provider/workspace_run_helpers.go`: When `createRun` errors and the flag is set, detect the missing-config-version error via a new `isConfigVersionMissingErr` helper, log, set a placeholder ID, and return success.
- `internal/provider/workspace_run_helpers_test.go`: Added a unit test covering the error-matching helper.
- `website/docs/r/workspace_run.html.markdown`: Documented the new argument.

`tfe_workspace_run` is a legacy SDKv2 resource (per the header comment in `resource_tfe_workspace_run.go`). This change follows the existing patterns rather than migrating the resource to the Plugin Framework, since it is a targeted bug fix.

## Test output

Unit test (`isConfigVersionMissingErr`):

~~~
$ go test ./internal/provider/ -run TestIsConfigVersionMissingErr -v
=== RUN   TestIsConfigVersionMissingErr
=== RUN   TestIsConfigVersionMissingErr/nil_error
=== RUN   TestIsConfigVersionMissingErr/unrelated_error
=== RUN   TestIsConfigVersionMissingErr/configuration_version_missing_error
=== RUN   TestIsConfigVersionMissingErr/lowercase_configuration_version_missing_error
--- PASS: TestIsConfigVersionMissingErr (0.00s)
    --- PASS: TestIsConfigVersionMissingErr/nil_error (0.00s)
    --- PASS: TestIsConfigVersionMissingErr/unrelated_error (0.00s)
    --- PASS: TestIsConfigVersionMissingErr/configuration_version_missing_error (0.00s)
    --- PASS: TestIsConfigVersionMissingErr/lowercase_configuration_version_missing_error (0.00s)
PASS
ok  	github.com/hashicorp/terraform-provider-tfe/internal/provider	0.862s
~~~

## External links

- [JIRA](https://hashicorp.atlassian.net/browse/TF-32999)

~~~
$ TF_ACC=1 ENABLE_BETA=1 go test ./internal/provider/ -run TestAccTFEWorkspaceRun -v
~~~


brandonc (commented):
Interesting. In terraform, we normally allow destroy actions to be no-op if the resource  cannot be found, usually meaning it was already destroyed. This is a similar situation. What if the `tfe_workspace_run` destroy didn't do anything if the error returned was `Configuration version is missing` because that would indicate there is nothing to destroy anyway?

I can't think of a reason why you'd want to do what `allow_config_version_missing = false` implies: You'd have to apply a config for it to be able to destroy!

brandonc (approved):

PR #2148, opened 2026-07-20
Add support for managing IP allowlists

nickmcdonnough:
## Description

This PR adds the resource `ip_allowlist` allowing organization owners to manage their IP allowlists. Users can manage the relationship between IP allowlists and agent pools (when the list is not scoped to the organization).

_Remember to:_

## Testing plan

1.  Create a new IP allowlist (org scoped by default)
1.  Create a second IP allowlist (it will need to be scoped to `specific_agent_pools` since org scoped already exists)
1.  Try updating attributes of the created lists (name, description, enabled boolean)
1.  Try updating agent pool assignments

```
# create an ip allowlist
resource "tfe_ip_allowlist" "example" {
  organization      = "my-org-name"
  name              = "corporate-network"
  description       = "Allowlist for the corporate network"
  enforcement_scope = "organization"

  cidr_range = [
    {
      range       = "10.0.0.0/16"
      description = "Corporate LAN"
    },
    {
      range       = "192.168.1.0/24"
      description = "VPN"
      enabled     = false
    },
  ]
}

# create an ip allowlist and assign it to an agent pool
data "tfe_agent_pool" "example" {
  organization = "my-org-name"
}

resource "tfe_ip_allowlist" "example" {
  organization      = "my-org-name"
  name              = "agent-pool-allowlist"
  description       = "Only enforced for the selected agent pool"
  enforcement_scope = "selected_agent_pools"
  agent_pool_ids    = [data.tfe_agent_pool.example.id]

  cidr_range = [
    {
      range       = "203.0.113.0/24"
      description = "Datacenter egress"
    },
  ]
}
```

## External links

- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/cidr-range-lists)

```
  === RUN   TestAccTFEIPAllowlist_basic
  --- PASS: TestAccTFEIPAllowlist_basic (9.76s)
```

## Rollback Plan
This code can be removed without affecting other resources.

?


internal/provider/resource_tfe_ip_allowlist.go
```diff
@@ -0,0 +1,581 @@
...
+		return
+	}
+	if err := r.reconcileRanges(ctx, listID, planRanges); err != nil {
+		resp.Diagnostics.AddError("Error updating IP allowlist CIDR ranges", err.Error())
+		return
+	}
+
+	// Reconcile agent pool assignments. For non-selected scopes the API clears
+	// assignments automatically when the scope changes, so only reconcile when
+	// the scope is selected_agent_pools.
```
markdecrane: I think we should include the agent relationships in the PATCH here (`data.SetRelationships(...)`) instead of doing the post-write reconciliation.

Separate but connected we ought to have a test for the `selected_agent_pools` and `organization` scope here.

internal/provider/data_source_tfe_ip_allowlist.go
```diff
@@ -0,0 +1,164 @@
...
+			"enforcement_scope": schema.StringAttribute{
+				Description: "Where the IP allowlist is enforced. One of `organization`, `all_agent_pools`, or `selected_agent_pools`.",
+				Computed:    true,
+			},
+			"agent_pool_ids": schema.SetAttribute{
+				Description: "The IDs of the agent pools the IP allowlist applies to.",
+				Computed:    true,
+				ElementType: types.StringType,
+			},
+			"cidr_ranges": schema.SetNestedAttribute{
```
markdecrane: We should have the attribute be `cidr_range` or change the model tag to `cidr_ranges`, currently there's a mismatch. Currently with the mismatch cidr ranges would be perserved.

nickmcdonnough:
@markdecrane I think this is ready for re-review.

markdecrane (commented):
I took a look at how CIDR ranges are handled in the API, and it looks like host bits are masked when returned. The problem that poses here is that we'll have a permanent diff pop up on refresh of the resource since we're setting the state value directly from the plan rather than reading back the response. 

Caught this doing a closer second pass, but any reason we're largely assigning the plan values to state rather than the responses? Typically state comes from API responses, it's hard to guarantee that plan values will ever be that authoritative state.

nickmcdonnough:
@markdecrane there was no good reason for that and so I changed it to fetch the record after create and update. I think this more closely fits existing patterns and keeps state in line with what is stored in HCPT.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Unreleased
 
+FEATURES:
+* **New Resource:** `r/tfe_ip_allowlist` and **New Data Source:** `d/tfe_ip_allowlist`: Adds a resource and data source to manage organization IP allowlists (CIDR range lists), including per-range enablement and agent pool scoping. Uses the go-tfe v2 client.
```
markdecrane, line 4: Small note, we typical append an attribute in the style of `@author [#PR_NUM](PR_LINK)`

internal/provider/resource_tfe_ip_allowlist.go
```diff
@@ -0,0 +1,565 @@
...
+
+	listID := *created.GetData().GetId()
+
+	// Assign agent pools for the selected_agent_pools scope.
+	if plan.EnforcementScope.ValueString() == ipAllowlistScopeSelectedAgentPools {
+		desired := setToStringSlice(ctx, plan.AgentPoolIDs, &resp.Diagnostics)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		if err := r.reconcileAgentPools(ctx, listID, desired); err != nil {
```
markdecrane, line 251: Think we ought to update this to be inline with Update in how it embed it in the relationships

internal/provider/examples_acc_test.go
```diff
@@ -177,6 +177,12 @@ func skipReason(cfg string) (string, bool) {
 		strings.Contains(cfg, `"tfe_agent_pool_excluded_workspaces"`) {
 		return "uses agent pool resources unavailable in CI", true
 	}
+	// data "tfe_agent_pool" resolves during the pre-apply plan (its arguments are
+	// literal and it has no dependency on the harness's injected agent-pool
+	// fixture), so the lookup runs before that pool is created and returns 404.
+	if strings.Contains(cfg, `data "tfe_agent_pool"`) {
+		return "uses data.tfe_agent_pool, which is read at plan time before the injected agent-pool fixture exists (404)", true
+	}
```
markdecrane, line 185: To avoid having to add this here, we can instead use a resource  and reference it's `.id` in the tests that currently necesitate the exception.

internal/provider/data_source_tfe_ip_allowlist.go
```diff
@@ -0,0 +1,164 @@
...
+	if !found {
+		resp.Diagnostics.AddError(
+			"Could not find IP allowlist",
+			fmt.Sprintf("IP allowlist %s/%s not found", organization, name),
+		)
+		return
+	}
+
+	// Reuse the resource read logic to fully populate the model.
+	r := &resourceTFEIPAllowlist{config: d.config}
```
markdecrane, line 144: Reusing the resource logic is a bit of a non-standard pattern and we probably don't want to introduce such a coupling. You can find a pretty typical example of this in `data_source_project.go` where the `modelDataSourceTFE....` is defined on its own separate from the related resource.

internal/provider/ip_allowlist_helpers_test.go
```diff
@@ -0,0 +1,246 @@
...
+		expectErr bool
+	}{
+		{"valid ipv4 cidr", types.StringValue("10.0.0.0/24"), false},
+		{"valid single host", types.StringValue("192.168.1.1/32"), false},
+		{"not a cidr", types.StringValue("10.0.0.0"), true},
+		{"garbage", types.StringValue("not-a-cidr"), true},
+		{"ipv6 cidr rejected", types.StringValue("2001:db8::/32"), true},
+		{"null skipped", types.StringNull(), false},
+		{"unknown skipped", types.StringUnknown(), false},
+	}
```
ctrombley, line 227: Consider adding validation for non-canonical CIDRs, which can still cause drift since they are normalized before being returned by the api:



```suggestion
		{"host bits rejected", types.StringValue("10.0.0.5/16"), true},
		{"ipv4-mapped ipv6 cidr rejected", types.StringValue("::ffff:192.0.2.1/120"), true},
    }
```

ctrombley (changes requested):

PR #2149, opened 2026-07-20
migrate tfe_workspace_run_task resource and data source to go-tfe v2

ctrombley:
## Description
This PR migrates the workspace_run_task resource and data source to use the generated go-tfe v2 client. The goal here is API/client migration only, keeping existing behavior intact while swapping the underlying calls.

## Notes
- Import format is unchanged: `<organization>/<workspace_name>/<task_name>`.
- Name-to-ID resolution behavior is unchanged (still paginates where required).

### DIFF
--- a/internal/provider/data_source_workspace_run_task.go
+++ b/internal/provider/data_source_workspace_run_task.go
@@ -7,10 +7,12 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


marzuqmazid-ibm (approved):
Looks good! Ready to merge!

PR #2154, opened 2026-07-23
Add provider set priority support

AadarshIBM:
## Description

Adds support for the Provider Set `priority` attribute in the TFE provider.

This PR:

- Adds `priority` to the `tfe_provider_set` resource, with a default of `false`.
- Adds `priority` to the `tfe_provider_set` data source.
- Sends both `priority = true` and `priority = false` to the API.
- Reads the priority value back into Terraform state.
- Adds unit and acceptance test coverage.
- Updates the example, documentation, and changelog.

The Provider Set migration to `go-tfe/v2` was completed in #2158. This PR builds on that work and only adds priority support.

The backend remains responsible for deciding which Provider Set takes precedence.

- [x] Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)

## Testing plan

1. Verify that omitting `priority` results in `false`.
2. Set `priority = true` and confirm it is sent and read back correctly.
3. Update `priority` from `true` back to `false`.
4. Confirm priority changes update the existing Provider Set instead of replacing it.
5. Verify that the Provider Set data source reads the priority value.
6. Verify that existing relationships and write-only HCL behavior remain unchanged.

## External links

- Jira: https://hashicorp.atlassian.net/browse/NOCODE-107
- Related Provider Set v2 migration: https://github.com/hashicorp/terraform-provider-tfe/pull/2158

## Test results

Focused Provider Set tests passed using the repository's committed dependencies:

```sh
GOWORK=off go test ./internal/provider \
  -run '^Test(ProviderSet.*|NewProviderSet.*|ModelFromTFEProviderSet|TFEProviderSetNotImportableInPrepStub|FrameworkProviderResources_includeProviderSet)$' \
  -count=1
```

The following checks also passed:

```sh
GOWORK=off go vet ./...
GOWORK=off make fmtcheck
git diff origin/main...HEAD --check
```

Additional priority tests and race testing also passed.

The following acceptance tests passed against a feature-enabled staging organization:

- `TestAccTFEProviderSet_priority`
  - Confirmed that omitted `priority` defaults to `false`.
  - Confirmed that priority can be updated from `false` to `true` and back to `false`.
  - Confirmed that priority changes update the existing Provider Set instead of replacing it.

- `TestAccTFEProviderSetDataSource_read`
  - Confirmed that the data source reads `priority = true`.

The staging organization was used through a temporary local test change. That change was reverted and is not included in this PR.

## Rollback Plan

To roll back this change, revert the PR. No state migration is required.

No changes to access controls, authorization, encryption, or logging.

### DIFF
--- a/internal/provider/resource_tfe_provider_set.go
+++ b/internal/provider/resource_tfe_provider_set.go
@@ -48,6 +48,7 @@ type modelTFEProviderSet struct {
 	Name           types.String `tfsdk:"name"`
 	Description    types.String `tfsdk:"description"`
 	Global         types.Bool   `tfsdk:"global"`
+	Priority       types.Bool   `tfsdk:"priority"`
 	Organization   types.String `tfsdk:"organization"`
 	WorkspaceIDs   types.Set    `tfsdk:"workspace_ids"`
 	ProjectIDs     types.Set    `tfsdk:"project_ids"`
@@ -99,6 +100,12 @@ func (r *resourceTFEProviderSet) Schema(ctx context.Context, req resource.Schema
 			},
 			"global": schema.BoolAttribute{
 				Description: "Whether the provider set applies globally.",
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(false),
+			},
+			"priority": schema.BoolAttribute{
+				Description: "Whether the provider set takes priority over provider sets with more specific scopes. Defaults to false.",
 				Optional:    true,
 				Computed:    true,
 				Default:     booldefault.StaticBool(false),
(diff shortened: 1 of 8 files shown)


internal/provider/resource_tfe_provider_set_test.go
```diff
@@ -665,6 +824,27 @@ EOT
 }`, organization)
 }
 
+func testAccTFEProviderSet_priority(organization string, priority *bool) string {
```
paladin-devops, line 827: 👍🏻

internal/provider/resource_tfe_provider_set.go
paladin-devops, line 1: The new test coverage looks great!

paladin-devops (approved):
LGTM! 🚀

ctrombley (approved):
This is looking good, thanks @AadarshIBM!

Issue #1257, opened 2024-02-20
Team token won't regenerate more than one time with force_regenerate to true.


emmleg:
#### Terraform Cloud/Enterprise version

Terraform Cloud

#### Terraform version

```plaintext
v1.7.3
```

#### Terraform Configuration Files

```terraform
resource "tfe_team" "test" {
  name = "test"
}

resource "tfe_team_token" "test" {
  team_id = tfe_team.test.id
  force_regenerate = true
}
```

#### Debug Output

```plaintext
https://gist.github.com/emmleg/bbc3d03d31e7fd1883a1dfec38678a42
```

#### Expected Behavior

The team token should regenerate each time the project run, according to the documentation.
#### Actual Behavior

The team token only regenerate one time, then never get regenerated after rerun the terraform project. It work only by switching between false and true.
#### Additional Context

Documentation say it should regenerate even if aready exist, but not mentionning if we can do it multiple times or not.

shaunakone:
Thanks for the report! After digging in, the behavior you're seeing is actually how `force_regenerate` is intended to work — the docs just weren't clear about it, and we're updating them in #2160.

`force_regenerate` triggers regeneration only when its *value changes* — it forces the resource to be replaced on change. It isn't designed to regenerate the token on every apply, which is why it worked once and then only again when you toggled it.

To regenerate on your own terms, use Terraform's native replacement mechanisms — a one-off with `terraform apply -replace="tfe_team_token.test"`, or scheduled rotation:

```hcl
resource "time_rotating" "token_rotation" {
  rotation_days = 30
}

resource "tfe_team_token" "test" {
  team_id = tfe_team.test.id

  lifecycle {
    replace_triggered_by = [time_rotating.token_rotation.id]
  }
}
```

Closing as working-as-intended — feel free to reopen if this doesn't cover your use case.

PR #2160, opened 2026-07-28
docs: clarify force_regenerate behavior on tfe_team_token

shaunakone:
## Description

Reorients this PR to a documentation clarification, per review feedback.

The `force_regenerate` attribute on `tfe_team_token` works as intended: it triggers token regeneration when its value *changes* (via the `RequiresReplace` plan modifier), not on every apply. Rather than changing resource behavior, this PR clarifies the attribute's documentation and points users to Terraform's native replacement mechanisms (`terraform apply -replace`, `replace_triggered_by`) for on-demand or scheduled token rotation.

Since docs are now generated by tfplugindocs, the wording lives in the `force_regenerate` `Description` field in `internal/provider/resource_tfe_team_token.go`, with `website/docs/r/team_token.html.markdown` regenerated to match. No resource behavior is changed; the only Go edit is to a description string.

- [ ] ~Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)~ - not needed; no user-facing behavior change

## Testing plan

1. No behavioral change; the only Go edit is the `Description` string on `force_regenerate`.
2. `docgen generate-docs` and `docgen validate` pass in CI, confirming the committed markdown matches generator output.

## External links

- Related issue: #1257

No acceptance tests apply; this change does not alter resource behavior.

## Rollback Plan

If this change needs to be reverted, we will roll out an update to the code within 7 days.

None. This change only edits a schema description string and its generated documentation.

### DIFF
--- a/internal/provider/resource_tfe_team_token.go
+++ b/internal/provider/resource_tfe_team_token.go
@@ -89,7 +89,7 @@ func (r *resourceTFETeamToken) Schema(_ context.Context, _ resource.SchemaReques
 				},
 			},
 			"force_regenerate": schema.BoolAttribute{
-				Description: "Only applies to legacy tokens without descriptions. If set to `true`, a new token will be generated even if a token already exists. This will invalidate the existing token! Regeneration is triggered only when this value changes, so leaving it set to `true` will not regenerate the token on every apply. To regenerate on demand, replace the resource with `terraform apply -replace=\"tfe_team_token.<name>\"`, or use the `replace_triggered_by` lifecycle argument to tie regeneration to changes in another resource. This cannot be set with `description`.",
+				Description: "Only applies to legacy tokens without descriptions. If set to `true`, a new token will be generated even if a token already exists. This will invalidate the existing token! Regeneration is triggered only when this value changes.",
 				Optional:    true,
 				PlanModifiers: []planmodifier.Bool{
 					boolplanmodifier.RequiresReplace(),
--- a/internal/provider/resource_tfe_team_token.go
+++ b/internal/provider/resource_tfe_team_token.go
@@ -89,7 +89,7 @@ func (r *resourceTFETeamToken) Schema(_ context.Context, _ resource.SchemaReques
 				},
 			},
 			"force_regenerate": schema.BoolAttribute{
-				Description: "Only applies to legacy tokens without descriptions. If set to `true`, a new token will be generated even if a token already exists. This will invalidate the existing token! This cannot be set with `description`.",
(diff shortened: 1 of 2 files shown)


Maed223 (commented):
Looking at the original GH issue, the documentation, and now the resource/test code, I think this is primarily an issue of [the documentation](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/team_token#:~:text=force_regenerate%20%2D%20(Optional)%20Only%20applies%20to%20legacy%20tokens%20without%20descriptions.%20If%20set%20to%20true%2C%20a%20new%20token%20will%20be%20generated%20even%20if%20a%20token%20already%20exists.%20This%20will%20invalidate%20the%20existing%20token!%20This%20cannot%20be%20set%20with%20description.) being imprecise. The GH issue states the expectation that with `force_regenerate=true` that the token would be regenerated on each apply (and the documentation is seemingly in agreement). Though looking at the code/tests here that doesn't seem to be the intention of the attribute. The use of `boolplanmodifier.RequiresReplace()` on the `PlanModifiers` suggests that it's a flag that, _when changed_, forces the resource replacement and thus the token itself. 

I don't think we need to meet that behavioral expectation in the resource attribute itself since the same thing can be done natively using a `lifecycle` block. I'd suggest reorienting this PR to making the documentation more precise to better reflect the intention of the attribute, and otherwise responding to GH issue pointing them in the direction to get the behavior they want out of the resource.

Otherwise, looks like you accidentally commited some work around adding a `current_user` datasource, we ought to have that in a separate PR.

shaunakone:
Screenshot 2026-07-31 at 10 28 44 AM
Screenshot 2026-07-31 at 10 28 51 AM

brandonc (dismissed):

internal/provider/resource_tfe_team_token.go
```diff
@@ -89,7 +89,7 @@ func (r *resourceTFETeamToken) Schema(_ context.Context, _ resource.SchemaReques
 				},
 			},
 			"force_regenerate": schema.BoolAttribute{
-				Description: "Only applies to legacy tokens without descriptions. If set to `true`, a new token will be generated even if a token already exists. This will invalidate the existing token! This cannot be set with `description`.",
+				Description: "Only applies to legacy tokens without descriptions. If set to `true`, a new token will be generated even if a token already exists. This will invalidate the existing token! Regeneration is triggered only when this value changes, so leaving it set to `true` will not regenerate the token on every apply. To regenerate on demand, replace the resource with `terraform apply -replace=\"tfe_team_token.<name>\"`, or use the `replace_triggered_by` lifecycle argument to tie regeneration to changes in another resource. This cannot be set with `description`.",
```
markdecrane:
```suggestion
				Description: "Only applies to legacy tokens without descriptions. If set to `true`, a new token will be generated even if a token already exists. This will invalidate the existing token! Regeneration is triggered only when this value changes.",
```

I think this suffices, calling out specific workarounds doesn't properly fit the docs here.

markdecrane (approved):

PR #2177, opened 2026-08-04
Migration batch #5: variable resources to go-tfe v2

ctrombley:
Migrates two spec-independent variable surfaces to go-tfe v2 through `ConfiguredClient.ClientV2.API`:

- `resource_tfe_variable.go`: workspace and variable-set variable CRUD
- `data_source_variable_set.go`: variable-set lookup and read

Also updates go-tfe v2 to a revision that sends the required JSON:API content type for bodyless mutation requests.

Workspace lifecycle, workspace runs, outputs, workspace ID filtering, variable lists, variable-set associations, and test variables were removed from this PR. Those migrations are deferred until the relevant Atlas OpenAPI contracts are corrected and regenerated into go-tfe v2.

The deferred specification work is split between Atlas PR #30853 and the follow-up `ctrombley/more-spec-changes-2` branch.

### DIFF
--- a/internal/provider/data_source_outputs.go
+++ b/internal/provider/data_source_outputs.go
@@ -5,12 +5,13 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"math/big"
 	"reflect"
 
-	"github.com/hashicorp/go-tfe"
+	tfev2 "github.com/hashicorp/go-tfe/v2"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
@@ -39,10 +40,7 @@ type outputsModel struct {
 	NonSensitiveValues types.Dynamic `tfsdk:"nonsensitive_values"`
 }
 
-func modelFromOutputs(v *tfe.Workspace, sensitiveOutputs types.Dynamic, nonSensitiveOutputs types.Dynamic) outputsModel {
-	orgName := v.Organization.Name
-	wsName := v.Name
-
+func modelFromOutputs(orgName, wsName string, sensitiveOutputs types.Dynamic, nonSensitiveOutputs types.Dynamic) outputsModel {
 	return outputsModel{
 		ID:                 types.StringValue(fmt.Sprintf("%s-%s", orgName, wsName)),
 		Organization:       types.StringValue(orgName),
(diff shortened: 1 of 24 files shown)


internal/provider/data_source_outputs.go
```diff
@@ -119,56 +117,83 @@ func (d *outputsDataSource) Read(ctx context.Context, req datasource.ReadRequest
 		return
 	}
 
-	log.Printf("[DEBUG] Reading the workspace %s in organization %s", config.Workspace.ValueString(), orgName)
-	opts := &tfe.WorkspaceReadOptions{
-		Include: []tfe.WSIncludeOpt{tfe.WSOutputs},
```
brandonc: Why can't these be included in the new version? We're turning a single API fetch into two for each data source read.

internal/provider/resource_tfe_variable.go
```diff
@@ -355,32 +414,25 @@ func (r *resourceTFEVariable) createWithWorkspace(ctx context.Context, req resou
...
 		resp.Diagnostics.AddError(
 			"Error creating variable",
 			fmt.Sprintf("Couldn't create %s variable %s: %s", category, key, err.Error()),
 		)
 		return
 	}
-	// Got a variable back, so set state to new values
-	result := modelFromTFEVariable(*variable, data.Value, config.ValueWOVersion)
+
+	result := modelFromV2Var(varResp, data.Value, config.ValueWOVersion)
```
shwetamurali, line 435: The three variable-set paths backfill after `modelFromV2Var` but the three workspace paths rely entirely on v2VarWorkspaceID pulling `workspace_id` out of the response relationships. If the v2 vars api response doesnt populate the workspace relationship then workspace id would be null and u would always have a diff when u read it. So should we backfil `result.WorkspaceID = data.WorkspaceID` here and in the read/update workspace paths to match?

internal/provider/resource_tfe_variable.go
```diff
@@ -59,71 +60,132 @@ type modelTFEVariableIdentity struct {
...
+
+// v2VarUpdateEnvelope constructs a VarsEnvelope for PATCH requests (value optional).
+func v2VarUpdateEnvelope(key, description string, hcl, sensitive bool, value *string) models.VarsEnvelopeable {
+	attrs := models.NewVars_attributes()
+	attrs.SetKey(ptr(key))
+	attrs.SetDescription(ptr(description))
+	attrs.SetHcl(ptr(hcl))
+	attrs.SetSensitive(ptr(sensitive))
+	if value != nil {
+		attrs.SetValue(value)
```
shwetamurali, line 178: Can u confirm with an acceptance test that updates only the description of a sensitive variable and asserts the value is preserved? We should make sure that the write-only protection still works and it omits value unless it actually changes

PR #2186, opened 2026-08-11
fix: initialize pagination options in the tfe_variables data source

### LINKED ISSUE #2185: data.tfe_variables panics with a nil pointer dereference when the variable set or workspace has more than one page of variables

N/A — HCP Terraform (app.terraform.io)

#### Terraform version

```plaintext
v1.15.8
```

```terraform
# Crashes whenever the target variable set holds more than one page
# of variables (>20).
data "tfe_variable_set" "example" {
  organization = var.organization
  name         = "example-varset"
}

data "tfe_variables" "example" {
  variable_set_id = data.tfe_variable_set.example.id
}
```

#### Debug Output

```plaintext
Stack trace from the terraform-provider-tfe_v0.79.0_x5 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xddb3ae]

goroutine 1205 [running]:
github.com/hashicorp/terraform-provider-tfe/internal/provider.(*dataSourceTFEVariables).readFromVariableSet(0xc000c97860, {0x14252a8, 0xc0008f3260}, {{0x0, {0x0, 0x0}}, {0x0, {0x0, 0x0}}, {0x2, ...}, ...}, ...)
        github.com/hashicorp/terraform-provider-tfe/internal/provider/data_source_variables.go:295 +0xee
github.com/hashicorp/terraform-provider-tfe/internal/provider.(*dataSourceTFEVariables).Read(0xc000c97860, {0x14252a8, 0xc0008f3260}, ...)
        github.com/hashicorp/terraform-provider-tfe/internal/provider/data_source_variables.go:208 +0x273
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ReadDataSource(...)
        github.com/hashicorp/terraform-plugin-framework@v1.19.0/internal/fwserver/server_readdatasource.go:103 +0x76e
[...]

Error: The terraform-provider-tfe_v0.79.0_x5 plugin crashed!
```

#### Expected Behavior

`data.tfe_variables` returns all variables in the variable set, paginating as needed.

#### Actual Behavior

The provider panics with a nil pointer dereference at `data_source_variables.go:295`.

#### Additional Context

The pagination options are declared as a nil pointer and never allocated, so the
provider crashes the first time it needs to request a second page
([`data_source_variables.go#L258-L296`](https://github.com/hashicorp/terraform-provider-tfe/blob/v0.79.0/internal/provider/data_source_variables.go#L258-L296)):

```go
func (d *dataSourceTFEVariables) readFromVariableSet(ctx context.Context, config modelVariables, resp *datasource.ReadResponse) {
	var (
		options *tfe.VariableSetVariableListOptions   // nil, never initialized
		...
	)

	for {
		variableList, err := d.config.Client.VariableSetVariables.List(ctx, variableSetID, options)
		...
		// Exit the loop when we've seen all pages.
		if variableList.CurrentPage >= variableList.TotalPages {
			break
		}

		options.PageNumber = variableList.NextPage   // line 295: nil dereference
	}
```

Passing `nil` to `List` is fine, so the first page succeeds. The crash is
reached only when `TotalPages > 1`, i.e. when the variable set holds more than
the API's default page size of 20 variables.

`readFromWorkspace` has the identical defect at
[line 249](https://github.com/hashicorp/terraform-provider-tfe/blob/v0.79.0/internal/provider/data_source_variables.go#L249),
so `data.tfe_variables` with `workspace_id` crashes the same way for workspaces
with more than 20 variables.

**Affected versions:** v0.65.0 through v0.79.0, and `main` as of this writing.
The defect looks to have been introduced in #1679, which migrated this data source to
terraform-plugin-framework. The prior SDKv2 implementation used a value struct
(`options := tfe.VariableSetVariableListOptions{}` at v0.64.0), which made the
same assignment harmless; the migration changed it to a pointer without adding
an initializer. v0.64.0 and earlier are unaffected.

**Suggested fix** — allocate the options in both functions:

```go
options := &tfe.VariableSetVariableListOptions{}   // readFromVariableSet
options := &tfe.VariableListOptions{}              // readFromWorkspace
```

justinclayton:
Fixes #2185.

`readFromWorkspace` and `readFromVariableSet` declared their list options as nil pointers, so `options.PageNumber = variableList.NextPage` panicked as soon as a second page was needed. Allocating the structs is the whole fix; see the issue for the full trace and affected version range.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ ENHANCEMENTS:
 
 BUG FIXES:
 * `r/tfe_saml_settings`: Fix `Provider produced inconsistent result after apply` error on the sensitive `private_key` attribute when updating any other attribute without changing the private key. By @tanushreegorai [#2201](https://github.com/hashicorp/terraform-provider-tfe/pull/2201)
+* `d/tfe_variables`: Fix a nil pointer dereference panic when reading a workspace or variable set containing more than one page (20) of variables. By @justinclayton [#2186](https://github.com/hashicorp/terraform-provider-tfe/pull/2186)
 
 ## v0.80.0
 
@@ -28,7 +29,6 @@ BUG FIXES:
 * `r/tfe_workspace`: Fixed a provider bug where `hyok_enabled` was not being set as a computed read-only attribute, causing unreconcilable drift to occur when HYOK was enabled on a provider-managed workspace. By @JarrettSpiker [#2134](https://github.com/hashicorp/terraform-provider-tfe/pull/2134)
 * `r/tfe_project_notification_configuration`: Fix `Provider produced inconsistent result after apply` error on the sensitive `token` attribute when creating a configuration (such as `slack`) without a `token`. By @maed223 [#2140](https://github.com/hashicorp/terraform-provider-tfe/pull/2140)
 * `r/tfe_project_notification_configuration`: Fix `Provider produced inconsistent result after apply` errors on the `triggers`, `email_addresses`, and `email_user_ids` attributes when they are configured as an empty set (`[]`). By @maed223 [#2140](https://github.com/hashicorp/terraform-provider-tfe/pull/2140)
-* `d/tfe_variables`: Fix a nil pointer dereference panic when reading a workspace or variable set containing more than one page (20) of variables. By @justinclayton [#2186](https://github.com/hashicorp/terraform-provider-tfe/pull/2186)
 
 ## v0.79.0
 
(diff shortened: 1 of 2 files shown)


shwetamurali:
Looks great thanks for the fix! It would be great to add a regression test with a workspace/varset with >20 variables

ctrombley (dismissed):
Looks good, thanks!

christian-doucette:
the changelog entry should be moved from 0.80 to Unreleased (I'm guessing it was moved to 0.80 automatically by a rebase)

christian-doucette (dismissed):

christian-doucette (approved):

PR #2197, opened 2026-08-21
Add tag_match_logic to tfe_policy_set resource and data source

anubhav-goel:
## Description

Adds a new `tag_match_logic` attribute to `tfe_policy_set` and `tfe_policy_set` data source to control how tag-based workspace scoping is evaluated when multiple tag selectors are configured.

**Values:** `"any"` (default, OR semantics) or `"all"` (AND semantics — workspace must match every configured tag).

Applies to both tag inclusions (`tfe_tag_policy_set`) and tag exclusions (`tfe_tag_policy_set_exclusion`). Schema descriptions on both resources have been updated to reference this attribute on the parent policy set.

- [x] Updated the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)
- [x] Updated the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)

> **Note:** This PR depends on [hashicorp/go-tfe#1469](https://github.com/hashicorp/go-tfe/pull/1469) being merged and released as a new go-tfe v1 release. Local dev & testing is done using `replace` directive in `go.mod` pointing to local go-tfe instance.

## TODO

- [ ] Bump go-tfe version after new v1 version is released for go-tfe [hashicorp/go-tfe#1469](https://github.com/hashicorp/go-tfe/pull/1469)

## Testing plan

1. Test cases should pass
2. Run tf scripts locally with local atlas setup

## External links

- [Related go-tfe PR](https://github.com/hashicorp/go-tfe/pull/1436)

```
#!/bin/bash
# run-tag-selector-tests.sh
export TFE_HOSTNAME="anubhavgoel.ngrok.app"
export TFE_TOKEN="<token>"
export TFE_ADMIN_PROVISION_LICENSES_TOKEN="<token>"
export ENABLE_BETA=1
export TF_ACC=1
go clean -testcache
go test ./internal/provider/ -run TestAccTFEPolicySet_tagMatchLogic -v -count=1
go test ./internal/provider/ -run 'TestAccTFEPolicySetImport|TestAccTFEPolicySet_importByIdentity' -v -count=1
go test ./internal/provider/ -run TestAccTFEPolicySetDataSource_tagMatchLogic -v -count=1
```

```
=== RUN   TestAccTFEPolicySet_tagMatchLogicAll
2026/09/10 16:59:31 [DEBUG] Configuring client for host "anubhavgoel.ngrok.app"
2026/09/10 16:59:31 [WARN] Unable to read CLI config or credentials file /Users/anubhav.goel/.terraformrc: open /Users/anubhav.goel/.terraformrc: no such file or directory
2026/09/10 16:59:31 [DEBUG] Service discovery for anubhavgoel.ngrok.app at https://anubhavgoel.ngrok.app/.well-known/terraform.json
--- PASS: TestAccTFEPolicySet_tagMatchLogicAll (11.52s)
=== RUN   TestAccTFEPolicySet_tagMatchLogicExclusion
--- PASS: TestAccTFEPolicySet_tagMatchLogicExclusion (10.84s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider   23.484s
=== RUN   TestAccTFEPolicySetImport
2026/09/10 16:59:56 [DEBUG] Configuring client for host "anubhavgoel.ngrok.app"
2026/09/10 16:59:56 [WARN] Unable to read CLI config or credentials file /Users/anubhav.goel/.terraformrc: open /Users/anubhav.goel/.terraformrc: no such file or directory
2026/09/10 16:59:56 [DEBUG] Service discovery for anubhavgoel.ngrok.app at https://anubhavgoel.ngrok.app/.well-known/terraform.json
--- PASS: TestAccTFEPolicySetImport (4.04s)
=== RUN   TestAccTFEPolicySet_importByIdentity
--- PASS: TestAccTFEPolicySet_importByIdentity (4.39s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider   8.901s
=== RUN   TestAccTFEPolicySetDataSource_tagMatchLogic
2026/09/10 17:00:06 [DEBUG] Configuring client for host "anubhavgoel.ngrok.app"
2026/09/10 17:00:06 [WARN] Unable to read CLI config or credentials file /Users/anubhav.goel/.terraformrc: open /Users/anubhav.goel/.terraformrc: no such file or directory
2026/09/10 17:00:06 [DEBUG] Service discovery for anubhavgoel.ngrok.app at https://anubhavgoel.ngrok.app/.well-known/terraform.json
--- PASS: TestAccTFEPolicySetDataSource_tagMatchLogic (6.58s)
PASS
ok      github.com/hashicorp/terraform-provider-tfe/internal/provider   7.070s
```


internal/provider/resource_tfe_policy_set.go
shwetamurali, line 234: Could we send the configured `tag_match_logic` in `PolicySetCreateOptions` here? go-tfe v1.111.0 exposes `TagSelectorMatchingLogic` on create, but this resource currently sends it only during update.

internal/provider/resource_tfe_policy_set.go
```diff
@@ -337,6 +345,12 @@ func resourceTFEPolicySetRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("policy_update_patterns", policySet.PolicyUpdatePatterns)
 	d.Set("agent_enabled", policySet.AgentEnabled)
 
+	if policySet.TagSelectorMatchingLogic != nil {
+		d.Set("tag_match_logic", *policySet.TagSelectorMatchingLogic)
+	} else {
+		d.Set("tag_match_logic", "")
```
shwetamurali: Should we preserve the configured value when the API returns nil instead of writing "" into state? The schema permits only "any" and "all", and this attribute is Optional plus Computed. The API can temporarily omit the value before tag scoping is visible, so replacing configured "all" with "" creates the documented second-apply drift. For the resource, skipping `d.Set` when the response is nil should retain the configured value. Please add coverage showing that the resource converges after one apply.

internal/provider/resource_tfe_policy_set_test.go
```diff
@@ -1475,3 +1477,265 @@ resource "tfe_policy_set" "foobar" {
...
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: testAccMuxedProviders,
+		CheckDestroy:             testAccCheckTFEPolicySetDestroy,
+		Steps: []resource.TestStep{
+			{
+				// First apply: workspace + policy set + tag created together.
+				// Server cannot return tag_match_logic yet (tag just attached),
+				// so state has tag_match_logic = "" while config has "all" — drift is expected.
+				Config:             testAccTFEPolicySet_tagMatchLogicWithTagSelectorAll(org.Name, rInt),
+				ExpectNonEmptyPlan: true,
```
shwetamurali, line 1507: This appears to encode provider drift as expected behavior. An explicit tag_match_logic = "all" should converge after the first apply rather than require a second apply. Once the create and read paths are corrected, could this test verify an empty follow-up plan instead? It would also be valuable to use two selectors, since "any" and "all" are behaviorally identical with only one selector.

internal/provider/resource_tfe_policy_set.go
anubhav-goel: On tag_match_logic and drift:

The flow here has a chicken and egg problem, and it's why drift is expected and why we couldn't do this in one apply.

1. tfe_policy_set is created first. At that point there are no tag selectors attached, so the backend has no basis to interpret tag_match_logic. It's ignored at create time.
2. tfe_tag_policy_set & tfe_tag_policy_set_exclusion (the tag scope) is created after. At this point the backend sees that tag_match_logic is null and assigns it the default value of "any".
3. So even though the config says tag_match_logic = "all", the backend only has "any" at this stage. The state drifts from the config, and a second terraform apply is required for the "all" to actually persist.
4. We can't avoid this by bundling the policy set and its tag selectors in a single apply, because there's an ordering constraint baked into the existing pattern: create the policy set first, then add its scopes (projects, workspaces, etc.). That's the established resource flow here. So the policy set + its tag selectors + tag matching logic can never be created in the same apply.

Alternative we considered: we could restructure everything and use a single go-tfe SDK method (a single policy set create / policy set update) instead of the independent scoping methods we use today (add project inclusion/exclusion, add workspace inclusion/exclusion, add tag inclusion/exclusion). That would let the policy set and its full scoping be created atomically in one call, eliminating the drift entirely.

But that's a large shift from existing behaviour, it would change how the resource is built, how inclusion/exclusion maps to state, and potentially change the state itself. It's a much bigger, riskier change than the scope of this PR, so I've kept the current per-scope flow and documented the expected drift/second apply instead.

internal/provider/resource_tfe_policy_set.go
anubhav-goel: Refactored it, instead of setting empty string, setting whatever is received from the backend.

internal/provider/resource_tfe_policy_set.go
shwetamurali: Ok that makes sense about how there ordered, but the configured value is still not included in `PolicySetCreateOptions`, even though go-tfe supports it. Is there a test or API result showing that `tag-selector-matching-logic = "all"` is ignored when the policy set is created without selectors? If is ignored and a second apply can't be avoided then I think we should have explicit agreement that this behavior is acceptable because the policy set can temporarily use "any" when the configuration requested "all".

internal/provider/resource_tfe_policy_set.go
shwetamurali: I don’t think this preserves the configured value when the API returns nil. The SDK treats a nil pointer passed to `d.Set` as nil and clears the value from state. Can we keep the nil check and only call `d.Set` when the API returns a value?

internal/provider/resource_tfe_policy_set_test.go
shwetamurali: Thanks for adding the second selector. This still expects drift after the first apply, though. Can we check the policy set through the API after the selectors are created to confirm whether the backend returns nil or "any"? That would make it clear why the second apply is needed.

internal/provider/resource_tfe_policy_set_test.go
```diff
@@ -1475,3 +1477,285 @@ resource "tfe_policy_set" "foobar" {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("tfe_policy_set.test", "tag_match_logic", "any"),
+				),
+				ExpectNonEmptyPlan: false,
+			},
+			{
+				ResourceName:            "tfe_policy_set.test",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"tag_match_logic", "overridable"},
```
shwetamurali: Do we need to ignore `tag_match_logic` during import verification? Once the selectors exist, I’d expect it to be read back correctly. Removing this ignore would give us coverage for importing the new attribute.

internal/provider/resource_tfe_policy_set_test.go
```diff
@@ -1475,3 +1477,285 @@ resource "tfe_policy_set" "foobar" {
...
+	})
+	t.Cleanup(orgCleanup)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: testAccMuxedProviders,
+		CheckDestroy:             testAccCheckTFEPolicySetDestroy,
+		Steps: []resource.TestStep{
+			{
+				// First apply: global policy set + workspace + exclusion tag created together.
```
shwetamurali, line 1735: Can we use two exclusion selectors here too? With one selector, "any" and "all" behave the same.

internal/provider/resource_tfe_policy_set.go
anubhav-goel: Yes, here is an example of the API call made when a tfe_policy_set is created with tag_matching_logic = any/all. At that point, there are no tag selectors attached to the policy set, so the backend has no basis on which to interpret tag_match_logic. As a result, tag_match_logic is ignored at creation time.

Screenshot 2026-09-04 at 12 27 30 PM

Hence, I have not added tag_match_logic to PolicySetCreateOptions, since adding it to terraform-provider-tfe would effectively result in a no-op.

internal/provider/resource_tfe_policy_set.go
anubhav-goel: The unconditional d.Set is intentional. nil from the backend is a meaningful signal, not noise to suppress. The backend sets tag_selector_matching_logic to nil when switching from tag scope to workspace scope. If we skip d.Set on nil, stale state (e.g. "any") would be preserved silently. We want it to reset back to nil (empty string)

TestAccTFEPolicySet_tagMatchLogicAll demonstrates this directly: step 5 removes all tag selectors, step 6 switches the policy set to workspace scope, the backend resets tag_match_logic to nil and the test asserts tag_match_logic = "" in state. The same flow is covered for the exclusion case in TestAccTFEPolicySet_tagMatchLogicExclusion.

internal/provider/resource_tfe_policy_set_test.go
anubhav-goel: Yes, sure. Here is the flow through the APIs used by the provider:
First apply: Creates the policy set, followed by creating the tag selectors associated with that policy set.
Second apply: Updates tag_matching_logic to all.

https://github.com/user-attachments/assets/63e1a83b-49c4-4419-9f37-8718ec61ac6a

This is why tag_matching_logic is handled during the update rather than during policy set creation.

internal/provider/resource_tfe_policy_set_test.go
anubhav-goel: Sure, removed, thanks!!

internal/provider/resource_tfe_policy_set_test.go
anubhav-goel: Sure, added, thanks!!

brandonc (changes requested):
We can't accept adding a resource attribute if that attribute cannot be created and requires two applies to take effect. That's not a behavior we can just document. Can you change the backend, provider resources, attribute location, or API call pattern to make this work as anyone would reasonably expect?

anubhav-goel:
> We can't accept adding a resource attribute if that attribute cannot be created and requires two applies to take effect. That's not a behavior we can just document. Can you change the backend, provider resources, attribute location, or API call pattern to make this work as anyone would reasonably expect?

Sure, have made the necessary changes in atlas & go-tfe @brandonc @shwetamurali

go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/hashicorp/go-retryablehttp v0.7.8 // indirect
 	github.com/hashicorp/go-slug v1.0.0
-	github.com/hashicorp/go-tfe v1.111.0
+	github.com/hashicorp/go-tfe v1.111.1-0.20260911160946-454b354429be
```
brandonc: I think this was released as `v2.11.0`

mattias-fjellstrom:
The current implementation is now broken after the tag-based policy set went GA (reported in #2223). This PR will likely fix that issue.

brandonc (dismissed):

go.mod
anubhav-goel: I thought the v2 version resolves to the `v2/` subdirectory, v1.go is outside of that folder. Can we use v1 code from v2? In go-tfe changes are made to v1.go

brandonc: Yes, the v2 subdirectory refers to the v2 module, but you can reference v1 module code from a v2.11.0 release.

brandonc: Just `go get -u github.com/hashicorp/go-tfe` and `mod tidy`

brandonc: You are right. I just backported your changes to 1.x and released as v1.111.1.

anubhav-goel: Thank you!! I'll need a re-approval for merging

brandonc (approved):

shwetamurali (approved):

PR #2198, opened 2026-08-22
fix: keep listing tfe_projects after a full 100-item page

### LINKED ISSUE #2196: Datasource tfe_projects does not retrieve all projects from organization

Experienced in HCP Terraform

#### Terraform version

```plaintext
1.15.9
```

```terraform

resource tfe_project one_oh_five {
    count = 105
    organization = "myorg"
    name = "my project ${count.index}"
}

data tfe_projects all_projects {
    organization = "myorg"
    depends_on = [ tfe_project.one_oh_five ]
}

output all_projects_count {
    value = "Found ${length(data.tfe_projects.all_projects.projects)} projects, expecting 105."
}


```

Output

```plaintext

Apply complete! Resources: 105 added, 0 changed, 0 destroyed.

Outputs:

all_projects_count = "Found 100 projects, expecting 105."

```

#### Expected Behavior

Provider returns all projects in organization

#### Actual Behavior

Provider returns only the first 100 projects organization

#### Additional Context

If the organization contains more than 100 projects, it seems that only the first 100 are returned. The datasource does not provide any means of pagination.

By way of illustration, in the sample code above, I have created 105 projects, but the datasource only returns 100.

Ideally the datasource needs to be updated to expose attributes for the total number of projects in the organization, and some means of providing a pagination selector or other filter.

AshSgDe29071999:
## Problem

`data.tfe_projects` returned only the first 100 projects. The list response types `meta` as `Paginationable`, but HCP Terraform nests pagination under `meta.pagination`, so `GetNextPage()` was always nil and the loop stopped after page 1.

Fixes #2196.

## Change

If `next-page` is missing, keep fetching while the current page was full (`advanceListPage`).

## Test

```
go test ./internal/provider -count=1 -run TestAdvanceListPage
# ok
```

### DIFF
--- a/internal/provider/client_v2_helpers.go
+++ b/internal/provider/client_v2_helpers.go
@@ -64,6 +64,20 @@ func nextPageNumber(p models.Paginationable) *int32 {
 	return p.GetNextPage()
 }
 
+// advanceListPage returns the next page to request. Prefer the API's
+// next-page pointer. When that pointer is missing — the projects list
+// types `meta` as Paginationable while HCP Terraform nests pagination
+// under `meta.pagination` — keep walking while the current page was full.
+func advanceListPage(current, pageSize int32, itemCount int, next *int32) (int32, bool) {
+	if next != nil {
+		return *next, true
+	}
+	if int32(itemCount) >= pageSize {
+		return current + 1, true
+	}
+	return current, false
+}
+
 // paginationMetaGetter is satisfied by every go-tfe v2 generated list
 // response's `meta` type (e.g. ItemSshKeysGetResponse_meta,
 // ItemTeamsGetResponse_meta): each is a distinct generated type, but all of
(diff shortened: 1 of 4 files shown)


shwetamurali:
Thank you for the fix.

Could we add a regression test using a realistic projects response with nested `meta.pagination` and verify that the second page is requested and included? The current helper test assumes `next == nil`, so it does not exercise the actual deserialization/request path that caused the bug.

shwetamurali:
Thank you again for investigating this and identifying the response-model mismatch. Your analysis pointed us to the root cause: the generated client expects pagination fields directly under meta, while HCP Terraform returns them under meta.pagination.

I’m going to close this PR in favor of another fix. Rather than infer another page whenever the current page contains 100 items, the new implementation reads the actual nested `meta.pagination.next-page` value preserved by Kiota’s additional data. This avoids an unnecessary request when the project count is an exact multiple of 100 and lets us cover the original failure with an HTTP-backed regression test that verifies the second page is requested and included.

I’ll make sure the replacement PR acknowledges your investigation and links back to this contribution. Thank you for taking the time to diagnose and propose a fix.

PR #2201, opened 2026-08-28
Add Site Auditor SAML attributes and migrate tfe_saml_settings to go-tfe/v2

tanushreegorai:
## Description

Adds `attr_site_auditor` and `site_auditor_role` to `tfe_saml_settings` (resource and data source), mirroring the existing `attr_site_admin` / `site_admin_role` pair, so the Site Auditor role can be managed as code instead of only through the admin UI.

Both components also move from the frozen go-tfe v1 client to go-tfe/v2 v2.8.0 — v1 has no Site Auditor fields. The endpoints were added to the Atlas OpenAPI spec in hashicorp/atlas#30874 and reached the SDK in v2.8.0.

**Version gating.** The attributes require TFE 2.1.0. Older releases ignore unknown attributes and omit them from responses, which would leave the plan holding `"SiteAuditor"` while state collapsed to `""`. So the version check errors only when the attributes are actually set in config (a schema `Default` leaves the config value null, so existing configs keep working), the request omits them when unsupported, and the mapper falls back to the prior value or the default.

The v2 setter for `sso_api_token_session_timeout` takes `int32` where v1 took a 64-bit `int`, so the value is clamped at the conversion rather than being allowed to wrap. No validation behaviour changes for existing configurations.

**One pre-existing bug is fixed**, because the new subtest makes it reachable: `Update` wrote `private_key = null` to state whenever the key was unchanged — the guard `len(privateKey.String()) > 0` never filtered nulls, since `types.String` renders null as `"<null>"`. Any update to an unrelated attribute failed with `.private_key: inconsistent values for sensitive attribute`. This is deliberately the only out-of-scope change: without it the new Site Auditor subtest fails on every TFE version.

## Testing plan

1.  `go test ./internal/provider/ -run 'TestSAMLSettings'` — three unit tests, no TFE needed, so they run in PR CI. `TestSAMLSettingsPrivateKeyNullNeverReachesState` was checked to **fail** with the old guard restored.
1.  Against TFE 2.1.0: `TESTARGS="-run TestAccTFESAMLSettings" make testacc`. The new subtest sets both attributes, reads them back via the data source, updates one, then drops both from config and confirms they return to the defaults.
1.  Against a pre-2.1.0 TFE: setting the attributes fails with `requires Terraform Enterprise 2.1.0 or later`; omitting them applies unchanged (the upgrade path for existing users).
1.  private_key fix: apply without changing `private_key`, then change `debug` and apply again. Before this PR the second apply failed.

## External links

- [Atlas PR adding the endpoints to the OpenAPI spec](https://github.com/hashicorp/atlas/pull/30874)
- [go-tfe PR generating the client](https://github.com/hashicorp/go-tfe/pull/1451) — released in v2.8.0
- [API documentation](https://developer.hashicorp.com/terraform/enterprise/api-docs/admin/settings)

```
$ go test ./internal/provider/ -run 'TestSAMLSettings' -v
--- PASS: TestSAMLSettingsPrivateKeyStateConsistency
--- PASS: TestSAMLSettingsPrivateKeyNullNeverReachesState
--- PASS: TestSAMLSettingsSiteAuditorFallback
ok  	github.com/hashicorp/terraform-provider-tfe/internal/provider
```

Adds configuration-as-code management of the Site Auditor role mapping, an access control. It exposes an existing TFE admin API rather than changing authorization: the server still enforces admin auth, and TFE grants the role at SAML login. `private_key` stays `Sensitive` and write-only, and a fix here stops a null being written to state for it.

### DIFF
--- a/internal/provider/resource_tfe_saml_settings.go
+++ b/internal/provider/resource_tfe_saml_settings.go
@@ -6,10 +6,12 @@ package provider
 import (
 	"context"
 	"fmt"
+	"math"
 
-	tfe "github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/go-tfe/v2/api/models"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -18,7 +20,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
-	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
@@ -27,9 +28,24 @@ const (
 	samlSignatureMethodSHA256                   string = "SHA256"
 	samlDefaultAttrUsername                     string = "Username"
 	samlDefaultAttrSiteAdmin                    string = "SiteAdmin"
+	samlDefaultAttrSiteAuditor                  string = "SiteAuditor"
 	samlDefaultAttrGroups                       string = "MemberOf"
 	samlDefaultSiteAdminRole                    string = "site-admins"
+	samlDefaultSiteAuditorRole                  string = "site-auditors"
 	samlDefaultSSOAPITokenSessionTimeoutSeconds int64  = 1209600 // 14 days
+
+	samlProviderTypeOkta    string = "okta"
+	samlProviderTypeEntra   string = "entra"
+	samlProviderTypeGeneric string = "saml"
+	samlProviderTypeUnknown string = "unknown"
+
+	// samlSettingsID is the ID of the SAML settings singleton resource.
+	samlSettingsID string = "saml"
+
+	// minTFEVersionSiteAuditor is the first Terraform Enterprise release that
+	// supports provisioning the Site Auditor role through SAML and SCIM. Shared
+	// with the tfe_scim_settings resource.
+	minTFEVersionSiteAuditor string = "2.1.0"
 )
 
 type modelTFESAMLSettings struct {
@@ -47,6 +63,8 @@ type modelTFESAMLSettings struct {
 	AttrGroups                types.String `tfsdk:"attr_groups"`
 	AttrSiteAdmin             types.String `tfsdk:"attr_site_admin"`
 	SiteAdminRole             types.String `tfsdk:"site_admin_role"`
+	AttrSiteAuditor           types.String `tfsdk:"attr_site_auditor"`
+	SiteAuditorRole           types.String `tfsdk:"site_auditor_role"`
 	SSOAPITokenSessionTimeout types.Int64  `tfsdk:"sso_api_token_session_timeout"`
 	ACSConsumerURL            types.String `tfsdk:"acs_consumer_url"`
 	MetadataURL               types.String `tfsdk:"metadata_url"`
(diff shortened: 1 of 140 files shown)


internal/provider/resource_tfe_saml_settings.go
```diff
@@ -61,38 +79,94 @@ type modelTFESAMLSettings struct {
 
 // resourceTFESAMLSettings implements the tfe_saml_settings resource type
 type resourceTFESAMLSettings struct {
-	client *tfe.Client
+	config ConfiguredClient
```
ajmera-naman, line 84: :❓ Type is Client, variable name is config. Same at [line](https://github.com/hashicorp/terraform-provider-tfe/pull/2201/changes#diff-c3ba5eac0ff8e4e90e4f7139d90dfcac97e9a2da2f399a1d45b13d8ac88ec62cR226). can we use client here instead

internal/provider/resource_tfe_saml_settings.go
```diff
@@ -280,17 +397,17 @@ func (r *resourceTFESAMLSettings) Schema(ctx context.Context, req resource.Schem
...
 				Validators: []validator.String{
 					stringvalidator.OneOf(
-						string(tfe.SAMLProviderTypeOkta),
-						string(tfe.SAMLProviderTypeEntra),
-						// `tfe.SAMLProviderTypeGeneric` is the string literal "saml", and is shown as `SAML` in the TFE UI.
-						string(tfe.SAMLProviderTypeGeneric),
-						string(tfe.SAMLProviderTypeUnknown),
+						samlProviderTypeOkta,
+						samlProviderTypeEntra,
+						// `samlProviderTypeGeneric` is the string literal "saml", and is shown as `SAML` in the TFE UI.
```
ajmera-naman: :🎨 I think we can remove this comment now, as it is defined in the same file.

tanushreegorai: Good point, removed it. I've kept the "the TFE UI shows this as SAML" part and moved it up to the constant definition, since that isn't derivable from the code and explains why the constant name and its value differ. Happy to drop it entirely if you'd prefer.

internal/provider/resource_tfe_saml_settings.go
tanushreegorai: Thanks for flagging this. I'd prefer to keep config, as it matches the existing convention in this package: all 63 struct fields typed `ConfiguredClient` are named config, while the 6 named client are all `*tfe.Client`. This resource has moved from the second category to the first, since it now needs `ClientV2` and `MeetsMinRemoteTFEVersion`.

It's also not strictly a client, `ConfiguredClient` wraps Client, ClientV2 and Organization — so r.config.ClientV2... reads more accurately than r.client.ClientV2... would.

On the second line you linked: you're right that `client, ok := ...` followed by `r.config = client` is inconsistent. But that's the same code every `Configure` function uses, it's in 71 other files and this PR didn't change it, so fixing it only here would make this file the odd one out. I'm happy to work on this if you'd like it addressed across the repo.

internal/provider/resource_tfe_saml_settings.go
```diff
@@ -484,28 +652,36 @@ func (r *resourceTFESAMLSettings) determinePrivateKeyForUpdate(plan, state, conf
...
+	attrs.SetIdpCert(m.IDPCert.ValueStringPointer())
+	attrs.SetCertificate(m.Certificate.ValueStringPointer())
+	attrs.SetPrivateKey(m.PrivateKey.ValueStringPointer())
+	attrs.SetSloEndpointUrl(m.SLOEndpointURL.ValueStringPointer())
+	attrs.SetSsoEndpointUrl(m.SSOEndpointURL.ValueStringPointer())
+	attrs.SetAttrUsername(m.AttrUsername.ValueStringPointer())
+	attrs.SetAttrGroups(m.AttrGroups.ValueStringPointer())
+	attrs.SetAttrSiteAdmin(m.AttrSiteAdmin.ValueStringPointer())
+	attrs.SetSiteAdminRole(m.SiteAdminRole.ValueStringPointer())
+	attrs.SetSsoApiTokenSessionTimeout(ptr(int32SessionTimeout(m.SSOAPITokenSessionTimeout.ValueInt64())))
```
shwetamurali: Could we reject values outside the supported range instead of silently clamping them? The schema currently accepts any int64, so a configured value can differ from the value sent and returned by TFE. For example, -1 is sent as 0, Atlas only validates this field for presence, so it can accept and return 0, causing Terraform to report an inconsistent result after apply because the plan still contains -1.

To preserve the previous behavior as closely as possible, I think we should validate that the value fits within int32 and then cast directly. If negative timeouts should now be prohibited, validating `0..math.MaxInt32` would also work, but that would be an intentional validation change.

tanushreegorai: Thanks for pointing that out, and I've confirmed it end to end. Atlas only has validates `:sso_api_token_session_timeout`, `presence: true`, and `0.present?` is true in Ruby, so a clamped -1 is accepted and echoed back as 0, breaking the apply exactly as you describe. It's also a regression: v1 took `*int `and sent the value through unchanged.

One wrinkle on the suggested fix, I tried the validator-plus-direct-cast version and gosec still fails CI:

`resource_tfe_saml_settings.go:675:46: G115: integer overflow conversion int64 -> int32 (gosec)`

G115 can't see a runtime validator, which is why the clamp was there. So I've done both: `int64validator.Between(math.MinInt32, math.MaxInt32)` on the schema for a plan-time error, and `int32SessionTimeout` now returns an error instead of clamping, which is what makes the conversion provably in range for gosec. make lint is back to 0 issues.

I've gone with `MinInt32..MaxInt32` rather than `0..` so this stays a bug fix and not a validation change - happy to tighten it to reject negatives if you'd prefer.

shwetamurali (approved):

PR #2209, opened 2026-09-03
Add Site Auditor SCIM group mapping and migrate tfe_scim_settings to go-tfe/v2

tanushreegorai:
## Description

Migrates `tfe_scim_settings` (resource and data source) from the frozen go-tfe v1 client to go-tfe/v2 (v2.8.0), and adds the Site Auditor group mapping shipped in Terraform Enterprise 2.1.0:

- `site_auditor_group_scim_id` — Optional + Computed, default `""`. Links a SCIM group to the Site Auditor role; `""` or omitting it unlinks.
- `site_auditor_group_display_name` — Computed. The linked group's display name.

Both mirror the existing `site_admin_group_*` pair exactly, per the RFC review decision. This is the SCIM half of the Site Auditor provider work; the SAML half was #2201.

Two things worth knowing while reviewing:

- **The empty string is what unlinks, not null.** The generated client omits a JSON key when its pointer is nil, and TFE reads an absent key as "leave this link unchanged". So `groupSCIMIDForRequest` maps both null and empty to `""` and never sends a nil pointer.
- **These attributes use `valueOrZero`, not the `stringOrPrior` helper #2201 added for SAML.** An unlinked group serializes as JSON `null`, indistinguishable from a pre-2.1.0 release omitting the attribute — falling back to the prior value would hide a genuine out-of-band unlink.

Gating reuses the `minTFEVersionSiteAuditor` constant from #2201 and errors only when the attribute is set in config, so existing configurations keep working on older releases. Also corrects the resource description, which said `paused` and `site_admin_group_scim_id` were "the only mutable arguments" — true until this change added a third.

- [x] Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)
- [x] Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation) — `make generate` run; doc content is a separate follow-up story

## Testing plan

Requires a licensed TFE running 2.1.0 or later, with SAML enabled and a recognized provider type.

1. Apply `tfe_scim_settings` with no group linked; confirm both auditor attributes are `""`.
2. Create a SCIM group out-of-band, set `site_auditor_group_scim_id` to its ID, and confirm the display name populates and the data source agrees.
3. Re-apply the same config; confirm an empty plan.
4. Confirm members of the linked group gain `is_site_auditor` via the admin users API.
5. Set the attribute to `""`, then re-link and remove it entirely — confirm both unlink and revoke the role.
6. Against a pre-2.1.0 instance, confirm setting the attribute fails with `requires Terraform Enterprise 2.1.0 or later`, and that a config without it still applies cleanly.

## External links

- [SCIM configuration docs](https://developer.hashicorp.com/terraform/enterprise/saml/login#scim)
- [SAML half of this work](https://github.com/hashicorp/terraform-provider-tfe/pull/2201)
- [JIRA](https://hashicorp.atlassian.net/browse/TF-40128)

The acceptance subtests for this resource are excluded from PR CI and run in the nightly TFE job (`ci.yml` skips `TestAccTFESCIM`), and need a licensed 2.1.0 instance. **They have not been run yet** — the live verification above is still outstanding.

Green locally: the four new unit tests (no `TF_ACC` needed), plus `go build`, `go vet`, `gofmt`, `make lint` (0 issues), `make exmplcheck`, `make generate`.

## Rollback Plan

If this change needs to be reverted, we will roll out an update to the code within 7 days. The change is additive, so reverting removes the two attributes and returns the resource to the v1 client. Any Site Auditor group already linked would need to be unlinked through the TFE admin UI, since a provider rollback does not undo the server-side link.

This adds a Terraform-managed path for granting the Site Auditor role (a read-only administrative role) to members of a SCIM group. It is the counterpart of the existing `site_admin_group_scim_id` control and uses the same server-side mechanism; the provider only sets the group link, and all grant/revoke logic stays in Terraform Enterprise. Requires an admin token, as this resource already did. No changes to encryption or logging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@ ENHANCEMENTS:
 * `r/tfe_policy_set`: Add `tfpolicy` as a valid value for the `kind` attribute. **NOTE:** This policy kind is currently in beta and not yet available to all users. By @subhro-acharjee-ibm [#2109](https://github.com/hashicorp/terraform-provider-tfe/pull/2109)
 * **New Resource List:** `tfe_provider_set` By @kierramarie [#2171](https://github.com/hashicorp/terraform-provider-tfe/pull/2171)
 * `r/tfe_hyok_configuration`: Added multi-region key support for AWS HYOK. By @helenjw [#2187](https://github.com/hashicorp/terraform-provider-tfe/pull/2187)
-* `r/tfe_saml_settings`, `d/tfe_saml_settings`: Add `attr_site_auditor` and `site_auditor_role` attributes for provisioning the Site Auditor role through SAML. Requires Terraform Enterprise v2.1.0 or later; on earlier releases the attributes are ignored unless set explicitly, in which case a minimum-version error is returned. By @tanushreegorai
-* `r/tfe_scim_settings`, `d/tfe_scim_settings`: Add `site_auditor_group_scim_id` and `site_auditor_group_display_name` attributes for granting the Site Auditor role to the members of a SCIM group, mirroring the existing site admin group mapping. Setting `site_auditor_group_scim_id` to `""` (or omitting it) unlinks the group. Requires Terraform Enterprise v2.1.0 or later; on earlier releases a minimum-version error is returned when `site_auditor_group_scim_id` is set. By @tanushreegorai
+* `r/tfe_saml_settings`, `d/tfe_saml_settings`: Add `attr_site_auditor` and `site_auditor_role` attributes for provisioning the Site Auditor role through SAML. Requires Terraform Enterprise v2.1.0 or later; on earlier releases the attributes are ignored unless set explicitly, in which case a minimum-version error is returned. By @tanushreegorai [#2201](https://github.com/hashicorp/terraform-provider-tfe/pull/2201)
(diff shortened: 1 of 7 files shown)


kkavish (dismissed):

vmacdougal-hc (dismissed):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ ENHANCEMENTS:
 * **New Resource List:** `tfe_provider_set` By @kierramarie [#2171](https://github.com/hashicorp/terraform-provider-tfe/pull/2171)
 * `r/tfe_hyok_configuration`: Added multi-region key support for AWS HYOK. By @helenjw [#2187](https://github.com/hashicorp/terraform-provider-tfe/pull/2187)
 * `r/tfe_saml_settings`, `d/tfe_saml_settings`: Add `attr_site_auditor` and `site_auditor_role` attributes for provisioning the Site Auditor role through SAML. Requires Terraform Enterprise v2.1.0 or later; on earlier releases the attributes are ignored unless set explicitly, in which case a minimum-version error is returned. By @tanushreegorai
+* `r/tfe_scim_settings`, `d/tfe_scim_settings`: Add `site_auditor_group_scim_id` and `site_auditor_group_display_name` attributes for granting the Site Auditor role to the members of a SCIM group, mirroring the existing site admin group mapping. Setting `site_auditor_group_scim_id` to `""` (or omitting it) unlinks the group. Requires Terraform Enterprise v2.1.0 or later; on earlier releases a minimum-version error is returned when `site_auditor_group_scim_id` is set. By @tanushreegorai
```
christian-doucette: very minor, but PR is usually linked in the changelog

christian-doucette (dismissed):
LGTM, just left one very minor comment

christian-doucette (approved):

PR #2213, opened 2026-09-07
fix `idp_cert` plan drift from TFE's PEM re-wrapping

skj-skj:
### What
`tfe_saml_settings`'s `idp_cert` now compares certificates semantically (via `StringSemanticEquals`) instead of doing a plain string comparison, so PEM formatting differences no longer cause apply errors or a plan that never converges.

### Why
TFE re-wraps `idp_cert` to 64-char lines on write, so a byte-for-byte comparison sees it as a different cert. That breaks the first apply, and later causes every plan to show drift with `Fingerprint is already trusted` on apply.

Tried a `PlanModifier` first ([TF-40682](https://hashicorp.atlassian.net/browse/TF-40682)), but it just pinned state to the old formatting, config could never match state again, so the plan never went clean. Switched to `StringSemanticEquals`, which lets us say two values are the same cert even if they're formatted differently.

### How
- Added `internal/provider/customtypes/pem_certificate.go`: `PEMCertificateType`/`PEMCertificateValue` implementing `basetypes.StringValuableWithSemanticEquals`. Certificates are compared by stripping PEM armor and whitespace, not by raw bytes.
- `idp_cert` in `resource_tfe_saml_settings.go` now uses `CustomType: customtypes.PEMCertificateType{}`.

## Testing plan

0. Spin up a nightly TFE build (v2.1.0+) (multi IdP cert support).
1. Configure `tfe_saml_settings` with an `idp_cert` value and apply, confirm no `Provider produced inconsistent result after apply` error.
2. Reformat the `idp_cert` value in config (e.g. change line wrapping or add/remove PEM armor) without changing the actual certificate, then run `terraform plan` and `terraform apply`, confirm it applies cleanly and a subsequent `plan` shows no diff.
3. Run `terraform import` on an existing `tfe_saml_settings` resource, then `plan`/`apply`, confirm no spurious `idp_cert` update or `Fingerprint is already trusted` error.
4. Run the updated unit and acceptance tests for `tfe_saml_settings` to confirm existing behavior is preserved.

## External links

- [JIRA](https://hashicorp.atlassian.net/browse/TF-40682)

```
$ TESTARGS="-run TestAccTFESAML" make testacc
```

image

## Rollback Plan

we can revert this pr

NA

[TF-40682]: https://hashicorp.atlassian.net/browse/TF-40682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/provider/customtypes/pem_certificate.go
nikolasrieble, line 1: ❓ I am curious to understand - why do we introduce this custom type? Is this not effectively a single unit testable parsing function after which can then resort to string comparison?

nikolasrieble: This might be an established pattern, I have not previously reviewed providers 😅

internal/provider/helper_test.go
nikolasrieble, line 4: ❓ If these helpers are solely for tests, should they possibly be in the provider_test package instead?

internal/provider/resource_tfe_saml_settings_test.go
```diff
@@ -38,7 +41,7 @@ const testResourceName = "tfe_saml_settings.foobar"
 // Should this test name ever change, you will also need to update the regex in ci.yml
 func TestAccTFESAMLSettings_writeOnly(t *testing.T) {
 	s := tfe.AdminSAMLSetting{
-		IDPCert:        "testIDPCertBasic",
+		IDPCert:        testIDPCertBody(t),
```
nikolasrieble, line 43: ❓ Does every test need a separate certificate? Or should we create a single test certificate and reuse it?

internal/provider/resource_tfe_saml_settings_test.go
```diff
@@ -378,6 +386,41 @@ func TestAccTFESAMLSettings_omnibus(t *testing.T) {
...
+				},
+				{
+					// Same cert without the armor. Semantic equality means the
+					// reformatted config value lands in state, so the step's
+					// implicit follow-up plan is empty: it converges.
+					Config: testAccTFESAMLSettings_idpCert(body),
+					Check:  resource.TestCheckResourceAttr(testResourceName, "idp_cert", body),
+				},
+				{
+					// Re-applying the same config is a no-op.
```
nikolasrieble: ❓ Does this third step add value to the test? Does the test otherwise pass previously? 

If I understand correctly, this third step can be removed. 
🚧 If that is the case, it should be removed.

internal/provider/resource_tfe_saml_settings_test.go
```diff
@@ -400,9 +443,10 @@ func testAccTFESAMLSettingsDestroy(_ *terraform.State) error {
 	if s.TeamManagementEnabled {
 		return errors.New("SAML settings TeamManagementEnabled is set to true")
 	}
-	if s.IDPCert != "" {
-		return fmt.Errorf("SAML settings IDPCert is not empty: `%s`", s.IDPCert)
-	}
+	// IDPCert is not checked: disabling SAML keeps the last valid certificate.
```
nikolasrieble, line 435: ❓ Do I understand correctly, that this effectively is an unrelated change? 

It seems to have nothing to do with the fix of the PR summary. What am I missing?

internal/provider/customtypes/pem_certificate.go
```diff
@@ -0,0 +1,144 @@
...
+		)
+		return false, diags
+	}
+
+	return samePEMBody(v.ValueString(), other.ValueString()), diags
+}
+
+// Matches any PEM armor line, so legacy labels like X509 CERTIFICATE are
+// stripped too.
+var pemArmor = regexp.MustCompile(`-{5}(BEGIN|END) [A-Z0-9 ]+-{5}`)
```
nikolasrieble, line 130: ❓ I wonder if this is too lax - do we really mean to define two content-identical certs with arbitrarily different armor as identical?

nikolasrieble (commented):
Thank you for the fix, this seems like a problem to solve for us. 
I have a few questions with regards to the implementation.

internal/provider/resource_tfe_saml_settings_test.go
skj-skj: Yep, the testing frameworks run an implicit plan after apply and by default `ExpectNonEmptyPlan` is false, which will fail if after apply the plan is not empty, so the 3rd step is not needed for no-op check. 🗑️ removed

internal/provider/resource_tfe_saml_settings_test.go
skj-skj: In TFE 2.1.0+ the `idp_cert` will be managed independently. They at the moment can't be removed by simply destroying the SAML settings resource in TFE 2.1.0+, so I removed the check to pass the test and added a comment here. to maintain the backward compatability I will raise another PR to fix this.

internal/provider/resource_tfe_saml_settings_test.go
skj-skj: Each test needs a valid cert. Reusing a single cert across tests doesn't provide any real benefit, and generating a new cert only takes a few ms, so the overhead is negligible.

vmacdougal-hc (dismissed):

internal/provider/helper_test.go
skj-skj: [provider_test.go](https://github.com/hashicorp/terraform-provider-tfe/blob/main/internal/provider/provider_test.go) is more of a test scaffolding (muxed providers, configure tests), and [helper_test.go](https://github.com/hashicorp/terraform-provider-tfe/blob/main/internal/provider/helper_test.go) are for shared function used in test.

internal/provider/customtypes/pem_certificate.go
skj-skj: a simple parsing+comparison func has no way to be plugged into tf `plan`/`apply` flow

`StringSemanticEquals` is the hook the framework looks for, and invokes automatically in Create/Update/Read when comparing what the server returns against what is already in state/plan.

We need it because TFE from 2.1.0+ will normalize `idp_cert` on write, and just string comparison sees that as a real change and will error out with `Provider produced inconsistent result after apply`. With semantic equality, when it is the same cert just formatted differently, the framework keeps the plan value in state instead of the server's reformatted one, so it won't drift.

internal/provider/customtypes/pem_certificate.go
skj-skj: Yep, regex is to match any `BEGIN` and `END` for legacy armour like `-----BEGIN/END TRUSTED CERTIFICATE-----` and `-----BEGIN/END X509 CERTIFICATE-----`.

kkavish (approved):

internal/provider/resource_tfe_saml_settings_test.go
```diff
@@ -474,10 +507,10 @@ resource "tfe_saml_settings" "foobar" {
 }`, s.IDPCert, s.SLOEndpointURL, s.SSOEndpointURL, s.Debug, s.AuthnRequestsSigned, s.WantAssertionsSigned, s.TeamManagementEnabled, s.AttrUsername, s.AttrSiteAdmin, s.AttrGroups, s.SiteAdminRole, s.SSOAPITokenSessionTimeout, s.Certificate, s.PrivateKey, s.SignatureSigningMethod, s.SignatureDigestMethod, s.ProviderType)
 }
 
-func testAccTFESAMLSettings_siteAuditor(attrSiteAuditor, siteAuditorRole string) string {
+func testAccTFESAMLSettings_siteAuditor(idpCert, attrSiteAuditor, siteAuditorRole string) string {
 	return fmt.Sprintf(`
 resource "tfe_saml_settings" "foobar" {
-  idp_cert          = "testIDPCertSiteAuditor"
+  idp_cert          = "%s"
   slo_endpoint_url  = "https://foobar.com/slo_endpoint_url"
```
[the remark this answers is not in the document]
skj-skj: fixed

ctrombley (approved):
Looks good, thanks @skj-skj!

PR #2207, opened 2026-09-02
[TF-40056]: Cannot create Vault OIDC configurations with ca cert

danieldnedialkov:
## Description

Bumped go-tfe version and updated a function call in `r/tfe_vault_oidc_configuration` to match updated api call in atlas 

_Remember to:_

- [x] _Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md)_
- [ ] _Update the [Documentation](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md#updating-the-documentation)_

## External links

- [Jira](https://hashicorp.atlassian.net/jira/software/c/projects/TF/boards/438?assignee=712020%3A486f62b2-b976-45cd-85bf-8c6058254907&selectedIssue=TF-40056)

## Rollback Plan

Revert go-tfe to a previous version and use the older API 

## Changes to Security Controls

N/A

PR #2217, opened 2026-09-09
Document the Site Auditor SAML and SCIM attributes

tanushreegorai:
## Description

Documentation for the Site Auditor attributes added in #2201 and #2209, which shipped minimal descriptions so CI would pass. Docs here are generated, so the changes are to Go description strings plus `make generate`.

- Polished the four attribute descriptions across both resources and data sources.
- Added a minimum-version note (v2.1.0) to both resource pages, plus a sentence on each attribute.
- Documented a side effect previously undocumented for **either** role: clearing a group mapping unlinks the group **and revokes that role from every member of it**.
- Added a usage example for linking a Site Auditor group.

Two things worth flagging:

- The version note drops the `docs/backwards-compatibility.md` template's "HCP Terraform or" clause — both resources are `(Only for Terraform Enterprise)` and their tests `skipIfCloud`, so naming HCP Terraform would imply support that doesn't exist.
- Fixes a pre-existing wrong description: on the SAML data source, `attr_site_admin` repeated `site_admin_role`'s text, describing the role rather than the SAML attribute that determines it.

- [x] Update the [Change Log](https://github.com/hashicorp/terraform-provider-tfe/blob/main/docs/changelog-process.md) — added the revoke consequence to the existing SCIM entry rather than adding a new one

## Testing plan

Docs-only — no behaviour changes.

1. `make generate` produces no diff beyond the four Site Auditor pages.
2. `make exmplcheck` passes, including `terraform validate` on the new example.
3. Check the rendered pages: version note on both resource pages, unlink/revoke note on the SCIM resource page, new example sorting after the site admin two-apply series.

## External links

- [SAML attributes](https://github.com/hashicorp/terraform-provider-tfe/pull/2201)
- [SCIM group mapping](https://github.com/hashicorp/terraform-provider-tfe/pull/2209)

No acceptance tests apply — description strings, generated markdown and one example file, with no runtime behaviour.

Green locally on current `main` (including the go-tfe/v2 v2.9.0 bump from #2207): `go build`, `go vet`, `gofmt`, `make lint` (0 issues), `make exmplcheck`, and the SAML/SCIM unit tests. `make generate` at v2.9.0 leaves all four pages unchanged.

`check-docs-current.sh` still fails on 11 unrelated pages — pre-existing on `main`, left out to keep this diff scoped. Worth its own ticket.

## Rollback Plan

If this change needs to be reverted, we will roll out an update to the code within 7 days. Documentation-only, so reverting restores the previous wording with no functional impact.

No functional changes. It does document an existing security-relevant behaviour more clearly: clearing a Site Admin or Site Auditor group mapping revokes that role from every member of the group.


kkavish (approved):

internal/provider/resource_tfe_saml_settings.go
```diff
@@ -316,13 +317,13 @@ func (r *resourceTFESAMLSettings) Schema(ctx context.Context, req resource.Schem
 				Default:     stringdefault.StaticString(samlDefaultSiteAdminRole),
 			},
 			"attr_site_auditor": schema.StringAttribute{
-				MarkdownDescription: fmt.Sprintf("Specifies the role for site auditor access. Overrides the \"Site Auditor Role\" method. Requires Terraform Enterprise %s or later.", minTFEVersionSiteAuditor),
+				MarkdownDescription: fmt.Sprintf("Specifies the role for site auditor access. Overrides the \"Site Auditor Role\" method. This attribute requires an instance of Terraform Enterprise at least as recent as v%s.", minTFEVersionSiteAuditor),
```
ctrombley: `attr_site_auditor` is the name of the boolean SAML attribute that designates site auditors, not the role itself; `site_auditor_role` below is the role value carried in the team attribute. As written, these two arguments appear to accept the same kind of value and may lead users to configure a role name here. Please reword this along the lines of the data-source description (`Name of the SAML attribute that determines site auditor access`) and regenerate the markdown.

ctrombley (changes requested):
Found one documentation correctness issue: `attr_site_auditor` is described as a role instead of the SAML attribute name. The rest of the change looks consistent, and CI is green.

ctrombley (approved):
Looks good, thanks @tanushreegorai !

PR #2222, opened 2026-09-14
Add tfe_registry_artifact_tag resource and datasource

mrinalirao:
## Description

Adds support for managing tag bindings on registry artifacts.
New resource: **tfe_registry_artifact_tag**
Manages the complete set of tags on a registry module, provider or component. 

_Remember to:_

## Testing plan

1.  Acceptance tests
2. E2E tests

## External links
- [Memo](https://ibm.sharepoint.com/:w:/r/sites/hermes/Shared%20Documents/DraftDocuments_New/TF-Terraform%20Provider%20support%20for%20Registry%20Artifact%20Tags.docx?d=wcf2dca0440e94d5e9283417e681f63c2&csf=1&web=1&e=VnyQX1)
- [go-tfe PR](https://github.com/hashicorp/go-tfe/pull/1318)
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/registry/manage/tags)

**DataSource Tests:**

Screenshot 2026-09-14 at 12 14 03 pm

Screenshot 2026-09-14 at 12 13 37 pm

Screenshot 2026-09-14 at 12 13 12 pm

Screenshot 2026-09-14 at 12 12 38 pm

**Resource Tests:**

Screenshot 2026-09-14 at 12 01 28 pm

Screenshot 2026-09-14 at 11 57 18 am

Screenshot 2026-09-14 at 11 56 44 am

Screenshot 2026-09-14 at 11 56 07 am

Screenshot 2026-09-14 at 11 49 42 am

Screenshot 2026-09-14 at 11 33 34 am

**E2E Tests:**

tested against staging with the following main.tf:

```
provider "tfe" {
  hostname = "app.staging.terraform.io"
  token    = <REDACTED>

data "tfe_registry_module" "this" {
  organization    = "torg"
  name            = "module"
  module_provider = "aws"
  registry_name   = "private"
}

resource "tfe_registry_artifact_tag" "this" {
  artifact = {
    type = "registry-module"
    id   = data.tfe_registry_module.this.id
  }

  tags = [
    { key = "env", value = "prod" },
    { key = "team", value = "platform" },
  ]
}


data "tfe_registry_artifact_tags" "verify" {
  artifact   = tfe_registry_artifact_tag.this.artifact
  depends_on = [tfe_registry_artifact_tag.this]
}

output "tags" {
  value = tfe_registry_artifact_tag.this.tags
}

terraform {
  required_providers {
    tfe = {
      source  = "hashicorp/tfe"
      version = "9.9.9"
    }
  }
}
```

Screenshot 2026-09-14 at 1 00 12 pm

Screenshot 2026-09-14 at 1 00 47 pm

## Rollback Plan

Revert PR

NA

## Documentation

**DataSource**

Screenshot 2026-09-14 at 2 21 53 pm
Screenshot 2026-09-14 at 2 22 01 pm

**Resource**

Screenshot 2026-09-14 at 2 22 23 pm
Screenshot 2026-09-14 at 2 22 44 pm
Screenshot 2026-09-14 at 2 22 52 pm


internal/provider/data_source_registry_artifact_tags.go
```diff
@@ -0,0 +1,157 @@
...
+	switch artifactType {
+	case ArtifactTypeRegistryModule:
+		return d.config.Client.RegistryModules.ListTagBindings(ctx, artifactID)
+	case ArtifactTypeRegistryProvider:
+		return d.config.Client.RegistryProviders.ListTagBindings(ctx, artifactID)
+	case ArtifactTypeRegistryComponent:
+		return d.config.Client.RegistryComponents.ListTagBindings(ctx, artifactID)
+	default:
+		return nil, fmt.Errorf("unsupported artifact type: %s", artifactType)
+	}
```
glennsarti-hashi: For discussion: This makes me a little nervous but I can understand why it is the way it is. I don't like magic lists this far down the abstraction tree.

I wonder if it would be possible to not make this a hardcoded list, or at least delegate this logic to the Client instead of in here.

internal/provider/resource_tfe_registry_artifact_tag.go
```diff
@@ -0,0 +1,383 @@
...
+	case ArtifactTypeRegistryModule:
+		return r.config.Client.RegistryModules.ListTagBindings(ctx, artifactID)
+	case ArtifactTypeRegistryProvider:
+		return r.config.Client.RegistryProviders.ListTagBindings(ctx, artifactID)
+	case ArtifactTypeRegistryComponent:
+		return r.config.Client.RegistryComponents.ListTagBindings(ctx, artifactID)
+	default:
+		return nil, fmt.Errorf("unsupported artifact type: %s", artifactType)
+	}
+}
```
glennsarti-hashi: This is a duplicate from `internal/provider/data_source_registry_artifact_tags.go` seems like this should be a utility or helper

glennsarti-hashi (dismissed):
Approved but we should DRY it up

internal/provider/data_source_registry_artifact_tags.go
mrinalirao: hmmm...delegating to the client is tricky since go-tfe has no generic tag-binding API and no shared interface across the three artifact type. I extracted this into a helper and removed the hardcoded list. Let me know if that looks better? 😅

internal/provider/data_source_registry_artifact_tags.go
```diff
@@ -0,0 +1,139 @@
...
+				Attributes: map[string]schema.Attribute{
+					"type": schema.StringAttribute{
+						Description: "The type of the registry artifact.",
+						Required:    true,
+						Validators: []validator.String{
+							stringvalidator.OneOf(helpers.RegistryArtifactTypes()...),
+						},
+					},
+					"id": schema.StringAttribute{
+						Description: "The external ID of the registry artifact.",
```
brandonc, line 85:
```suggestion
						Description: "The ID of the registry artifact.",
```

internal/provider/resource_tfe_registry_artifact_tag.go
```diff
@@ -0,0 +1,365 @@
...
+						Required:    true,
+						Validators: []validator.String{
+							stringvalidator.OneOf(helpers.RegistryArtifactTypes()...),
+						},
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.RequiresReplace(),
+						},
+					},
+					"id": schema.StringAttribute{
+						Description: "The external ID of the registry artifact.",
```
brandonc, line 130:
```suggestion
						Description: "The ID of the registry artifact.",
```

brandonc (commented):
Hey, there. I am curious about the design choice inherent in the PR. For example, we already use a string map on tfe_project and tfe_workspace for managing tag bindings. I'm guessing the choice to make it a separate resource has something to do with being able to bind to something called a "registry-component" which is not a resource you can manage with the provider?

I am not necessarily against a separate resource for tag management, but I'm wondering about your thoughts on these questions:

1. Could this resource potentially support other things that have tag bindings in the future? (in which case maybe we drop the word "registry" from it)
2. Would a "tags" map be more suitable on the resources being managed since there is very little possibility of a dependency loop on tags? Is there a perceived drawback to using this data schema that makes it unsuitable? (Provided we add a resource to manage whatever a "registry-component" is) 
3. Where did the name "artifact" come from? Is this a word we already use in the context of registry resources?

mrinalirao:
@brandonc Happy to provide a bit more context below.

> I'm guessing the choice to make it a separate resource has something to do with being able to bind to something called a "registry-component"

Not exactly. Registry tags have a different set of rules compared to workspace or project tags. Registry artifacts (modules, providers, and components) can only be associated with project tags and cannot have arbitrary tags attached to them. In addition, registry artifact versions can only be associated with a special `reserved registry version` tag. Because these requirements are unique to registry-related resources, the same model doesn't apply to other artifact types.

Another consideration was implementation scope. As outlined in [TF-Terraform Provider support for Registry Artifact Tags](https://ibm.sharepoint.com/sites/hermes/_layouts/15/Doc.aspx?sourcedoc=%7BCF2DCA04-40E9-4D5E-9283-417E681F63C2%7D&file=TF-Terraform%20Provider%20support%20for%20Registry%20Artifact%20Tags.docx&action=default&mobileredirect=true)  extending existing resources introduced a fair amount of additional complexity and scope growth, which was one of the reasons we landed on a dedicated resource approach.

> Could this resource potentially support other things that have tag bindings in the future? (in which case maybe we drop the word "registry" from it)

Potentially yes, but I would be hesitant to broaden it today. The tag binding model for registry artifacts differs from how tags are managed for workspaces and projects, so the resource is intentionally scoped to registry-related use cases for now. If we eventually introduce other resource types that share the same binding semantics, it could make sense to revisit the naming and scope at that point.

> Would a "tags" map be more suitable on the resources being managed since there is very little possibility of a dependency loop on tags? Is there a perceived drawback to using this data schema that makes it unsuitable? (Provided we add a resource to manage whatever a "registry-component" is)

A map would definitely be simpler from a syntax perspective, but it becomes more limiting once we consider some of the registry-specific requirements. For example, reserved registry version tags can support multiple values for a single key (up to 10 values). While that functionality isn't in scope for this PR, we wanted to leave room for future extensions that could support it.

We also anticipate needing additional validation around tag metadata, such as enforcing value limits and validating registry specific attributes (for example, whether a tag is marked as a registry version tag). The current schema gives us a bit more flexibility to evolve in that direction without introducing breaking changes later.

I'm also completely open to using a tags map if that's your preference. It does have the advantage of being simpler, and we can reassess the design in the future once we start adding value validations and other registry-specific constraints 🙂 

> Where did the name "artifact" come from? Is this a word we already use in the context of registry resources?

Artifact is the terminology we already use in the registry domain, so I was aiming to stay consistent with existing language. For example, the Registry documentation refers to registry resources as artifacts in several places: 
 - https://developer.hashicorp.com/terraform/cloud-docs/registry/manage/tags
 - https://developer.hashicorp.com/terraform/cloud-docs/registry/using#find-artifacts

Let me know if anything still feels unclear 🙂

glennsarti-hashi (dismissed):

brandonc (dismissed):

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 
 FEATURES: 
-* `r/tfe_workspace_hyok_enabled`: Adds a resource to enable HYOK (Hold Your Own Key) on a workspace. Destroying the resource leaves HYOK enabled on the workspace (non-destructive). By @danieldnedialkov [#2192](https://github.com/hashicorp/terraform-provider-tfe/pull/2192)
+* `r/tfe_workspace_hyok_enabled`: Adds a resource to enable HYOK (Hold Your Own Key) on a workspace. Destroying the resource leaves HYOK enabled on the workspace (non-destructive). By @danieldnedialkov [#TF-39152](https://github.com/hashicorp/terraform-provider-tfe/pull/2192)
```
brandonc: Sorry- one last thing- Can this line be reverted?

shwetamurali (approved):

